PR #5, opened 2015-05-07
init method

haugene:
Got rid of the base image. Running everything through service hooks in OpenVpn. Cleaner.


ianneub:
:+1: For closing transmission when the VPN disconnects.

PR #6, opened 2015-05-09
Save about 49mb from the final image

ianneub:
Save space by combining all apt commands into a single RUN statement.


ianneub:
Replaced with #9

PR #10, opened 2015-05-09
Control PIA and Transmission from env vars

ianneub:
Also reducing size of Docker image.


ianneub:
Ok I think this is ready to go now. Phew, sorry for all the clutter. I messed up the rebase/merging. But everything should be good now, and it's all condensed into one commit :+1:

runOpenVpn.sh
```diff
@@ -5,14 +5,37 @@ then
...
+
+# add PIA user/pass
+if [ "$PIA_USERNAME" != "**None**" ];
+then
+  echo "Setting PIA credentials..."
+  mkdir -p /config
+  echo $PIA_USERNAME > /config/pia-credentials.txt
+  echo $PIA_PASSWORD >> /config/pia-credentials.txt
+else
+  echo "Not setting PIA credentials."
```
haugene: Should probably not allow startup without credentials. Thinking exit here...

runOpenVpn.sh
```diff
@@ -5,14 +5,37 @@ then
...
+		OPEN_VPN_CONFIG=/etc/openvpn/Netherlands.ovpn
 	fi
 else
 	echo "No VPN configuration provided. Using default: Netherlands"
-	exec openvpn --config /etc/openvpn/Netherlands.ovpn
+	OPEN_VPN_CONFIG=/etc/openvpn/Netherlands.ovpn
 fi
 
+# override resolv.conf
+if [ "$RESOLV_OVERRIDE" != "**None**" ];
```
haugene: Don't have to init env variables as None, can just check for empty

runOpenVpn.sh
```diff
@@ -5,14 +5,37 @@ then
...
+  echo "Setting PIA credentials..."
+  mkdir -p /config
+  echo $PIA_USERNAME > /config/pia-credentials.txt
+  echo $PIA_PASSWORD >> /config/pia-credentials.txt
+else
+  echo "Not setting PIA credentials."
+fi
+
+dockerize \
+  -template /etc/transmission-daemon/settings.json:/etc/transmission-daemon/settings.json \
```
haugene: Setting transmission-daemon setting from environment variables. Then, if the port updater should when TRANSMISSION_RPC_AUTHENTICATION_REQUIRED is set to true, the updater should also read the environment variables

haugene:
Left some notes, I'm gonna do a little refactor. But good, merging now!

runOpenVpn.sh
ianneub: Actually did this by design. Totally ok if you don't what it there. I like to have it there as a kind of reminder that the option is available. Some services will read the ENV vars from the Dockerfile and show them to the user (tutum.co). 

But if you don't want it there no worries at all :-)

runOpenVpn.sh
ianneub: :+1:

runOpenVpn.sh
ianneub: Good idea! :+1:

Dockerfile
```diff
@@ -36,9 +28,81 @@ ADD transmission/runUpdates.sh /etc/transmission-daemon/startPortUpdates.sh
 ADD transmission/down.sh /etc/transmission-daemon/stop.sh
 ADD runOpenVpn.sh /etc/openvpn/start.sh
 
+ENV PIA_USERNAME=**None** \
+    PIA_PASSWORD=**None** \
```
ianneub: I also set these to a default bogus value. You can remove these as well if you'd like.

runOpenVpn.sh
haugene: Fixed in faa3972ab42e528334958f69ac1bab956feb9cda

runOpenVpn.sh
haugene: Ah ok, did not know that. Just thought it looked verbose. But, then it has a purpose :)

Issue #4, opened 2015-02-23
Running in Boot2Docker, PIA can't be resolved


justrhysism:
I'm pretty new to Docker, so I'm still struggling a touch to figure it all out, and it probably doesn't help that I'm using it with Boot2Docker. Anyhow...

The script is failing because www.privateinternetaccess.com can't be resolved:
```
Mon Feb 23 12:04:15 2015 TUN/TAP device tun0 opened
Mon Feb 23 12:04:15 2015 do_ifconfig, tt->ipv6=0, tt->did_ifconfig_ipv6_setup=0
Mon Feb 23 12:04:15 2015 /sbin/ip link set dev tun0 up mtu 1500
Mon Feb 23 12:04:15 2015 /sbin/ip addr add dev tun0 local ##.###.#.6 peer ##.###.#.5
Mon Feb 23 12:04:15 2015 Initialization Sequence Completed
head: cannot open '/config/transmission-credentials.txt' for reading: No such file or directory
tail: cannot open '/config/transmission-credentials.txt' for reading: No such file or directory
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:00:19 --:--:--     0curl: (6) Could not resolve host: www.privateinternetaccess.com
Got new port  from pia
transmission auth not required
localhost:9091/transmission/rpc/ responded: "success"
Checking port...
Error: portTested: http error 0: No Response
```

Would you have any ideas on what this might be?

justrhysism:
Actually, after some further research, I have a pretty strong feeling that this is why: https://github.com/boot2docker/boot2docker/issues/357

haugene:
Hey! I've been away for a while. Did you resolve the issue? Was it docker dns settings that was the problem? I've stumbled upon a similar problem on linux in another project. I ended up setting global docker daemon dns servers so that all containers had a working internet connection...

ianneub:
I'm also seeing this problem. Not using boot2docker, but I am on OS X running Docker inside a VirtualBox VM, I'm thinking it's related to that.

@justrhysism Did you resolve this issue?

ianneub:
I was able to get this to work by replacing /etc/resolv.conf after the container started. Not ideal, but it works, just FYI.

justrhysism:
[...]

PR #12, opened 2015-05-11
Externalized configuration by environment variables.

### LINKED ISSUE #8: Configure Transmission using env vars
If you would accept a PR, I'd like to create env vars that will override the transmissions config options at start up. This would let the image be a lot more configurable, without requiring a user to mount a config folder.

What do you think?

haugene:
Can now control PIA credentials, VPN server selection and Transmission settings from environment variables. closes #8 and also provides a fix for #4


haugene:
Tested it, and looks to behave well!

Issue #11, opened 2015-05-09
Feature: Bind transmission ipv4 address to OpenVPN client IP


ianneub:
When starting up transmission we should bind it to the IP given to us from PIA. This should prevent leaking our IP.

The transmission setting is `bind-address-ipv4`.

## Script to get the IP of `tun0`:
```
# ifconfig tun0 | sed -n '2 p' | awk '{print $2}' | cut -d: -f2
10.181.1.6
```
## Things to consider
* [ ] What if the IP changes?
* [x] Will the interface always be `tun0`?

haugene:
Ref: https://openvpn.net/index.php/open-source/documentation/howto.html
> First, define a static unit number for our tun interface, so that we will be able to refer to it later in our firewall rules:
>
>dev tun0

Looks like we can update the config files with "dev tun0" instead of "dev tun" to statically set it...

haugene:
If the IP changes because the interface went down and came up again, we should be OK. Transmission will be killed on interface down, and restarted on interface up.

So, can the IP change without interface going up and down?

PR #16, opened 2015-05-23
Bind transmission to tun0

ianneub:
I believe this will solve #11 

It forces openvpn to use `tun0`. Then determines the IP of said interface and updates the settings.json config before starting transmission.

This will not work if the user has created a /data/transmission-data folder and is using that.

Maybe we should get rid of that? The user could simply replace the `/etc/transmission-daemon/settings.json` with their own file as easily as mounting it to /data.


haugene:
Yes, looks good!

About the /data/transmission-data, I'm using that one. But it's not for the settings.json, that will be rewritten from the template on every startup anyways. It's because Transmission stores torrent state there, like which torrents are added, active, paused, etc.
I have this container in a docker-compose and I have been tearing down and setting the containers up a lot, then it's nice not to have to add the torrents again.
But... I will admit, it's not that big deal. If people want they can mount the /etc/transmission-daemon as well to keep state. So yes, we should maybe remove it...

Can you recreate the pull request against the dev branch? Then I can make some small modifications as well, before pushing new version.

PR #18, opened 2015-05-24
Bind transmission IPv4 address

haugene:
I've removed the /data/transmission-daemon option, and it cleans things up a bit.
Tried to move the dockerize templating to transmission startup script but the env vars weren't accessible in the transmission user. So now I write them to a file on container startup, and source them into the shell that starts transmission.

I felt it looked cleaner to let the transmission start script handle settings.json generation. And then dockerize handles it all, instead of manually regex'ing the settings.

Any thoughts @ianneub ?


haugene:
Yeah, and because of the minor change in our "API" by removing KEEP_TRANSMISSION_STATE functionality I'm bumping to 1.1

ianneub:
I think that's a great idea. LGTM :+1:

PR #25, opened 2015-08-13
Fix transmission not resuming torrents on docker stop/start

### LINKED ISSUE #23: Torrents don't resume on docker container stop/start
Torrents aren't resuming when the docker container is stopped/started, even though the `/data/incomplet` and `/data/watch` has `*.torrent.added` and `*.part` files.

I think there must be some other Transmission configuration or directory that is managing this state that needs to be located in the `/data` or `/config` volumes so it can be restored on docker container restart.

dougnukem:
Closes #23 
- Ensure TRANSMISSION_HOME is on a persistent volume to store torrent resume info
  - https://trac.transmissionbt.com/wiki/ConfigFiles

References #24


dougnukem:
@haugene 
"Hey. Nice addition, I like it. Would you mind submitting your PR to the dev branch? I've made some changes to the image as well and then I could test it all together before merging.
I also addressed the persistent state issue. I separated transmission scripts copied into the container from the folder keeping transmission state. This allows for mounting a separate volume for the transmission state.

I propose to keep both options. Both provide transmission_home as a configurable path, and allow for a separate mounted volume for the state. Sometimes you might want a separate mounted volume, but for the most part /data/transmission-home is the easiest option. Maybe set transmission_home to the /etc-path as default? I don't know, I prefer that the default run option starts with a clean slate."

dougnukem:
@haugene there is something a little odd about creating the `/data/transmission-home` in the `/data` volume, but I don't think we want to use `/etc/transmission` as the default `TRANSMISSION_HOME` as it will not allow torrents to resume on docker container stop/start.

I wish there was a way to configure and specify the `/etc/transmission/resume`, `/etc/transmission/torrents` locations, because it also feels odd that `/data/transmission-home` will contain a persistent `settings.json`.

haugene:
Yeah, I was thinking that people could choose to activate it since it actually writes data on the disk. But I think that most users will want it, and so it should be default.

Persisting settings.json comes with it, but it doesn't matter. Any change will be overridden on startup anyways. But I agree, it would be nicer to have it inside the container.

PR #28, opened 2015-08-18
Added TigerVPN as a new Provider

tsromana:
added Tiger vpn as the new OPENVPN provider


openvpn/start.sh
```diff
@@ -4,9 +4,12 @@ if [ "$OPENVPN_PROVIDER" = "BTGUARD" ]
 then
 	echo "VPN PROVIDER: BTGUARD"
 	vpn_provider="btguard"
-else
+else if [ "$OPENVPN_PROVIDER" = "PIA" ]
```
dougnukem: For backwards compatibility PIA should probably be the "default" VPN provider and any new ones should be called out explicitly.

dougnukem: Might not be a bad idea to just make `$OPENVPN_PROVIDER` a required input when running the container.

haugene: I agree @dougnukem . Thinking something like:
if provider = btguard
else if provider = pia
else if ...
else echo "No valid provider given"; exit 1
fi

ish?
This will also be backward compatible as OPENVPN_PROVIDER is set to PIA in Dockerfile. Could of course remove the default and make provider a required input as well. Defaulting to PIA makes less sense when the number of providers grows...

tsromana:
hey sorry i did a spell error have opened a pull request with the correct filename

haugene:
I didn't see it either. Merged now :+1:

PR #31, opened 2015-08-19
corrected spell eror

tsromana:
corrected deafult.ovpn to default.ovpn


haugene:
also merged to master

PR #33, opened 2015-10-07
Add NordVPN as provider

jaysonsantos:


jaysonsantos:
I kind didn't get it at first glance but it is fixed and tested on the last commit.

haugene:
Nice! Would you mind recreating the PR against the dev branch though?

PR #54, opened 2016-03-09
Add a link from nginx to openvpn in docker-compose

trentsummerfield:
I had some trouble getting nginx to reverse-proxy to transmission using my local IP address. The solution in the end was to add a link from the nginx container to the openvpn container. This causes docker-compose to map the name 'openvpn' to the IP address of the actual container at start up.

This would also allow you to include a portable nginx.conf file in the repository that didn't require the user to change anything. Let me know if you would like me to open a second pull request adding a copy of my nginx.conf that works with this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ http {
 }
 ```
 Your Transmission WebUI should now be avaliable at "your.host.ip.addr:8080/transmission/web/".
-Change the port in the docker run command if 8080 is not suitable for you.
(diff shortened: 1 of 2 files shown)


haugene:
Yes, I use container linking myself from the nginx. Adding this to the docker-compose would allow people to use openvpn as hostname out of the box.

It would be nice if you appended a couple of lines to the "How to fix this"  section of WebUI access in the README. Something like "if you use container linking either in the run command or via docker-compose, host.ip.address can be changed to openvpn or whatever you called your link".

trentsummerfield:
Sure. I've updated the README.

PR #73, opened 2016-05-08
Added IVPN support

aaronblair:
Didn't end up creating separate certificates as they were embedded directly in the ovpn files. Works perfectly on my machine.


haugene:
Sweet :+1: This looks good! Just a couple of comments. You should also update the OPENVPN_PROVIDER config section of the README, just add IVPN as a valid option. At least I've been doing it so far. We'll see what happens when the list gets too long :)

Also, did you use the "adjustConfigs.sh" script to modify the .ovpn files? If you see my [last commits](https://github.com/haugene/docker-transmission-openvpn/commit/1be2b9fa439e2e4ee91a79573e19a1c751daeae1) on the dev branch I've moved the up/down part of the config into a launch option for openvpn, so it no longer needs to be inserted into each .ovpn file.

If you update your PR with this, I'm looking forward to merging a new provider!

aaronblair:
No problem. I'll try to change that around tonight when I get home. I believe I used the adjustConfigs.sh before you made that change. Also, I'll break out the keys to their own files as it's definitely a cleaner way of doing it.

aaronblair:
I re-updated the ovpn files with the latest of your script and added IVPN to the OPENVPN_PROVIDER list.

Additionally, I made a couple compatibility changes to the "adjustConfigs.sh" script itself so that it will work on both BSD and Linux flavours. Changing to double square brackets is the only potential risk as it doesn't work with legacy SH (BASH, ZSH, etc are fine). See the relevant articles from stackoverflow on the [SED change](http://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux) and the [double bracket change](http://stackoverflow.com/questions/13781216/bash-meaning-of-too-many-arguments-error-from-if-square-brackets).

haugene:
Great stuff! LGTM

Merging this do dev now. I'm just gonna update the instructions in the readme for how to add a new provider. It changed slightly since the up/down script was set up inside the .ovpn files, as you well know :) It'll be available with the dev tag and I'll merge it to master within the week. 

Thanks for your contribution!

Issue #69, opened 2016-04-28
Would you be interested in documentation for systemd integration?


d5ve:
Hi,

Thanks for this very useful piece of software.

I've set up docker-transmission-openvpn as a systemd service so that it starts at boot time, and restarts if it fails.

Would it be useful for me to document this for addition to the README? If so, I'll raise a pull request.

Cheers,

Dave

haugene:
Hey Dave, and thanks.

Yeah, that would be a great addition for the README. Just out of curiosity, how are you doing the restart on fail thing? Are you checking if the connection is working, or restart the container if it exits?

Looking forward to your PR

Kristian

d5ve:
Currently it is just triggered by the container not being running, so won't pick up the case where the VPN isn't active inside the running container.

haugene:
Allright. Then you might be able to combine it with the --ping-exit option for OpenVPN. If you set openvpn to exit when ping fails, then the container will stop. Then if you run the container with --restart=always, it should be restarted every time it goes down. See #61.

Was also thought to go into the README at some point :)

d5ve:
The following `/etc/systemd/system/transmission-openvpn.service` appears to work pretty well in testing. It does leak occasional zombie processes, depending on how I kill the container or things running on the container. My VPN connection is quite stable, so I've not yet had a real instance of --ping-exit triggering.

Do you have any suggestions on the zombie process issue?

```
[Unit]
Description=haugene/transmission-openvpn docker container
After=docker.service
Requires=docker.service

[Service]
User=bittorrent
TimeoutStartSec=0
ExecStartPre=-/usr/bin/docker kill transmission-openvpn
ExecStartPre=-/usr/bin/docker rm transmission-openvpn
ExecStartPre=/usr/bin/docker pull haugene/transmission-openvpn
ExecStart=/usr/bin/docker run \
        --name transmission-openvpn \
        --privileged \
        -v /home/bittorrent/data/:/data \
        -e "OPENVPN_CONFIG=Netherlands" \
        -e "OPENVPN_OPTS=--inactive 3600 --ping 10 --ping-exit 60" \
        -e "OPENVPN_PASSWORD=hunter2" \
        -e "OPENVPN_PROVIDER=TORGUARD" \
        -e "OPENVPN_USERNAME=bittorrent@example.com" \
        -e "TRANSMISSION_RATIO_LIMIT=5" \
[...]

PR #76, opened 2016-05-13
systemd integration

d5ve:
Re: #69 

README changes explaining about systemd integration.
Plus Dockerfile change to install and use Yelp's dumb-init to avoid zombies.

I notice that there is a new version of dumb-init release since I started using it 12 days ago but I left the Dockerfile with v1.0.1, as that's the one I've been testing.

Cheers,

Dave


haugene:
Nice! Would you mind putting the section in the readme at the end? It's a more specific use case than running it on your NAS, so I feel it should be further down the file.

haugene:
Doesn't matter, I'm going to do some adjustments on the README now anyways. I'll move it. 
Thanks for your contribution!

PR #81, opened 2016-05-30
Add support for local network route

lloesche:
This PR adds support for e.g. `-e LOCAL_NETWORK=10.0.0.0/24` so the container stays available on the local network and access doesn't require a proxy.


aaronblair:
This is a good very good addition. I think this is way cleaner than using a proxy.

haugene:
Nice! Definately nice to have it working without a proxy! Would you mind re-creating the pull-request towards the dev-branch? That way we get to test it first, and I want to bundle some more changes into a new version/tag of the container anyways.

lloesche:
Oh of course, recreated as #83.

PR #82, opened 2016-06-02
Updating configs NordVPN

CooperBarrett:
Updating .ovpn configs NordVPN. Add news, removal of old and updating
of existing ones


haugene:
Hey, quick fixing ;)

One thing though, you just imported the ovpn files directly from NordVPN? They need to be adjusted for the container. There is a script called adjustConfigs.sh that can help you with this. I'll also add a comment in the code to show what I mean.

Please also re-create this pull-request against the dev branch. I like to merge and test it there along with other changes before releasing to master.

PS: Do you want to change the default.ovpn? Is it pointing to an old config? If you're on linux you should be able to do something like "ln -s wanted-config.ovpn default.ovpn" and it will update the symlink.

openvpn/nordvpn/AU3 UDP.ovpn
```diff
@@ -28,8 +28,7 @@ reneg-sec 0
 remote-cert-tls server
 
 #mute 10000
-# Read authentication persisted from env
-auth-user-pass /config/openvpn-credentials.txt
```
haugene: Should not remove path to username/password

openvpn/nordvpn/AU10 UDP.ovpn
```diff
@@ -0,0 +1,92 @@
...
+persist-tun
+ping 15
+ping-restart 0
+ping-timer-rem
+reneg-sec 0
+
+remote-cert-tls server
+
+#mute 10000
+auth-user-pass
```
haugene: All configs needs path to username/password-file

CooperBarrett:
Oh ok sorry, actually, I really do anything ...
I'll repeat it to zero before returning. And ok, on the dev branch.

CooperBarrett:
Damn, I think I'm still on the master branch ...

haugene:
Yep, still on master. Don't think you can change branches for a pull request. You will need to close this and create a new one.

PR #83, opened 2016-06-02
Add support for local network route

lloesche:
This PR adds support for e.g. -e LOCAL_NETWORK=10.0.0.0/24 so the container stays available on the local network and access doesn't require a proxy.

This duplicates #81 against dev branch.


haugene:
LGTM. Great addition to the image! 
Thanks for contributing :)

PR #101, opened 2016-07-18
New PIA Profiles

wryspy:
Two changes:

1. Pulled new profiles from PIA after recent changes.  See https://www.privateinternetaccess.com/forum/discussion/21779/we-are-removing-our-russian-presence

**Note: I have not extensively tested each of these new profiles in the container, but the few I did test seemed to work properly.**

~~2. Copied a snippet of code that allows you to access the web interface from the local subnet.  Requires that you pass a new parameter HOST_SUBNET to the container.~~  Didn't notice that '-e LOCAL_NETWORK' was now a thing.  I was working off an old version.  Never mind.  Removed that commit.

Proceed at your own peril... Take it or leave it... if you choose to leave it, no hard feelings.

PR #102, opened 2016-07-18
Replaced OpenVPN profiles with current versions from PIA

wryspy:
Replaced OpenVPN profiles with current versions from PIA.  Modified slightly to point to proper paths where necessary.

Russia profile removed.  See https://www.privateinternetaccess.com/forum/discussion/21779/we-are-removing-our-russian-presence

Supersedes #101 - this PR goes against the dev branch rather than the master.


haugene:
LGTM! Thanks for contributing! :)
Will merge it to master after testing it.

haugene:
Works like a charm. Merged to master.

PR #111, opened 2016-07-31
Pull Updated TorGuard Configuration

bshuler:
This pulls the updated TorGuard Configuration from their servers and
configures it for this image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,16 @@ RUN apt-get update \
 ADD openvpn/ /etc/openvpn/
 ADD transmission/ /etc/transmission/
 
+# Get updated configuration from TorGuard
+RUN curl -L https://torguard.net/downloads/OpenVPN-UDP.zip -o OpenVPN-UDP.zip \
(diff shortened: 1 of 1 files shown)


haugene:
Hey! First of all, thank you for contributing and sorry I'm late in responding. Summer vacation and all :)

I really like the idea that the container would just build with the latest and greates of configs, but I'm not sure how it will scale when we look at all the providers. Right now there are 20 providers added and the number is growing steadily. I think the Dockerfile would get quite messy and also that all the HTTP requests would result in a lot of broken builds when servers aren't responding or one of the services goes out of business etc. Or, what do you think?

I would very much like to add your code to the project though. But if you could take those lines of code and make a bash script in the torguard folder called updateConfigs.sh or something that could be really handy. Then we could encourage such a script for all other providers and make a superscript that loops them all and automates the update of all configs (eventually).

Does that sound like a plan? Anyone else has any comments on this?

haugene:
Created #119 instead, which is towards the dev branch. WIll merge that one and move the code into a shell script as proposed. Thanks again!

PR #123, opened 2016-09-12
Updated nordvpn list

pziegler:


haugene:
Hey, and thanks for contributing! I see that the list hasn't been updated in some months so it's probably time for a config update :+1: 

**I do however have a couple of requests:**
It looks like you're committing a lot of duplicate files. There are three versions of almost every .ovpn file and they're named like someconfig.ovpn, someconfig.ovpn-- and someconfig.ovpn----
I did a quick check, and in [this zip](https://nordvpn.com/api/files/zip) there's 1200-something configs, while you're committing well over 3000.

Also if you could submit your PR towards the dev branch instead of master that would be great.
And last, please reset and commit fresh or squash your commits before your new PR so the 3000-files commit isn't merged.

Thanks again and I'll be waiting for a new pull request :smile:

pziegler:
hey, sorry about that. Was a bit to trigger happy :) 
will make a new PR then

Issue #124, opened 2016-09-15
Add SmartVPN to providers


tralce:
I'll help if I can. I can't figure out how to add it in myself; this is my first foray into Docker.

haugene:
Hi! I understand, not always easy getting started :) 

If you could provide me with a link to the config files for the provider I'll add them for you. Most providers have a link somewhere for setup on Linux or something. It's usually a zip file and it contains all the .ovpn files for the different server locations. 

You should also confirm that these aren't generated for you personally, with a private token instead of username/password. In that case you should see the issues for adding ExpessVPN and AirVPN. Start here in that case #105 

Let me know what you figure out!

tralce:
The configs /appear/ pretty universal, but I'm also not really an OpenVPN wizard. I'm learning OpenVPN and Docker specifically for this purpose. [Here](http://cloud.tralce.com/index.php?f=157da9a07f109a) is a link to my configs supplied by SmartVPN. Let me know how those go. Thanks for your fast reply.

haugene:
Yep, the configs looks very universal. @pziegler added the configs in the pull request referenced above, and they are now merged to the dev branch. That means they should be available if you use the dev tag of the image (haugene/transmission-openvpn:dev) and set OPENVPN_PROVIDER=SMARTVPN
Try it and see how it goes! 

As for the transmission config part there is no simple solution to add your existing config. The container is configured to read it from environment variables and persist a new file for each startup.  You could modify the build and add your own, but I would recommend that you check which of the settings you have actually changed from their default values and then setting them explicitly on the docker run command as the README describes under the "Transmission configuration options" section.

haugene:
This is now merged to master and available on the latest tag, closing issue

PR #127, opened 2016-09-16
Added smartvpn provider

pziegler:
Think this should work for #124


haugene:
Agreed, looks good! Thanks for solving the issue :+1: 
When adding a new provider there's also a list of providers in the README to update. But that can be added later, I'll merge this now so that @tralce can test it.

Issue #129, opened 2016-09-21
Move env into separate file


pkishino:
A suggestion to enhance this and simplify running:
Instead of passing in the various environmental variables on the command line, place them all in one file, perhaps if you switch vpns, one per provider. File looks like this
```
OPENVPN_PROVIDER=PROVIDER_NAME
OPENVPN_CONFIG=PROVIDER_CONFIG
OPENVPN_USERNAME=USERNAME
OPENVPN_PASSWORD=PASSWORD
LOCAL_NETWORK=0.0.0.0/24
no_proxy=*.local,169.254/16
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
TRANSMISSION_ALT_SPEED_DOWN=50
TRANSMISSION_ALT_SPEED_ENABLED=false
TRANSMISSION_ALT_SPEED_TIME_BEGIN=540
.
.
.
```
then when you load docker, instead of having all the -e XXX you just need one parameter:
`--env-file filename`
this will read in all the env variables then.
Well, I guess you could compile the whole thing into one compose or such but this way makes it easy to switch between vpn providers by just having a separate file for each.
You could even externalise the transmission part and just have one and source it into each file..
anyway, just an idea

I was thinking this could be added to readme or create a template env-file and add.

pkishino:
I'll look at creating a PR with this

haugene:
Hey!

Yes, this can greatly reduce the number of arguments to the `docker run` command and adding a hint on how to use this in the README under tips and tricks and maybe also a template file showing how it's done is a good idea.

I do however have a strong feeling that a lot of users run the container on their NAS and then they rather want to configure all the env variables through the NAS web interface instead of having it as a file in the nas system. But that'll be up to them.

See what you end up with, a PR with some usage tricks is always great :smile:

haugene:
Closing this after merge of #142

PR #133, opened 2016-09-28
Suggestion reported in #129

pkishino:
Here I have added the DockerEnv file I mentioned in #129 
*Also added SmartVPN to supported providers.


haugene:
Hey, neat to have this trick in the README. But should it really contain all the settings as a default? For my setup I would only put the variables I wanted to override from defaults, maybe it's just me :)

Anyways. Maybe move the docs down a little as we should introduce the most core configs to begin with. Maybe under the Transmission configuration part, as this really starts being handy when you override a lot of transmission settings. Or under tips and tricks. I think I'll also move the whole reverse proxy using nginx down a bit, it's not as relevant now that the LOCAL_NETWORK option is here.

And in your DockerEnv file, what's the "no_proxy" env variable you're setting?

pkishino:
well, I thought I would include a file that contains ALL the settings so people can make copies of it and customise.
Ill redo the readme a bit then.

Issue #132, opened 2016-09-27
Add Windscribe to VPN Providers


carlyman:
Switched VPN providers and forget to check whether you supported or not.  If it could be added, it'd be much appreciated.

Here's a scrubbed version of the opvn config:
[Windscribe.ovpn.txt](https://github.com/haugene/docker-transmission-openvpn/files/496831/Windscribe.ovpn.txt)

-JC

carlyman:
Missed the ability to add custom VPN from the readme as well as notes in #60 ..as such, closing issue.

carlyman:
I followed the process of a customer VPN per the readme and #60; however, the log looks like its getting stuck:
`Using OpenVPN provider: CUSTOM`
`No VPN configuration provided. Using default.`
`Setting OPENVPN credentials...`
`adding route to local network 192.168.0.0/24 via 172.17.0.1 dev eth0`
`Wed Sep 28 00:49:28 2016 OpenVPN 2.3.2 x86_64-pc-linux-gnu [SSL (OpenSSL)] [LZO] [EPOLL] [PKCS11] [eurephia] [MH] [IPv6] built on Dec  1 2014`

pkishino:
I can take a look at this, but please provide me with a windscribe.ovpn config WITH the certificates so I can add it to the repo

carlyman:
Which certificates do I share (want to make sure I don't share anything private obviously)...just the text in the ovpn file that I removed?   Additionally, dere's a link to the CA Cert and TLS Auth Key: https://assets.windscribe.com/desktop/other/openvpn_cert.zip

haugene:
If the .ovpn files are not public, ie they are generated for you personally then there's no way around the CUSTOM approach. As long as you make the appropriate changes in paths to ca/tls keys in your ovpn files it should work...

carlyman:
Ok, I checked and from what I can tell the ovpn file is not uniquely created or contain anything unique to my account.  There are differences based on which country you choose it seems, so not sure how best to get you that info?  I've attached one example: 
[Windscribe-Switzerland.ovpn.txt](https://github.com/haugene/docker-transmission-openvpn/files/501782/Windscribe-Switzerland.ovpn.txt)

pkishino:
You should be able to download or link me a zip containing all the ovpn
configurations

carlyman:
Here you go
[...]

PR #136, opened 2016-10-02
Windscribe support

pkishino:
added windscribe #132


haugene:
Hello again Patrick, and thank you for your continued contributions to this container! This all looks good to me, I just have a couple of comments on this before merging.

I suggest a rename of the files, removing the "Windscribe-" prefix. The user will already specify provider as being windscribe and I think it would be cleaner to then just specify "Australia-tcp" etc as provider. And also, adding windscribe to the README would be great ;)

pkishino:
good suggestion, will fix

haugene:
Sweet! But after renaming the files you need to recreate the symlink for the default.ovpn.
Do you mind doing that last fix, and I'll merge it? Been setting up a new computer over the weekend and haven't really gotten to the development part yet. Thanks again!

PR #139, opened 2016-10-02
Suggestion reported in #129

pkishino:
Cleaned up Readme and removed the no_proxy..no idea where that came from


DockerEnv
```diff
@@ -0,0 +1,78 @@
+OPENVPN_PROVIDER=**None**
+OPENVPN_CONFIG=**None**
+OPENVPN_USERNAME=**None**
+OPENVPN_PASSWORD=**None**
+LOCAL_NETWORK=**None**
```
haugene: Leaving this as default will probably fail on startup, right? It's not an ip address so I guess that would be tricky for the ip tables commands reading that variable.

DockerEnv
```diff
@@ -0,0 +1,78 @@
...
+PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```
haugene: Is the PATH override necessary?

README.md
```diff
@@ -84,6 +87,20 @@ As you can see the variables are prefixed with `TRANSMISSION_`, the variable is
 PS: `TRANSMISSION_BIND_ADDRESS_IPV4` will be overridden to the IP assigned to your OpenVPN tunnel interface.
 This is to prevent leaking the host IP.
 
+#### Use docker env file
+It is recommended to use a docker env file where you can easily store all your env variables and maintain multiple configurations for different providers.
```
haugene: How about: "You can also use a docker env file to easily store all your env variables and ..." ?

haugene:
Hey again! This is looking good 😄 I think the placement of the docker env info is good after all the config stuff, but I've merged a couple of PRs so you need to rebase to resolv the conflicts after those before I can merge.

I've also added a couple of comments on some of the default values. And a morge general comment is if we need to specify all of them in the example config? It's more backwards compatible to only specify the settings you want to override, as new transmission settings or other could change the list. Maybe have all the mandatory ones and a couple of relevant transmission settings set.

Also, I want to be a little picky on the text in the readme. Whether it's recommended to use the env file or not depends on the platform you're running on and the need you have for a completely separate configuration. I have therefore also suggested a small rewrite to the first line of the readme section.

DockerEnv
pkishino: yeah, I will look if the env file supports comments

DockerEnv
pkishino: probably not

README.md
pkishino: I aggree, after moving I didnt check the text

pkishino:
Thanks for checking, will rebase and fix

PR #140, opened 2016-10-02
Update README.md

mrwolf78:
Synology command doesn't work; -name should be --name, name should be between "", and the position in the command doesn't work.


haugene:
Yep, that looks right ;) Thanks for contributing!

PR #147, opened 2016-10-05
lux-c02.ovpn

madbean:
add luxembourg


haugene:
LGTM! :+1: 

Thanks for your contribution. If newshosting have updated their list of providers I'll look into doing a full refresh of all the configs.

Issue #151, opened 2016-10-15
Vyprvpn provider


starsys:
Hello.
I'm just discovering Docker and all its possibilities.
I'd like to use transmission with Vyprvpn on my DS 411 II+ Nas.
Until now I cannot make it work, as it seems I've to set different settings as explained on the hub.docker page.
I'm not so familiar with all these tweaks, I hope I'll manage to do it (if anyone can, help me, I would really appreciate and thanks).
I cannot see any Vyprvpn configuration. 
Is it please possible to add it ?
Thanks again in advance.

pkishino:
can you provide a link to the ovpn files from vypr please?

starsys:
Hello.
Thanks a lot for your help.

The config page is here:
https://support.goldenfrog.com/hc/en-us/articles/204088673-VyprVPN-OpenVPN-Setup-for-Synology-NAS


2016-10-15 17:40 GMT+02:00 Patrick Kishino <notifications@github.com>:

starsys:
Hello.
Let me know if you need something else.

haugene:
Most providers have a .zip file with all the .ovpn configuration files. If you can find that and post the link here, that would be great...

starsys:
Hello Kristian.
Here it is.

Thanks.

2016-10-16 22:11 GMT+02:00 Kristian Haugene <notifications@github.com>:

haugene:
Can't see that your comment contained any link to configs. Seems like you're replying by mail, don't think GitHub supports attachments if that's what you're up to...

starsys:
Hello.
You can download open vpn files here :
https://support.goldenfrog.com/hc/en-us/articles/225607167-Where-can-I-find-the-OpenVPN-files-

Le lundi 17 octobre 2016, Kristian Haugene <notifications@github.com> a
écrit :

starsys:
Hello Patrick. I'm not sure to have well understood your last mail. Have
you everything what you need ?

haugene:
Yes he did :) Patrick fixed it and I've now merged his pull request to the development branch.

You should be able to use VyperVPN as provider using `OPENVPN_PROVIDER=VYPERVPN` and image `haugene/transmission-openvpn:dev`

Shout out when you've tested it if it works as expected ;)

starsys:
Hello.
I'm a newbee regarding docker.
Would you accept to help me a bit ?
I tried following command line via SSH on my synology NAS :

[...]

PR #156, opened 2016-10-19
Added support for VyperVPN #151

pkishino:
Fix for #151


haugene:
Sweet, thanks again! 👍 
Probably about time to make you a collaborator here. If you want to join, that is ;)

haugene:
One quick note. Had to rename the vypervpn folder as it was not named the same as the OPENVPN_PROVIDER param in the table. Also created symlink to default.ovpn.

And this provider was special as it had a keepalive statement in its configs. When I tried to run the container with this provider I got `Options error: --keepalive conflicts with --ping, --ping-exit, or --ping-restart.  If you use --keepalive, you don't need any of the other --ping directives.`

So I removed that config from all the .ovpn files as we're having the case where containers lose dns and need to exit to be restarted and load networing again. You can see it in my last two commits.

pkishino:
aha, good spotting there.
Sure, I find it quite useful so would be interested in helping out :)

PR #163, opened 2016-11-14
Added wget

Christilut:
My setup uses wget to initiate sickrage postprocessing. Its only a minor change but nice for those (if any) that also use this so we dont need to fork.


haugene:
I agree, wget is a handy addition! :+1:
Thanks for contributing!

Issue #167, opened 2016-11-29
No port forwarding?


ouafnico:
Hi

Using PIA, it seems the port forwarding does not work anymore.

root@b5ef704dd9bc:/# /etc/transmission/updatePort.sh MYIP
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   149  100    57  100    92     66    106 --:--:-- --:--:-- --:--:--   106
Got new port  from pia
transmission auth not required
localhost:9091/transmission/rpc/ responded: "success"
Checking port...
Error: portTested: http error 400: Bad Request

Any idea how to fix?
It seems the URL https://www.privateinternetaccess.com/vpninfo/port_forward_assignment is not good anymore.

blakejohnson86:
While running that command, I got a new port assignment back, it does not automatically set it in Transmissioin anymore.

> root@4bab00549c2f:/# /etc/transmission/updatePort.sh IP
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
> 100   105  100    14  100    91      2     13  0:00:07  0:00:06  0:00:01    26
> Got new port 34*** from pia
> transmission auth required
> Unexpected response: <h1>301: Moved Permanently</h1>t: transmission-remote/2.92 (14714)
> Host: localhost:9091
> Accept: */*
> Accept-Encoding: deflate, gzip
> 
> Unexpected response: <h1>301: Moved Permanently</h1>t: transmission-remote/2.92 (14714)
> Host: localhost:9091
> Accept: */*
> Accept-Encoding: deflate, gzip
> 
> Checking port...
> Unexpected response: <h1>301: Moved Permanently</h1>t: transmission-remote/2.92 (14714)
> Host: localhost:9091
> Accept: */*
> Accept-Encoding: deflate, gzip

ouafnico:
It seems to work for me using France pia server. Using other does not work

rosshettel:
I'm seeing this issue too - not getting any port back from updatePort.sh

rosshettel:
Just tried the France PIA server as @ouafnico reported, and that worked fine

haugene:
Thanks to @rosshettel for providing a fix for this issue. The script is updated so that it doesn't mess with the transmission port configuration if the port call to PIA fails or times out.

[...]

PR #171, opened 2016-12-20
Added more robust error handling around checking for new port from PIA

rosshettel:
Hi, 

After experiencing port issues in #167, I tracked down what might be causing them. In my experience, the curl request to PIA would timeout, and the `updatePort.sh` script wouldn't handle that error properly, and proceed to set an empty port to transmission which resolved to port 0.

I added some additional error checks to make sure we've got a port from PIA before setting it, as well as fail if curl receives an HTTP error. I also made the curl downloading output silent, as it was cluttering up logs.

Hope this helps, please let me know if I should make any changes!

### DIFF
--- a/transmission/updatePort.sh
+++ b/transmission/updatePort.sh
@@ -31,10 +31,27 @@ if [ -z ${pia_client_id} ]; then
 fi
 
 # Get the port
-pia_response=$(curl -d "user=$pia_username&pass=$pia_passwd&client_id=$pia_client_id&local_ip=$local_vpn_ip" $port_assignment_url)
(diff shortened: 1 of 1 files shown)


rosshettel:
Thanks! I just pushed a really quick fix to actually log the curl exit code

haugene:
Agreed, good solution and code looks good. Thanks for helping out and contributing!

I usually prefer the pull requests to go to the dev branch for some testing before a merge to master. But this is an isolated fix and there are more people having this issue that can verify that it works, so I say let's merge :)

PS: If the PIA timeouts still pose a problem we could try increasing the timeout values of the curl command ([something like this](http://unix.stackexchange.com/questions/94604/does-curl-have-a-timeout/94612)). But let's see first. The logs will tell us more, and now they'll be uncluttered as well ;)

PR #181, opened 2017-01-06
Fix docker-compose errors and introduce image names

samthecodingman:
- Removed double quotes from the "OPENVPN_OPTS=" line due to literal interpretation that would cause an issue at /openvpn/start.sh:57 when executed.
- Fixed failing "proxy/build" instruction. Splitting into "context" and "dockerfile" subkeys prevents the error "TypeError: You must specify a directory to build in path"
- [Proposal] Introduced "image: haugene/rpi-transmission-openvpn*" lines to each docker-compose structure to transition to Docker Hub hosted images in the future. (see https://resin.io/blog/building-arm-containers-on-any-x86-machine-even-dockerhub/)

### DIFF
--- a/docker-compose-armhf.yml
+++ b/docker-compose-armhf.yml
@@ -5,6 +5,7 @@ services:
   build: 
      context: .
      dockerfile: Dockerfile.armhf
+  image: haugene/rpi-transmission-openvpn
   privileged: true
   restart: always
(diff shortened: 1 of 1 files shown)


haugene:
LGTM 👍 Thanks for contributing

PR #183, opened 2017-01-15
PureVPN config files actualization.

sl0n:
Hello,

I updated config files for PureVPN provider. There are now many more new servers added in different locations.


haugene:
Looks good. Thanks for contributing! :+1: Merged it to the dev tag now. Will add a couple more providers and merge it to master later.

PR #190, opened 2017-01-20
Added giganews config.

cassegrain2:
I have fixed vyprvpn for giganews users. There are in fact two ways to connect to vyprvpn:

1. With a GoldenFrog account, this is supported already with VYPRVPN setting (server looks like ***.vyprvpn.com)

2.With a giganews account, this did not work as the server is slightly different (***.giganews.com).


haugene:
Great stuff! Thanks for contributing :+1:

PR #219, opened 2017-03-06
Reduce the amount of privileges needed

aaronjwood:
Thought it'd be nice for various security reasons to reduce how many privileges this container has. Running with only the necessary privileges also makes it possible to use user namespaces (you need `--userns="host"` if you're using `--privileged`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ The container is available from the Docker registry and this is the simplest way
 To run the container use this command:
 
 ```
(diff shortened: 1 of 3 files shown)


haugene:
Hey. First, sorry for the late reply. There's always too much to do :)

Totally agree with your edits though, better and more secure. 
Thanks for contributing! :+1:

aaronjwood:
No problem, thanks for the merge! :)

PR #231, opened 2017-04-06
Added firewall support

ullbergm:
My VPN puts the client straight on the internet with no firewall, so I added UFW and configured it to block everything except the peer port from the internet. It will also allow the LOCAL_NETWORK (if configured) to access the rpc port as well as the container default gateway.

If transmission is configured for random ports it will open the whole range of potential ports.

I made the firewall is disabled by default so that it won't break existing setups but if this looks right perhaps it should default to on.

I have not tested this on ARM.


haugene:
Adding firewall rules is a great addition to the image, thanks for contributing!
Looking forward to merging this, but I want to do some testing etc to see how it behaves. And I agree that this should maybe default to on, just not sure if it should be optional for a while first to let people test it with different providers.

There's been more activity with this project than I anticipated, and finding the time to maintaining it is a challenge. Just wanted to let you know it's on the todo-list and hopefully I'll find time soon ;)

ullbergm:
Cool, i'm not surprised, its a well made container. :)

I think that once it has been tested then the firewall should be on by default, but i didn't want to do that until everything had been tested. It really shouldn't hurt anything but you never know.

I don't use PIA so I don't know how it will work with that provider since it seems to have some special port handling.

haugene:
So, finally got around to playing with this. It seems to behave like a charm! :)

As you mention there is some special port handling with PIA that needs changes in UFW. For PIA the peer port is changed after startup. I use PIA and have added the necessary rule updates for that now. Will add it after the merge.

I'm going to let UFW be disabled by default for now, push the code to mater, and hopefully change the default on the next release to master. Thanks again for contributing :+1:

PR #244, opened 2017-05-06
Add IPredator OpenVPN config

janschulz:
https://ipredator.se

Config files from the detailed overview page with all three linux confs, which you see when you are logged in. But they are also available from the Guides pages.

I changed three things:
* renamed the files and ending (.conf -> .ovpn)
* commented out: `+#keepalive 10 30` because the ping setting conflicts with it
* commented out: `+#tls-version-min 1.2` because the openvpn version is too old :-/

### DIFF
--- a/openvpn/ipredator/default.ovpn
+++ b/openvpn/ipredator/default.ovpn
@@ -1,0 +1,89 @@
+# VER: 0.25
+client
+dev tun0
+proto udp
+remote pw.openvpn.ipredator.se 1194
+remote pw.openvpn.ipredator.me 1194
+remote pw.openvpn.ipredator.es 1194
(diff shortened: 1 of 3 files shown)


haugene:
Great! I'll merge it to dev to test it first, also want to add it in the providers list in the README.
Thanks for contributing, and sorry about the wait!

haugene:
Hmm. I was a little quick there. Seems IPredator is already added as a provider on the dev branch.
Can you try running the dev tag of the container image and see if that works for you?

This is a symptom that I need to get the dev stuff into the "latest" image tag, and tag a new release. If you want to add other configs than the default.ovpn already on the dev branch, please rebase your branch and open a new PR.

PR #246, opened 2017-05-11
Update HMA server list.

romain38:
Server list comes from https://www.hidemyass.com/vpn-config/vpn-configs.zip.


haugene:
LGTM 👍 
Thanks for contributing and sorry about the wait!

PR #253, opened 2017-05-25
update nordvpn config files

rmorais:
I created this on a mac machine so not sure if there is any extra checks to do.
Also I think the default.ovpn is pointing to a missing file.
I couldn't find the proper way to update it on a mac


haugene:
Hey! No worries about the default.ovpn file, I'll fix that. Merging now and doing a quick check that everything works.

Thanks for contributing! 😄

Issue #109, opened 2016-07-28
Add HTTP proxy support


atmosx:
I have a similar setup on an RPi. It's a great idea to dockerize this. Since we're routing all the traffic through VPN it would be a good idea to add a lightweight HTTP proxy.

haugene:
The readme describes how to use nginx as a proxy and a preconfigured image is provided as well: 
https://hub.docker.com/r/haugene/transmission-openvpn-proxy/

Dockerfiles are available both for x64 and RPi:
https://github.com/haugene/docker-transmission-openvpn/tree/master/proxy

What exactly are you looking for here?

ouafnico:
Maybe he means a Squid proxy on the docker? It could be great!

atmosx:
Exactly. I mean an HTTP proxy, not a reverse proxy. I had something like privoxy/polipo in mind, that will allow users to also browse with some privacy - except from downloading torrents.

haugene:
Aha! Yes, that would be a neat feature :+1: 

Could configure/activate it using environment variables as all the other features, or provide it on a separate tag. Anyways, a good suggestion!

haugene:
Gotta love Docker! I was playing around with this and figured it should be possible to make something from the container networking mode. And indeed it is :smile: 

You can make any container join the network stack of another container by specifying --network container:<name|id> (or --net=container:<name|id> for older docker versions). See the [run reference](https://docs.docker.com/engine/reference/run/#/network-container).

So if you start the transmission-openvpn container and also specify a port bind for privoxy (e.g. -p 8118:8118) then it will be set up in the stack and you can find some privoxy image on hub and just hook it in to your existing network.

I testet with a couple of images and it worked like a charm! Here's one with tor and privoxy:
docker run -d --name tor-privoxy --net=container:<name|id> rdsubhas/tor-privoxy-alpine

Run that one, and it should be good to go :+1: But remember to bind the ports you need when starting the openvpn container. And also set the LOCAL_NETWORK environment variable to allow the proxied traffic to "escape" the container.

Let me know how it goes. Should probably include a section in the README on this as well...

haugene:
I'll conclude that my previous answer is the recommended way to use a proxy with this container and close this issue. It leverages the docker platform instead of creating more custom components, introduces flexibility in which proxy to follows separation of concerns principle. It makes it all more maintainable to not bundle more running services into this container.

PR #256, opened 2017-06-09
Expose openvpn proxy via a webproxy

jandrews377:
Some private trackers need to see your browser traffic come from the same IP as your torrent client. This change allows you to use the container as a web proxy tunnelling through the same connection.


haugene:
Hey. First of all, thanks for contributing! 

This could be a cool addition to the image and have been requested earlier in #109 for example.
Just a couple of questions. Should this be a opt-in feature? That it should only be activated if some environment variable is set, like the firewall feature?

Could you also update the README on how to use this? And does it work with the mentioned firewall opt-in? Should maybe have a `ufw allow` statement in the start script if ufw is enabled.

Thoughts?

jandrews377:
Hi Haugene.
Sorry I am just getting started with github and git, im struggling a bit :)
I have just committed my latest round of changes, hopefully they are automatically visible on this pull request. It now is controllable via env vars and there is documentation. Any feedback and suggestions is appreciated. I dont know about the firewall stuff, ill have to test that.

niXta1:
In: "11  openvpn/start.sh"
It says: "+    # Alway default back to port 8888"
Alway --> Always

michaelvandonselaar:
Is this going to be merged into master?

haugene:
Stalled this for a while, sorry about that. I wrote a little about this feature in #314 but I'll try to summarize it here. First of all I've been unsure if this should be a part of this image as I generally want my containers to be single-purpose and you can find a proxy container and start it with `--net=container:<transmission-openvpn-container>` and the traffic will then go through VPN.

**BUT:** This is a neat feature that has been proposed before and that I think a lot of people would appreciate, especially those that aren't that experienced with Docker. So I'm positive to adding it.

There are a couple of things we need to do before it can be merged.
- The tinyproxy service should be started and stopped in the same way as transmission. That is, it should be started when the VPN connection is established, and killed if the connection goes down to ensure that your traffic is actually proxied through the vpn.
- If the proxy is enabled, and the UFW option is enabled we need to open that port in the

So we'll need a common script that starts transmission, and the proxy if enabled, and pass it to the --up directive. Same goes for killing the services on --down.

I won't be able to do it myself for the time being, but if anyone has the time I'll be ready to review and merge it. Don't know if @jandrews377 is still active and up for making these changes. Or maybe @wbvreeuwijk or @pandrez can have a look?

jandrews377:
Hi haugene, I think you are correct in pointing out that a seperate proxy container could be used and, imo, would better reflect how docker should be leveraged. I just havent tested that to ensure that I can access the transmission interface LAN-side. As far as the bullet list is concerned, I thought this is how I have implemented the proxy :\ If someone could peer review my pull request and provide feedback that would be appreciated. I have been using my branch for nearly a year. Now and again the VPN connection is lost and traffic just stops flowing. I simply restart the container and it works again. This happens about once a week. If there is a way to completely stop the container if the VPN goes down, that would be better. If that cannot be determined, then maybe a seperate proxy container is the correct solution.

haugene:
Hey @jandrews377, nice to see you're following up!

Right now you're starting the tinyproxy in the openvpn start script. Transmission is not started with openvpn, but instead openvpn starts and stops it on [tunnel up and down](https://github.com/haugene/docker-transmission-openvpn/blob/master/openvpn/start.sh#L47). That line should be changed to somethig like `--up /etc/openvpn/tunnelUp.sh --down /etc/openvpn/tunnelDown.sh`, naming shit is hard :)

Then those scripts need to be added and call transmission/start.sh and tinyproxy start for the up script and some kill/stop statements in the down script. And I'll do a review and add some other comments regarding UFW.

docker-compose.yml
```diff
@@ -21,7 +22,8 @@ services:
     - OPENVPN_PASSWORD=password
     - OPENVPN_OPTS="--inactive 3600 --ping 10 --ping-exit 60"
     - LOCAL_NETWORK=192.168.0.0/24
-
+    - WEBPROXY_ENABLED=true
```
haugene: Since this is a opt-in feature, I think we should remove it from the default docker-compose and add description in README that the port needs to be opened as well.

openvpn/start.sh
```diff
@@ -78,4 +78,15 @@ if [ -n "${LOCAL_NETWORK-}" ]; then
   fi
 fi
 
+
+if [ "${WEBPROXY_ENABLED}" = "true" ]; then
+  if [ -z "$WEBPROXY_PORT" ] ; then
+    /opt/tinyproxy/setport.sh $WEBPROXY_PORT
```
haugene: This whole addition needs to be moved to the new up script. And in addition we need a section [like this](https://github.com/jandrews377/docker-transmission-openvpn/blob/0877052a13315791a8ece76d3036f36567db17e1/openvpn/start.sh#L74) to open port in UFW if both UFW and proxy is enabled.

docker-compose.yml
jandrews377: Thanks for looking at this. I am keen to dispense with my custom branch, I'll have a crack at this and test this weekend.

haugene: Great! One more thing I got to think of. You asked if there was a way to make the container go down when the vpn dropped. It should do that if you set the `OPENVPN_OPTS="--inactive 3600 --ping 10 --ping-exit 60"` option. That should exit the openvpn process when it's not able to ping anymore, and setting `--restart=always` or `--restart=unless-stopped` should make the container restart itself.

I've seen that config for some providers have the `ping-restart` option set. That should probably be removed to ensure that it actually shuts down.

jandrews377:
Hi haugene, I have pushed up all my changes for review. I have tested locally but am having issues using the proxy from my PC (external from my docker host). Testing using telnet on the docker host works fine. I am uncomfortable about how we configure the WEBPROXY_PORT as this port needs to be exposed in the Dockerfile (i.e. EXPOSE 8888). Maybe it cant be made configurable afterall?

TMaddox:
@haugene when will this be added? really looking forward to this feature

haugene:
Long overdue, but this is finally added to the dev branch. I've done some modifications to make it work with the UFW feature. Would be great if people could test it on the dev tag and let me know if it works :)

jandrews377:
Hi  haugene, ill start using this as my daily-driver. I appreciate this is a labour-of-love for you, thanks for your hardwork :)

haugene:
Great. And thank you for contributing with this feature, it has been requested several times in the past. I got chrome running through the proxy locally, but traffic was quite slow. I had some "waiting for proxy tunnel" messages on my mac, but it worked so I'm happy for now. Could seem to be related to ipv6, but that's for another day. Let me know if you experience something similar.

h45rd:
Only did a short test, but seems to be working quite nicely for me (UFW enabled). Thanks!

Mossop:
Seems to work fine here

PR #264, opened 2017-06-29
Display WAN IP

AustinSaintAubin:
Display WAN IP in Console

### DIFF
--- a/openvpn/start.sh
+++ b/openvpn/start.sh
@@ -78,4 +78,7 @@ if [ -n "${LOCAL_NETWORK-}" ]; then
   fi
 fi
 
+echo WAN IPv4: $(curl --silent --max-time 2 http://ipv4.plain-text-ip.com)
+echo WAN IPv6: $(curl --silent --max-time 2 http://ipv6.plain-text-ip.com)
+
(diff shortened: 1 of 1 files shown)


AustinSaintAubin:
This addition might be best to do to ( /etc/transmission/updatePort.sh )

haugene:
Hi, and thanks for contributing to the container!

Just so I understand the reason for this change. You want to know the IP you're on before connecting to VPN? OpenVPN starts on the next line, so you won't see that if this is executed before.

AustinSaintAubin:
I want to know my WAN ip before and after VPN connection is established. In a newer commit on my container I put the display wan ip command in one of the transmission shell scripts so it's the last to run.

haugene:
Wanna update this PR with that? So we get both IP's posted in the logs?

himred:
This can be done with the new "Custom pre/post scripts" feature.

- Mount /scripts
- Create a file named /scripts/transmission-post-start.sh
- Make the script executable using : `chmod a+x /scripts/transmission-post-start.sh`
- Edit the script and put this code inside:

```
#!/bin/bash
ip=$(curl -s 'https://api.ipify.org?format=json' | cut -f4 -d'"')
echo "Current WAN IP is : $ip"
```

You can read the documentation about Custom pre/post scripts for details.

AustinSaintAubin:
Awesome, finally tested at that works!

PR #273, opened 2017-07-12
NordVPN Update

davestephens:
Spent an hour trying to work out why I couldn't connect to a specific host, then realised that a lot of the configs are out of date! So, I've:
* Updated NordVPN with latest configs
* Created a NordVPN config update script for future updates
* Updated the default NordVPN config


haugene:
Sweet! Thanks for contributing :) Merging now and testing locally, will probably make a new release to master branch soon. But for now it will be available on the dev tag.

PR #274, opened 2017-07-13
Ok, upgraded to 16.04,

gflarity:
Pretty straight forward.  I did notice that transmission is storing/looking for stuff in a slight different place. Pretty sure this a transmission change between versions.

Thanks again for publishing this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,28 +2,32 @@
 #
 # Version 1.15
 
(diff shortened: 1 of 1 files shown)


haugene:
Great! Thanks for contributing, will test it locally and update transmission settings as I think there's some new setting for this version.

Issue #236, opened 2017-04-20
Starting from docker-compose


jagorski2:
Using the default docker-compose.yml file and current repo.

When running 
    docker-compose up

I get:
transmission_1  | Using OpenVPN provider: PIA
transmission_1  | No VPN configuration provided. Using default.
transmission_1  | Setting OPENVPN credentials...
transmission_1  | adding route to local network 192.168.0.0/24 via 172.18.0.1 dev eth0
transmission_1  | Options error: the --down directive should have at most 1 parameter.  To pass a list of arguments as one of the parameters, try enclosing them in double quotes ("").

Any ideas?

rourke:
Change
        - OPENVPN_OPTS="--inactive 3600 --ping 10 --ping-exit 60"
to
        - OPENVPN_OPTS=--inactive 3600 --ping 10 --ping-exit 60

That worked for me.

haugene:
Any luck?

byronmejia:
Just had same issue as @jagorski2 and can confirm following @rourke advice worked in my case.

Running Moby on Windows 10 in my situation.

byronmejia:
My compose file, for reference:

```yaml
version: "3"
services:
  transmission:
    container_name: transmission
    privileged: true
    restart: always
    image: haugene/transmission-openvpn
    environment:
      - OPENVPN_PROVIDER=PIA
      - OPENVPN_CONFIG=Netherlands
      - OPENVPN_OPTS=--inactive 3600 --ping 10 --ping-exit 60
      - LOCAL_NETWORK=192.168.1.0/24
    volumes:
      - ./transmission:/data
    env_file:
      - transmission.env
    dns:
      - 8.8.8.8
      - 8.8.4.4
  transmission-proxy:
    container_name: transmission-proxy
    image: haugene/transmission-openvpn-proxy
    ports:
      - "8080:8080"
    links:
      - transmission:transmission
```

jordan-heemskerk:
Ran into this as well. @rourke solution fixed it up for me. I'm on Windows 10. Can anyone using docker on other platforms confirm whether this is a Windows specific issue? Otherwise it'd be nice to have a fix in for the default docker-compose file.

rourke:
It was a problem for me on Ubuntu Server 16.04 LTS.

jordan-heemskerk:
Seems like the quotes should be gone by default then

haugene:
Merged the fix, thanks @jordan-heemskerk

PR #276, opened 2017-07-16
docker-compose.yml: remove quotes

jordan-heemskerk:
Would be great if this could be tested on a couple other platforms. This change is required for docker on Windows 10. 

Addresses #236


sickill:
I can confirm this fixes it on Linux also (tested on Ubuntu 16.04).

haugene:
You're right, that's the correct syntax now. Worked before on older docker daemon, but I get the same error here now.

Thanks for contributing and sorry for the delay!

PR #282, opened 2017-07-29
Add support for VPNArea.com

packruler:
Using config files found in http://vpnarea.com/OpenVPN-ubuntu.zip


haugene:
Hi! First of all, sorry for taking so long before merging this. Have put this project away for a while with a growing bad conscience.

Great contribution, thanks! Seems you've added perfect privacy as well but not updated the README on that. You meant to push them both, right? :) I'll update the README for now. Thanks again.

PR #326, opened 2017-10-26
Support transmission-rss plugin

bperel:
This PR adds the ability to use the Transmission RSS plugin (https://github.com/nning/transmission-rss) which downloads torrents using a monitored RSS feed URL. 

- It requires ruby, gcc and make to compile the plugin itself (I'm open to suggestions for a more lightweight install).
- I wasn't sure about in what path to put `transmission-rss.conf`.

Feel free to comment and suggest changes as I don't know this project very well :-)


haugene:
Hey! Thanks for contributing - this looks like a nice feature to have.
Then there's the question of how to do it "right" in this container. And I'm not sure either on what is the best way to go.

First, the transmission-rss.conf. We're using dockerize for templating and we should consider using it for this conf file as well. Basically save the file as transmission-rss.tmpl and then do a `dockerize -template /etc/transmission/transmission-rss.tmpl:/some/path/transmission-rss.tmpl`. The `/some/path/` could either be TRANSMISSION_HOME as with settings.json or in the /config directory where we store openvpn and transmission credentials. Then you could reference the RSS_URL and TRANSMISSION_DOWNLOAD_DIR environment variables to set it correctly depending on the user config.

Then there's the issue of a lightweight install. We could make this RSS feed as a sibling container. So create a new folder in the root of the project, with a separate Dockerfile that only installs this plugin. Linking the RSS container to the transmission-container it can still communicate with transmission, but we keep the image size smaller. It adds a step to the setup of the RSS feed as you need another container, but it would separate it and thus reduce complexity in the main image.

Thoughts?

bperel:
Hi,

Good ideas :+1: I followed your advice and created a separate Dockerfile.

Worth noting:
* I respectively created and updated the armhf-specific Dockerfile and docker-compose files but couldn't test them by lack of dedicated hardware
* If the user wants to use the plugin, he has to specify the `TRANSMISSION_DOWNLOAD_DIR` variable when running the plugin's container in addition to specifying it in the base container. Maybe there's a way to not need to do that, but apart from using `extends` in the Dockerfile to inherit the base container's environment variable I don't see an alternative to that.

What do you think?

haugene:
Had a quick look now, and it looks good! We don't need the RSS_URL in DockerEnv anymore though?
Will have a closer look soon and then I'm looking forward to merge it :)

bperel:
In the README it is mentioned that the DockerEnv file contains "all the current transmission and openvpn environment variables", shouldn't it include the RSS plugin ones as well? I don't have a strong opinion on this, so I'm ok to remove it if you prefer that.

haugene:
Yeah it should contain all envs, but for the openvpn-container. Or at least that was the intention.
I've been super slow here, and your updates looks great, so I'll just merge it now and remove that one line.

Thanks again for a great contribution. Now it's christmas, which means some free time to hack on this project ;) I'll do a merge to master before new years I guess. Happy holidays!

bperel:
Remember to push the `haugene/transmission-rss` image to Docker Hub :-)

haugene:
On it ;) https://hub.docker.com/r/haugene/transmission-rss/tags/
Doing a build from dev branch to latest tag now, until it's merged.

Also, if a README is placed in the plugins/rss folder we would have a better description for the image: https://hub.docker.com/r/haugene/transmission-rss/
It can be used with any transmission container after all :)

And one more thing I came to think of. For people using authentication on transmission-rpc it won't work. The rpc passwords are persisted to the /config volume besides openvpn-credentials.txt so it should be possible to link the rss-plugin container and make it read that file if present. As a future enhancement :)

jdk:
I just did a pull request #448 to update the documentation. The previous wasn't correct, unless I'm missing something.

Also, regarding authentication. I assumed you meant pulling the password from the settings.json file. I premounted the shared directory where that lives, but it looks like I was incorrect. There is a /config directory where it is in clear text.

Curious, why not just use environmental variables: 
TRANSMISSION_RPC_PASSWORD and TRANSMISSION_RPC_USERNAME

then transmission-rss.tmpl becomes:
```
++login:
++  username: {{ .Env.TRANSMISSION_RPC_USERNAME }}
++  password: {{ .Env.TRANSMISSION_RPC_PASSWORD }}
```

Also, quick question, transmission RSS supports more than one feeds url. How is that accomplished using our single environmental variable RSS_URL?

haugene:
Merged the documentation update.

For the authentication, yes, they're encdoded(base64?) or encrypted somehow in settings.json and you need to look in /config to find the clear text ones. One could use environment variables for now. But that is a feature of --link, and that's [deprecated](https://docs.docker.com/network/links/).

As for more than one rss feed url. I'm not sure. How is the plugin handling it? If it accepts comma separated feeds I guess that would solve it. Or else we need to extend the rss plugin image to support it.

PR #331, opened 2017-10-31
VPN configuration for anonvpn

printgeek:
added 7 profiles for Anonvpn.io. I originally was given these profiles as part of a help desk response, so I'm not sure if they are public knowledge.


haugene:
Looks good, thanks for contributing! Missing entry in the provider list in the README, but I'll fix that :)

printgeek:
Awesome. My first official contribution to any project. 

Thank you

PR #332, opened 2017-11-05
Added smartdnsproxy vpn provider

choldi:
Dockerized and tested with some countries for try, working ok


haugene:
Great stuff, LGTM :+1: 
Thanks for contributing!

PR #337, opened 2017-11-08
Add VPN.AC to providers

pandrez:
Added [https://vpn.ac/](url) to providers


haugene:
Looks good, thanks for contributing!

PR #343, opened 2017-11-16
updating frostVPN files

g-be:
I've updated the .ovpn files to match the format provided by frostVPN, and generated an .ovpn file for each server/gateway, with a UDP and TCP version


haugene:
Verified that frostVPN was not working, merged your PR and updated default.ovpn symlink and it seems to work again. Thanks for your contribution! :+1:

PR #346, opened 2017-11-19
Updated NordVPN OpenVPN config files

levifig:
Hey,

Just noticed something up with my NordVPN connections today, and the logs were telling me there was an issue with the TLS handshake. When I searched past issues, I noticed there was a similar one that was fixed by updated config files. Sure enough, NordVPN released new config files today, probably updating the TLS certificates.

Of note, NordVPN removed the port numbers from the configuration file names. I for one appreciate the change BUT… this will mean people will have to update their Docker environment variables for the new configuration paths. I'm choosing to not append the port numbers to make the config files backward compatible because 1) I think this is a good move and 2) seems like this is the way they're choosing to go forward and having to append the port numbers every time there's an update is probably not a wise choice. :)

Cheers! o/


levifig:
(sorry for the spam: the UI wasn't updating after I pushed the latest commit… 🤦‍♂️)

haugene:
Thanks for the quick update, and for contributing!

I agree with the naming of the ovpn files, better to keep the default names from NordVPN even if that means people have to change their config.

There is one thing though. I see that the NordVPN configs declare values for `ping`and `ping-restart`. For this container, because of being in a Docker container, we've generally seen that it's a better approach to set `ping` and `ping-exit` so that the openvpn process exits if no connection and then we start the container with `restart=always` so that the container itself is restarted when the process dies. This is a more thorough restart, as the networking of the container is set up from scratch and that also fixes docker networking problems if any.

The default `docker run` command from the readme sets these options, and when also setting `ping-restart` in the config files I get a warning during startup. Should we either remove the ping options from the config files, or set the `ping-exit` instead of `ping-restart`?

levifig:
Absolutely!! I'll make the change and update the PR! Good call! 👍

levifig:
Updated config files. Let me know if there's anything else I should change… :) 👍

haugene:
Was just reviewing it now actually. Looks good! I want to try and merge one more PR to dev before doing a merge to master and creating a new release tag. But if I don't find the time within a couple of days I'll just merge it and postpone the other one.

Thanks again for contributing and following up on your PR! :)

PR #347, opened 2017-11-19
Command line for synology: add --sysctl

Stibbons:
To avoid issue with latest version of Synology's Docker package 17.05.0


haugene:
Thanks for the update 👍

PR #356, opened 2017-11-24
Update NordVPN ovpn files to newest

karl-run:
I noticed there were some missing ovpn files when I tried to connect to some servers that kept failing. :) I've just pulled the newest files, I did the same changes as discussed in #346. Seems some of the certs have been updated which should be reflected in this PR as well.


haugene:
Looks correct to me, not easy reviewing a bulk of ovpn files :P
Thanks for contributing!

karl-run:
They should be good. It was a bunch of `sed`-ing until the existing files had no diff - then I can be pretty sure the new ones are in the correct format as well. :)

haugene:
Aha. Yes, then it should be all good ;)

PR #363, opened 2017-11-30
restart openvpn if the up script is executed and no ipv4 is bound 

bkmartin:
During testing with PureVPN, there are some situations where the up script is executed when a VPN tunnel is not actually established.  See log sample below.   This script update will kill the openvpn and have the tunnel re-established when this occurs rather then starting transmission with no VPN tunnel established

```
Thu Nov 30 14:55:38 2017 TUN/TAP device tun0 opened
Thu Nov 30 14:55:38 2017 /etc/transmission/start.sh tun0 1500 1558   init
Up script executed with tun0 1500 1558   init
ERROR, unable to obtain tunnel address
```


haugene:
Good catch, and looks like a good fix to me. Merging to dev for some testing and it will be included in the next release.

Thanks for contributing!

PR #377, opened 2018-01-01
Allow a pre-existing /config/openvpn-credentials.txt file.

ksanislo:
Instead of bailing when the OpenVPN user and password aren't defined in the environment, we should instead check and see if we already have an existing /config/openvpn-credentials.txt file from a persistent volume that can be used first.

This is important because --link will share environment variables with other containers, and it would be preferable to not need to include external auth information with that.


haugene:
Good point with the shared environment, should be able to provide it manually with files in persistent volume. Thanks for contributing!

PR #384, opened 2018-01-08
Updating IPVanish configurations

izeberq:


haugene:
Great, looks good to me! I've moved, and changed job, so haven't found the time for this project in a while. Will try to make a new release from dev soon. But until then, there's always the dev tag 😄 

Thanks for contributing!

PR #386, opened 2018-01-11
Modify Netherlands OVPN config for SlickVPN

shuaiscott:
changing ports to 8888


haugene:
Changed base to dev. Looks good. Thanks for contributing!

PR #387, opened 2018-01-11
Modifying SlickVPN config for Toronto

shuaiscott:
Trying out adding all hostnames AND additional ports

### DIFF
--- a/openvpn/slickvpn/Canada-Toronto.ovpn
+++ b/openvpn/slickvpn/Canada-Toronto.ovpn
@@ -1,4 +1,14 @@
+remote gw1.yyz1.slickvpn.com 8888 udp
+remote gw2.yyz1.slickvpn.com 8888 udp
+remote gw3.yyz1.slickvpn.com 8888 udp
+remote gw4.yyz1.slickvpn.com 8888 udp
+remote gw1.yyz2.slickvpn.com 8888 udp
+remote gw1.yyz1.slickvpn.com 443 udp
+remote gw2.yyz1.slickvpn.com 443 udp
 remote gw3.yyz1.slickvpn.com 443 udp
+remote gw4.yyz1.slickvpn.com 443 udp
+remote gw1.yyz2.slickvpn.com 443 udp
+
 auth-user-pass /config/openvpn-credentials.txt
 client
 redirect-gateway


haugene:
Looking good, thanks! :D

maikito26:
Question - How is this supposed to work?  I set this up last night but couldn't get Toronto to work.  Looks like it was wanting port 8888 but it kept defaulting to 443, then timing out for inactivity.  Ultimately I switched to UK and it connected fine, though I'd prefer Toronto.   

Is there a way to force a gateway & port in this file?  Or should the application be retrying the next gateway in the list if it fails?

The docker log on the Synology NAS shows this:

```
2/1/2018 5:53 | stdout | Using OpenVPN provider: SLICKVPN
2/1/2018 5:53 | stdout | Starting OpenVPN using config Canada-Toronto.ovpn
2/1/2018 5:53 | stdout | Setting OPENVPN credentials...
2/1/2018 5:53 | stdout | adding route to local network x.x.x.x/24 via 172.17.0.1 dev eth0
2/1/2018 5:53 | stdout | Thu Feb  1 05:53:54 2018 OpenVPN   2.4.4 x86_64-pc-linux-gnu [SSL (OpenSSL)] [LZO] [LZ4] [EPOLL] [PKCS11]   [MH/PKTINFO] [AEAD] built on Sep 26 2017
2/1/2018 5:53 | stdout | Thu Feb  1 05:53:54 2018 library versions: OpenSSL   1.0.2g  1 Mar 2016, LZO 2.08
2/1/2018 5:53 | stdout | Thu Feb  1 05:53:54 2018 NOTE: the   current --script-security setting may allow this configuration to call   user-defined scripts
2/1/2018 5:53 | stdout | Thu Feb  1 05:53:54 2018 TCP/UDP:   Preserving recently used remote address: [AF_INET]198.144.157.100:443
2/1/2018 5:53 | stdout | Thu Feb  1 05:53:54 2018 UDP link   local: (not bound)
2/1/2018 5:53 | stdout | Thu Feb    1 05:53:54 2018 UDP link remote: [AF_INET]198.144.157.100:443
2/1/2018 5:54 | stdout | Thu Feb  1 05:54:54 2018 [UNDEF] Inactivity timeout   (--ping-exit), exiting
2/1/2018 5:54 | stdout | Thu Feb  1 05:54:54 2018 SIGTERM[soft,ping-exit]   received, process exiting
```

Just curious if there's a step/setting I'm not yet aware of to help me configure this?  If not, I was thinking of proposing addition ovpn configs for SlickVPN... such as Toronto_Port8888 and Toronto_Port443 or something like that.  

Thoughts?

shuaiscott:
Yeah, if I remember correctly, it should round-robin through the list of servers. For some reason, it looks like it’s not doing that. We could separate the configs by ports, but it’s likely that those hosts will go down again sometime in the future. For example, 8888 will work now and 443 doesn’t, but later on it might switch. We might want to see why OpenVPN isn’t reverting to another server on the list.

In the meantime, if you wanted to fork the project, add those configs like you mentioned, and add a pull request, that should get you the ports you desired.

maikito26:
Your response made me go reading...  So I found this which probably explains why it's not.  Not sure how to confirm or troubleshoot:

Reference: https://openvpn.net/index.php/open-source/documentation/manuals/65-openvpn-20x-manpage.html

> Note the following corner case: If you use multiple --remote options, AND you are dropping root privileges on the client with --user and/or --group, AND the client is running a non-Windows OS, if the client needs to switch to a different server, and that server pushes back different TUN/TAP or route settings, the client may lack the necessary privileges to close and reopen the TUN/TAP interface. This could cause the client to exit with a fatal error.

So I tried two things.  I include the in the environment variable OPENVPN_OPTS:

`--remote-random`
This didn't work as expected.  Each restart of the service resolved the same remote gateway and port.  Sad panda.

`--remote gw1.yyz1.slickvpn.com --rport 8888`
This worked as expected and connected immediately.  

So this seems like a band-aid which will work pretty well.  But what you've proposed is technically better.  That same page I referenced does indicate that it should loop through the remote gateways in the ovpn configuration so it should work.  I'll need to see how I can troubleshoot this and find out if Synology can be made to work per design.

maikito26:
Also, for that particular port 443 is working.  Seems it was defaulting always to gw3.yyz1 for me.  gw1 is where it's at. :)

PR #394, opened 2018-01-16
add option to drop the default route on providers, that do not override it properly

T4cC0re:
I discovered that with some VPN providers (e.g. Mullvad) the routes have a potential issue.
This addresses this issue by allowing the user to drop the default route after OpenVPN has connected, but before Transmission is started.

Before:
```
root@62a71b505acd:/# ip r
0.0.0.0/1 via 10.8.0.1 dev tun0
default via 172.19.0.1 dev eth0
10.0.0.0/8 via 172.19.0.1 dev eth0
10.8.0.0/16 dev tun0  proto kernel  scope link  src 10.8.0.6
82.103.140.213 via 172.19.0.1 dev eth0
128.0.0.0/1 via 10.8.0.1 dev tun0
172.19.0.0/16 dev eth0  proto kernel  scope link  src 172.19.0.4

```

After (with `DROP_DEFAULT_ROUTE`=`true`:
```
root@e9150d61d626:/# ip r
0.0.0.0/1 via 10.8.0.1 dev tun0
10.0.0.0/8 via 172.19.0.1 dev eth0
10.8.0.0/16 dev tun0  proto kernel  scope link  src 10.8.0.8
128.0.0.0/1 via 10.8.0.1 dev tun0
134.90.149.138 via 172.19.0.1 dev eth0
172.19.0.0/16 dev eth0  proto kernel  scope link  src 172.19.0.4
```

In both cases `10.0.0.0/8` is my `LOCAL_NETWORK`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -367,3 +367,12 @@ $ sudo systemctl stop transmission-openvpn.service
 # Later ...
 $ sudo systemctl start transmission-openvpn.service
 ```
+
+## Drop the default route
+
+Some VPNs do not override the default route, but rather set other routes with a lower metric.  
+This might lead to te default route (your untunneled connection) to be used.
+
+To drop the default route set the environment variable `DROP_DEFAULT_ROUTE` to `true`.
+
+*Note*: This is not compatible with all VPNs. Please check yourself if your provider overrides the default route properly.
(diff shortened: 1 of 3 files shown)


haugene:
Good catch, and the change looks good. For the README, I think it can be higher up. This is related to the security of the container which most people feel is quite important.

Maybe put the readme-section up with the other environment variable descriptions? Between "User configuration options" and "RSS Plugin"? (as of now).

Also. Last sentence "Please check yourself if your provider overrides the default route properly.".
Maybe not "check yourself", but "check for yourself". Or just: "Please check if your provider overrides the default route property before using this option".

That being said, if the provider have already removed it - you'd hit the `exit 1` and that would be it. So that's a strong hint :)

haugene:
Doesn't matter, I'll fix the readme. Thanks for contributing!

T4cC0re:
Ah, sorry. Totally forgot it :/ Thanks!

PR #395, opened 2018-01-18
Use CVE-patched packages

aerickson:
Proof of concept. Uses patched packages per https://usn.ubuntu.com/usn/usn-3533-1/ vs the PPAs (not updated yet). I tried moving to 17.10, but openvpn complains about initscripts not being installable.

Also, fixes issues with 'True' vs 'true' being in settings.json... not sure when that started.

---

```
root@46ce04091d8f:/# dpkg --list | grep transmission
ii  transmission-cli            2.84-3ubuntu3.1                       amd64        lightweight BitTorrent client (command line programs)
ii  transmission-common         2.84-3ubuntu3.1                       all          lightweight BitTorrent client (common files)
ii  transmission-daemon         2.84-3ubuntu3.1                       amd64        lightweight BitTorrent client (daemon)
```


Dockerfile
```diff
@@ -8,7 +8,7 @@ VOLUME /config
 RUN apt-get update \
     && apt-get -y upgrade \
     && apt-get -y install software-properties-common wget git \
-    && add-apt-repository ppa:transmissionbt/ppa \
+#    && add-apt-repository ppa:transmissionbt/ppa \
```
haugene: Why comment instead of remove? Should we move back to the ppa packages afterwards? Haven't really gotten into the sources for transmission and what versions are available from ubuntu repositories.

haugene:
Hi. Changes looks good, I just have one comment. If you could enlighten me on that one, I'll merge.
And thanks for contributing!

aerickson:
Yeah, I left it in because we might want to switch back once the PPA has been fixed. The official Ubuntu debs were on 2.84 and the PPA was at 2.92. 

https://launchpad.net/~transmissionbt/+archive/ubuntu/ppa

Thanks. :)

nicbou:
Woah there! It would have been nice to be notified when the whitelist was added, as this was not the case before. Took me quite a bit of research to find out why it suddenly broke!

haugene:
woah indeed. was a bit quick there, the witelist should be disabled by default.

nicbou:
@haugene, the previous default for this docker image was false. The result was that updating to this image suddenly made the web UI return a 421 error code that's only visible if you open the Chrome web inspector.

haugene:
There was no previous default, this is a [new config value](https://github.com/transmission/transmission/wiki/Editing-Configuration-Files). But the rpc-whitelist-enabled was false by default, that's right. And it also has "enabled=true" as it's default from transmission. Question is if we should have other defaults than transmission sets. And if not (that we should also set them to true), what will be the correct values for the whitelists?

PR #397, opened 2018-01-19
Add a Dockerfile to build an Alpine based version

e7d:
Hey Kristian,

Here is a pull request that will let you build an Alpine based version of your image.
The result is running exactly the same as your Ubuntu based version, without the heavy weight of the final image. We're talking ~90MB instead of 361MB.

You'll notice I commented out the `MAINTAINER` section, as I don"t know if you will keep this updated or not.


haugene:
Hey! This looks great! Light weight distros are awesome for containers and I've just been stuck on "as long as it works" for this image. Not sure if I'll be able to maintain it either, but I've never done anything with the armf image either but it keeps getting updated so I have hope ;)

I also like the simplification of the stop script and generalization of start. Thanks for contributing!

haugene:
Oh yeah... And I'll set up an automated build

Issue #392, opened 2018-01-16
Upgrade to 2.93 ASAP to prevent DNS rebinding attack


aaronjwood:
https://github.com/transmission/transmission/pull/468

Hopefully 2.93 will be released soon...

nebhead:
Might I also suggest adding details on how to set the RPC login and password in the readme, after this has been patched.

aerickson:
Ubuntu has a patched version available.

https://github.com/haugene/docker-transmission-openvpn/pull/395

Mikiya83:
Do you have any news about a release with the 2.93 update ? It's a major security update :/

haugene:
FYI. The software source change made in #395 is now reverted due to errors #418. You can still run the patched ubuntu version from tag 1.20, but we're still waiting for the ppa to have a patched version and when it is released we need to add README instructions if configuration breaks the default run command.

haugene:
2.93 is now released (https://launchpad.net/~transmissionbt/+archive/ubuntu/ppa). Activated the ppa repository and pushed to dev.

nebhead:
Just a suggestion here, but can we add instructions to the dockerhub on how to setup a username/password for remote access so that we can fully mitigate the security issue here?

haugene:
@nebhead, you mean rpc authentication? If so, check the gitter chat.
Fix is now merged to master and the latest tag is on Transmission version 2.93, closing this.

PR #401, opened 2018-01-21
Upgrade armhf image to debian stretch. This will remove the need for …

infothrill:
…a custom dumb-init as well as upgrade to the latest transmission.

Should resolve issue #392 for the arm platform.

Also works hand in hand with #395


haugene:
Sweet! I don't run on Raspberry myself and can't verify the changes, but it looks good :)
Thanks for contributing!

PR #408, opened 2018-01-25
Added UFW_EXTRA_PORTS, Updated start.sh script, Added python2.7

alron:
UFW_EXTRA_PORTS Allows for exposing a list of ports when using UFW. Accepts comma separated list. If you want everything behind the transmission-openvpn using container mode networking while keeping UFW in place, you'll need something like this.

Changed from /bin/sh to /bin/bash. If you got bash, why use dash? Plus dash doesn't allow for advanced string manipulations. ${UFW_EXTRA_PORTS//,/ } for example (at least in my tests).

Added python2.7 and pysqlite2. torrentToMedia from the nzbToMedia package needs python2.7. Gives more compatibility.


haugene:
A good extension to the UFW functionality, and I agree that bash is better than dash and should be used. I think it was a bit of overengineering to prepare for running on other distros without bash but not important. The code looks good to me. Merging. Thanks for contributing!

PR #416, opened 2018-02-01
Updated cryptostorm openvpn files, fixes #371

### LINKED ISSUE #371: Please update Cryptostorm openvpn config files
The VPN provider Cryptostorm has updated their CA certificate, which means all clients need to switch to the new one.  The config files, with embedded certificates, used in this docker expired today and consequently openvpn can not connect when using cryptostorm as a VPN provider. 

This should be fixable by simply replacing the files in the /openvpn/cryptostorm/ directory with these new ones (from Cryptostorm's repo): https://github.com/cryptostorm/cryptostorm_client_configuration_files/tree/master/linux

BTW: Thanks for a great container.

flipflopsandrice:


haugene:
Looking good! Thanks for contributing!

I'll merge it now, but wondering if we should cut the prefix of cstorm and linux for the cryptostorm ovpn files. But that's another improvement and now it's more important to make it connect at all :)

Also, the `ca /etc/openvpn/cryptostorm/ca.crt` line in the configs is probably not needed. I see it's there in the old configs as well, but there is no ca.crt file. The ca certificate is inlined in the config files.

Issue #418, opened 2018-02-06
Local network KO after 1.20 update


Mikiya83:
Hi
Thanks for the security update.
Unfortunately, after this update, i cannot access port 9091 in local network.
I use the same conf file as 1.19 (ran without problems), with  "-e LOCAL_NETWORK=192.168.1.0/24" and "-p 9091:9091" but now with browser or RemoteTransmission, i cannot reach  it in 1.20 version.
Do you have any ideas ? 
Thanks !

PS : in logs i can't see any relevant errors, i will post it later if needed.

haugene:
I have an idea: https://github.com/haugene/docker-transmission-openvpn/pull/395
There's a new config variable `TRANSMISSION_RPC_HOST_WHITELIST` that I assume is the culprit. Haven't tried it myself so I don't know the correct value to set it to. You can disable the whole thing with `TRANSMISSION_RPC_HOST_WHITELIST_ENABLED` but I guess that removes the security fix.

Mikiya83:
Ok so i will play with TRANSMISSION_RPC_HOST_WHITELIST variable until it works, thanks ;)

Mikiya83:
I can't succeed to connect even if i set "TRANSMISSION_RPC_HOST_WHITELIST=* " :(
But i found this, maybe it's related : https://github.com/transmission/transmission/issues/476

dcrdev:
I already had TRANSMISSION_RPC_HOST_WHITELIST_ENABLED set to false and the update last night broke things for me also.

haugene:
The issue linked above mentions a bug in the ubuntu package. Not sure what to do until the ppa source is updated with 2.93 version in that case, ref #392. Just downgrade in the meantime? Could do that and tag it 1.21 so that people could run the patched version on 1.20 if they want. Or enable the ppa source on the dev tag and wait to release 1.21 until transmission 2.93 is out?

clowrym:
I've switched back to the previous commit as well.

jfmcbrayer:
When people are saying they've switched back to the previous commit; is it sufficient to add the version tag on the docker pull call in your startup script (e.g., from systemd)? Or is it necessary to clone the repository from the particular tag and rebuild it?

haugene:
Adding the tag :1.19 should be sufficient. But since so many people are having trouble with this, I think I'll revert the change for the latest tag as well. The change will come back with version 2.93 of Transmission, but a setup example should be provided in the README at that time

[...]

PR #420, opened 2018-02-08
set /bin/bash as shell for user when starting transmission

heckface:
Possible fix for #418


haugene:
Sweet, makes sense setting the shell env. Thanks for contributing!

PR #372, opened 2017-12-23
Added pre/post custom scripts

himred:
In order to enhance customization, it would be handy to add custom scripts fired on events
This PR add 4 events:

- /config/custom-pre-start.sh is fired just before transmission is started
- /config/custom-post-start.sh is fired just after transmission is started
- /config/custom-pre-stop.sh is fired just before transmission is stopped
- /config/custom-post-stop.sh is fired just after transmission is stopped

This will enable a lot of possibilities, using custom scripts:
- add custom iptables rules
- change the vpn server dynamicly
- any other idea I don't have yet ;)

Himred

PR #421, opened 2018-02-08
Dev

himred:
In order to enhance customization, it would be handy to add custom scripts fired on events
This PR add 4 events:

/scripts/transmission-pre-start.sh is fired just before transmission is started
/scripts/transmission-post-start.sh is fired just after transmission is started
/scripts/transmission-pre-stop.sh is fired just before transmission is stopped
/scripts/transmission-post-stop.sh is fired just after transmission is stopped
This will enable a lot of possibilities, using custom scripts:

add custom iptables rules
change the vpn server dynamicly
any other idea I don't have yet ;)

Himred

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,21 @@ You may set the following parameters to customize the user id that runs transmis
 |`PUID` | Sets the user id who will run transmission | `PUID=1003`|
 |`PGID` | Sets the group id for the transmission user | `PGID=1003` |
 
+If you ever need to run custom code before or after transmission is executed or stopped, you can use the custom scripts feature.
+Custom scripts are located in the /scripts directory which is empty by default.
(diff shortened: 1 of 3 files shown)


himred:
This is the new PR for issue #372

haugene:
Sorry for the delay. Looks good 👍 Merging now, thanks for contributing!

h45rd:
Thanks for this.

Thought I would ask before diving into it myself:
Do you - by any chance - already have sample scripts for custom iptable rules and dynamic vpn server switching written and handy, @himred :)?

I think adding custom iptable rules e.g. to provide another layer of security against possible ip leakage (c.f. [Deluge VPN docker](https://github.com/binhex/arch-delugevpn/blob/master/run/root/iptable.sh)) will give some peace of mind to the paranoid among us.

PR #429, opened 2018-02-11
Add US OVPN.com servers

avion540:
Automatically chooses between Los Angeles, CA and Miami, FL servers


haugene:
Looks good. Thanks for contributing!

Issue #388, opened 2018-01-12
PIA: CA Vancouver and CA Montreal sites missing


millercentral:
Hello, and thanks for your hard work on this project!

A couple of sites on PIA that support port forwarding appear to be absent from the container configurations, specifically:

CA Montreal  (ca.privateinternetaccess.com) - this may be called "CA North York" currently?
CA Vancouver (ca-vancouver.privateinternetaccess.com)

Would it be possible to add these? Thanks!

haugene:
Closing this as the configs have now been added. Thanks to @eugenioseveri for providing the update.

PR #431, opened 2018-02-16
Add PIA strong profiles

eugenioseveri:
Added Private Internet Access (PIA) "strong" profiles, which use more secure algorithms:

- symmetrical cipher: _AES-256-CBC_ instead of _AES-128-CBC_;
- hashing: _SHA-256_ instead of _SHA-1_;
- CA key: _RSA-4096_ instead of _RSA-2048_.

Also added some missing PIA "regular" profiles and renamed "CA North York" to "CA Montreal". This solves issue #388 .


haugene:
Nice 👍Thanks for contributing

PR #435, opened 2018-02-23
Fix settings.json file attributes on Synology NAS.

### LINKED ISSUE #423: Synology NAS: Default file attributes problem
Hi, what are the default file attributes?

When i run container on Synology Nas with root use i see that:
```
# ls -al /volume1/Transmission-vpn/config/
total 8
d---------+ 1 root         root   106 Feb  8 23:44 .
drwx------+ 1 transmission users   12 Feb  8 23:44 ..
d---------+ 1 root         root     0 Feb  8 23:44 blocklists
d---------+ 1 root         root     0 Feb  8 23:44 resume
----------+ 1 root         root  2299 Feb  8 23:44 settings.json
d---------+ 1 root         root     0 Feb  8 23:44 torrents
----------+ 1 root         root  2192 Feb  8 23:45 transmission.log
```
And all work fine, but when i change user to not root, he can't read file with ----------+ permissions and fails.

```
docker run \
-p 19091:9091 \
--name transmission-openvpn \
--cap-add=NET_ADMIN \
--device=/dev/net/tun \
--restart=always \
--dns 8.8.8.8 \
--dns 8.8.4.4 \
--log-driver json-file \
--log-opt max-size=10m \
-v /volume1/Transmission-vpn:/data \
-v /etc/localtime:/etc/localtime:ro \
-e "OPENVPN_PROVIDER=xxx" \
-e "OPENVPN_CONFIG=xxx" \
-e "OPENVPN_USERNAME=xxx" \
-e "OPENVPN_PASSWORD=xxx" \
-e "OPENVPN_OPTS=--inactive 3600 --ping 10 --ping-exit 60" \
-e "TRANSMISSION_DOWNLOAD_DIR=/data/downloads/complite" \
-e "TRANSMISSION_INCOMPLETE_DIR=/data/downloads/incomplete" \
-e "TRANSMISSION_HOME=/data/config" \
-e "TRANSMISSION_ENCRYPTION=2" \
-e "TRANSMISSION_RATIO_ENABLED=true" \
-e "TRANSMISSION_RATIO=3" \
haugene/transmission-openvpn:latest
```

evsasha:
By default in Synology NAS the user Root creates files with the permissions "---------" (0).
Other users can't access to file after change file owner.
I add `chmod 640` after file creates and before `chown `command if attributes = 0.

```
echo "Check transmission settings.json file attributes"
SETTINGS_PATH=${TRANSMISSION_HOME}/settings.json
SETTINGS_ATTRIBUTES=$(stat -c %a ${SETTINGS_PATH})
if [ ${SETTINGS_ATTRIBUTES} = 0 ]; then
  # Ensure settings.json have non zero attributes (actual for Synology NAS)
  chmod 640 ${SETTINGS_PATH} && echo "INFO: Attributes of setting.json changed to 640"
fi
```

P.S. This is my first pull request.

Fix #423

### DIFF
--- a/transmission/start.sh
+++ b/transmission/start.sh
@@ -39,6 +39,14 @@ echo "Generating transmission settings.json from env variables"
 mkdir -p ${TRANSMISSION_HOME}
 dockerize -template /etc/transmission/settings.tmpl:${TRANSMISSION_HOME}/settings.json
 
+echo "Check transmission settings.json file attributes"
+SETTINGS_PATH=${TRANSMISSION_HOME}/settings.json
+SETTINGS_ATTRIBUTES=$(stat -c %a ${SETTINGS_PATH})
+if [ ${SETTINGS_ATTRIBUTES} = 0 ]; then
+  # Ensure settings.json have non zero attributes (actual for Synology NAS)
+  chmod 640 ${SETTINGS_PATH} && echo "INFO: Attributes of setting.json changed to 640"
+fi
+
 echo "sed'ing True to true"
 sed -i 's/True/true/g' ${TRANSMISSION_HOME}/settings.json
 
--- a/transmission/start.sh
+++ b/transmission/start.sh
@@ -39,14 +39,6 @@ echo "Generating transmission settings.json from env variables"
 mkdir -p ${TRANSMISSION_HOME}
 dockerize -template /etc/transmission/settings.tmpl:${TRANSMISSION_HOME}/settings.json
 
-echo "Check transmission settings.json file attributes"
-SETTINGS_PATH=${TRANSMISSION_HOME}/settings.json
-SETTINGS_ATTRIBUTES=$(stat -c %a ${SETTINGS_PATH})
-if [ ${SETTINGS_ATTRIBUTES} = 0 ]; then
-  # Ensure settings.json have non zero attributes (actual for Synology NAS)
-  chmod 640 ${SETTINGS_PATH} && echo "INFO: Attributes of setting.json changed to 640"
-fi
-
 echo "sed'ing True to true"
 sed -i 's/True/true/g' ${TRANSMISSION_HOME}/settings.json
 
--- a/transmission/userSetup.sh
+++ b/transmission/userSetup.sh
@@ -16,6 +16,14 @@ if [ -n "$PUID" ] && [ ! "$(id -u root)" -eq "$PUID" ]; then
         ${TRANSMISSION_DOWNLOAD_DIR} \
         ${TRANSMISSION_INCOMPLETE_DIR} \
         ${TRANSMISSION_WATCH_DIR}
+        
+    echo "Setting permission for files (644) and directories (755)"
+    chmod -R go=rX,u=rwX \
+        /config \
+        ${TRANSMISSION_HOME} \
+        ${TRANSMISSION_DOWNLOAD_DIR} \
+        ${TRANSMISSION_INCOMPLETE_DIR} \
+        ${TRANSMISSION_WATCH_DIR}        
 fi
 
 echo "


haugene:
Hey! First of all thanks for finding the issue, and providing a fix. It looks good, and would solve this problem. But I have a thought.

You say it works fine as root, because root can do whatever it wants. So this is more related to the PUID/PGID feature. That stuff happens in userSetup.sh, and I think we instead of just changing the one file can do broader chmod/chown there instead.

This is the code for setting permissions for custom a custom user:
```
    echo "Setting owner for transmission paths to ${PUID}:${PGID}"
    chown -R ${RUN_AS}:${RUN_AS} ${TRANSMISSION_HOME}
    chown ${RUN_AS}:${RUN_AS} \
        /config \
        ${TRANSMISSION_DOWNLOAD_DIR} \
        ${TRANSMISSION_INCOMPLETE_DIR} \
        ${TRANSMISSION_WATCH_DIR}
```

What if we also did a recursive chmod for transmission_home, set 644 for everything. And I think we can make the last chown recursive as well. Transmission will run as this user, it should own all config and transmission data including subfolders. This should also make it possible changing user easier as I imagine that could cause ownership issues if you try deleting data downloaded and owned by another UID.

What do you think?

evsasha:
I think that's a good idea.
We need 644 for files and 755 for dirs:

```
echo "Setting permission for files and directories"
chmod -R 644 \
    /config \
    ${TRANSMISSION_DOWNLOAD_DIR} \
    ${TRANSMISSION_INCOMPLETE_DIR} \
    ${TRANSMISSION_WATCH_DIR}
chmod -R a+X \
    /config \
    ${TRANSMISSION_DOWNLOAD_DIR} \
    ${TRANSMISSION_INCOMPLETE_DIR} \
    ${TRANSMISSION_WATCH_DIR}
```

haugene:
Not tested, but I would expect the result there to be that all files and folders have 755? I think we're looking for something like the accepted answer here: https://stackoverflow.com/questions/3740152/how-do-i-set-chmod-for-a-folder-and-all-of-its-subfolders-and-files-in-linux-ubu

Plus the extra chmod for transmission_home

evsasha:
```
    echo "Setting permission for files (644) and directories (755)"
    chmod -R go=rX,u=rwX \
        /config \
        ${TRANSMISSION_HOME} \
        ${TRANSMISSION_DOWNLOAD_DIR} \
        ${TRANSMISSION_INCOMPLETE_DIR} \
        ${TRANSMISSION_WATCH_DIR}
```
755 for directories and 644 for file is a default permissions in Linux.
    644 (-rw-r-r--) -  All users have the right to read files; owner can edit.
    755 (drwxr-xr-x) - All users have the right to browsing and read directories content; owner can edit.
I think we dont need grant execut permmission on files.

Command in this variant `chmod -R go=rX,u=rwX` is safe, no overrite, for example if file before chmod have 0 permission, after its 644, But if file already have 755 permission, chmod doing nothing.

haugene:
Looking good! Had to read up on chmod though. I saw it worked with 644 and 755 but couldn't understand why everything wasn't 755. But there's a difference between x and X it seems ;) Merging now. But... For the section above:

```
    echo "Setting owner for transmission paths to ${PUID}:${PGID}"
    chown -R ${RUN_AS}:${RUN_AS} ${TRANSMISSION_HOME}
    chown ${RUN_AS}:${RUN_AS} \
        /config \
        ${TRANSMISSION_DOWNLOAD_DIR} \
        ${TRANSMISSION_INCOMPLETE_DIR} \
        ${TRANSMISSION_WATCH_DIR}
```

I think we can safely and also more correctly set:
```
    echo "Setting owner for transmission paths to ${PUID}:${PGID}"
    chown -R ${RUN_AS}:${RUN_AS} \
        /config \
        ${TRANSMISSION_HOME} \
        ${TRANSMISSION_DOWNLOAD_DIR} \
        ${TRANSMISSION_INCOMPLETE_DIR} \
        ${TRANSMISSION_WATCH_DIR}
```

What do you think? All files under these folders should be owned by the transmission user?

evsasha:
I agree with all your changes.
Looks simply and clearly.

PR #436, opened 2018-02-23
Add missing bash dependency in Alpine image

e7d:
The "bash" environment was absent from the Alpine image, making "dumb-init" unable to start the main script "start.sh"

Also, kettu was not installed in the proper directory


haugene:
Looks good. Thanks for fixing this.

Issue #366, opened 2017-12-17
Feature request: Backup server


himred:
Hello,

First of all thanks for this amazing container, I use it everyday.
Yesterday I got an issue, the VPN server of my provider went down.
Of course, the container did its work and lost connectivity.
To fix it, I had to manually change the settings to use another server from my VPN provider.

I would be very handy to have the ability to define a backup server.
So when the main server goes down, the backup server would be used.

Something like:

OPENVPN_CONFIG=mymainserver
OPENVPN_BACKUPCONFIG=server2

Regards,

Salim

haugene:
Good thought, but... It will require some code and complexity and KISS is a good principle. Then you have the "fail fast" principle and the manual change isn't the end of the world.

The complexity would be that we then probably need to persist to file that the connection failed, so that it chooses a new connection on restart (given that you run with a restart flag). It's either that or adding more scripts to handle exits signals from the VPN and this complexity would increase the risk of errors that could affect the killswitch etc. I'm just not sure it's worth it. But if someone wants to give it a go and submit a PR I'd be interested if the solution is a clean one.

Any suggestions?

pjvds:
We could accept multiple OPENVPN_CONFIG_* env vars being set in the container and pick just a random one on startup. This way, when one configuration fails, the container gets restarted till a random config that works is picked.

The configs could just be OPENVPN_CONFIG_1, OPENVPN_CONFIG_2, OPENVPN_CONFIG_3,... and so forth. This would keep the solution KISS and there will be no smartness other than random config picking.

Curious about your thoughts.

himred:
I have an idea to implement the backup server feature, but before coding it and submitting a PR, I wanted to make sure it will be accepted.

I propose the following logic:

1. We add a new env variable OPENVPN_BACKUP_CONFIG which point to the backup server config
2. User will need to add a health check using the docker healtchcheck feature pointing to a custom script located in /scripts
3. The script will do the connectivity checks (ping something like 8.8.8.8 or anything the user wants)
4. If a failure is detected (maybe more than once) the script will touch a file in /scripts named .use_backup_config and kill openvpn
5. This will trigger via dumb-init a restart of the container
6. When the container will restart, openvpn/start.sh will look if .use_backup_config is present, if yes it will use the variable OPENVPN_BACKUP_CONFIG in place of OPENVPN_CONFIG

This will require:

1. To write the healtcheck script
[...]

PR #438, opened 2018-02-23
Add support for multiple servers in OPENVPN_CONFIG

himred:
This PR is for issue #366 

it adds support for multiple servers in OPENVPN_CONFIG
If a list of files is provided in OPENVPN_CONFIG, a file in the list is chosen randomly

### DIFF
--- a/openvpn/start.sh
+++ b/openvpn/start.sh
@@ -11,6 +11,15 @@ echo "Using OpenVPN provider: $OPENVPN_PROVIDER"
 
 if [ ! -z "$OPENVPN_CONFIG" ]
 then
+	n=$(echo "$OPENVPN_CONFIG" | wc -w)
+	if [ $n -gt 1 ]
+	then
+        	rnd=$((RANDOM%n+1))
+        	srv=$(echo "$OPENVPN_CONFIG" | awk -vrnd=$rnd '{print $rnd}')
+        	echo "$n servers found in OPENVPN_CONFIG, $srv chosen randomly"
+        	OPENVPN_CONFIG=$srv
+	fi
+
 	if [ -f $vpn_provider_configs/"${OPENVPN_CONFIG}".ovpn ]
   	then
 		echo "Starting OpenVPN using config ${OPENVPN_CONFIG}.ovpn"
--- a/README.md
+++ b/README.md
@@ -70,6 +70,9 @@ If you provide a list, a file will be randomly chosen in the list, this is usefu
 ```
 -e "OPENVPN_CONFIG=ipvanish-AT-Vienna-vie-c02 ipvanish-FR-Paris-par-a01 ipvanish-DE-Frankfurt-fra-a01"
 ```
+If you provide a list and the selected server goes down, after the value of ping-timeout the container will be restarted and a server will be randomly chosen, note that the faulty server can be chosen again, if this should occur, the container will be restarted again until a working server is selected.  
+
+To make sure this work in all cases, you should add ```--pull-filter ignore ping``` to your OPENVPN_OPTS variable.  
 
 As you can see, the container also expects a data volume to be mounted.
 This is where Transmission will store your downloads, incomplete downloads and look for a watch directory for new .torrent files.
--- a/README.md
+++ b/README.md
@@ -65,6 +65,12 @@ Find available OpenVPN configurations by looking in the openvpn folder of the Gi
 -e "OPENVPN_CONFIG=ipvanish-AT-Vienna-vie-c02"
 ```
 
+You can also provide a list of openvpn configuration filenames separated by a space.  
+If you provide a list, a file will be randomly chosen in the list, this is useful for redundancy setups. For example:
+```
+-e "OPENVPN_CONFIG=ipvanish-AT-Vienna-vie-c02 ipvanish-FR-Paris-par-a01 ipvanish-DE-Frankfurt-fra-a01"
+```
+
 As you can see, the container also expects a data volume to be mounted.
 This is where Transmission will store your downloads, incomplete downloads and look for a watch directory for new .torrent files.
 By default a folder named transmission-home will also be created under /data, this is where Transmission stores its state.


pjvds:
Awesome change. Simple and effective. But can we do comma instead of a space as separation character?

haugene:
Elegant solution, I like it. Any particular reason you want it to be comma separated @pjvds? It's no big change of course, but with space separation we follow a [default way to create arrays in bash](http://tldp.org/LDP/Bash-Beginners-Guide/html/sect_10_02.html). Not sure, but I think we have to split by comma, and then remove the commas in the other approach. 

I wonder if we should add a half sentence in the readme describing how this works for redundancy in the setup. Ie that if a server is down, connection will fail and the container gets restarted and picks a new server at ransom. Or is that obvious?

himred:
Thanks.

About the readme, are you sure  the container will be restarted if the connection fails ?
For me, the container is restarted if openvpn exit and not if the connection is stall.
Let me check this first.
And yes i think a half sentence is welcome.

And about the space/comma choice, it's not a Big deal

himred:
As I have suspected, if the connexion is stall, openvpn does not exit even with `OPENVPN_OPTS=--inactive 3600 --ping 10 --ping-exit 60`

Here the log (I blocked packets at 17:57:00)

```
STARTING TRANSMISSION
NO PORT UPDATER FOR THIS PROVIDER
Executing /scripts/transmission-post-start.sh
/scripts/transmission-post-start.sh returned 0
Sun Feb 25 17:56:24 2018 /sbin/ip route add xx.xx.xx.xx/32 via 172.30.3.1
Sun Feb 25 17:56:24 2018 /sbin/ip route add 0.0.0.0/1 via 10.8.8.1
Sun Feb 25 17:56:24 2018 /sbin/ip route add 128.0.0.0/1 via 10.8.8.1
Sun Feb 25 17:56:24 2018 Initialization Sequence Completed
Sun Feb 25 18:04:41 2018 [xx.xx.xx.xx] Inactivity timeout (--ping-restart), restarting
Sun Feb 25 18:04:41 2018 SIGUSR1[soft,ping-restart] received, process restarting
Sun Feb 25 18:04:41 2018 Restart pause, 5 second(s)
Sun Feb 25 18:04:46 2018 NOTE: the current --script-security setting may allow this configuration to call user-defined scripts
Sun Feb 25 18:04:46 2018 TCP/UDP: Preserving recently used remote address: [AF_INET]185.145.66.247:1194
Sun Feb 25 18:04:46 2018 Socket Buffers: R=[212992->425984] S=[212992->425984]
Sun Feb 25 18:04:46 2018 UDP link local: (not bound)
Sun Feb 25 18:04:46 2018 UDP link remote: [AF_INET]xx.xx.xx.xx:1194
```

Several points to be noted:
- it took 8 minutes and not 60 secs to openvpn to detect the link stall
- openvpn does not exited, it just received SIGUSR1 which I guess is handled by signal()

In conclusion, in the default configuration, when a link is stall, the container will not be restarted, openvpn does not exit, and thus openvpn/start.sh is not started and thus a new server is not elected.

This is strange as the official documentation state that --ping-exist exit openvpn and --ping-restart send SIGUSR1, I use --ping-exit but in the log we see `SIGUSR1[soft,ping-restart] received, process restarting`

So far, adding this feature will not help a lot.

I have a workaround but it may be considered complex by users.
Unless we find why openvpn does not exit even with --ping-exit option enabled, adding this feature is useless.

My workaround would be the following:
- Add a healthcheck in my docker-compose.yml file, ie:
```
 healthcheck:
    test: ping -w 8 -n -c 1 8.8.8.8 || exit 1
    interval: 10s
    timeout: 10s
    retries: 3

```

This will tag the container as unhealthy if it loose connectivity within 30s.
Currently docker does not automaticly restart unhealthy containers (weird)
I have coded a little container that restart unhealthy containers, I use it a lot:
https://github.com/himred/hcd

I'm still investigating why openvpn does not exit, I let you know if I find something.

haugene:
First of all, thanks for testing it thoroughly. This is important behavior for the container as we're depending on the container exiting so that docker can restart it.

Which provider are you using? Could the ping-exit option option be in conflict with other options set in the ovpn file?

himred:
You are welcome,

Here is my docker-compose.yml file:

```
torrent:
  image: haugene/transmission-openvpn
  container_name: torrent
  hostname: torrent
  networks:
    - net
  cap_add:
    - NET_ADMIN
  devices:
    - /dev/net/tun
  restart: always
  ports:
    - "9091:9091"
  dns:
    - 8.8.8.8
    - 8.8.4.4
  volumes:
    - /etc/localtime:/etc/localtime:ro
    - torrent:/data
    - scripts:/scripts

  logging:
    driver: "json-file"
    options:
      max-size: "50m"
      max-file: "2"

  healthcheck:
    test: ping -w 8 -n -c 1 8.8.8.8 || exit 1
    interval: 10s
    timeout: 10s
    retries: 3

  environment:
    - OPENVPN_PROVIDER=NORDVPN
    - OPENVPN_USERNAME=hidden
    - OPENVPN_PASSWORD=hidden
    - OPENVPN_OPTS=--inactive 3600 --ping 10 --ping-exit 60
    - OPENVPN_CONFIG=de110.nordvpn.com.udp
    - "TRANSMISSION_BIND_ADDRESS_IPV6=fe80::"
    - TRANSMISSION_WATCH_DIR_ENABLED=false
    - TRANSMISSION_INCOMPLETE_DIR_ENABLED=false
    - LOCAL_NETWORK=192.168.56.0/24

```

and here is the ps fax on the container:

```
1997 pts/0    Ss     0:00 /bin/bash
 2035 pts/0    R+     0:00  \_ ps fax
    1 ?        Ss     0:00 dumb-init /etc/openvpn/start.sh
    8 ?        Ss     0:00 openvpn --script-security 2 --up-delay --up /etc/transmission/start.sh --down /etc/transmission/stop.sh --inactive 3600 --ping 10 --ping-exit 60 --config /etc/openvpn/nordvpn/de110.nordvpn.com.udp.ovpn
   43 ?        Ssl    0:01 /usr/bin/transmission-daemon -g /data/transmission-home --logfile /data/transmission-home/transmission.log

```

As you can see there is no specific directives, pretty straightforward setup.
My provider is NORDVPN.

himred:
I finally found why openvpn was ignoring the --ping-exit and behaving like I specified --ping-restart
ie sending SIGUSR1 in place of SIGTERM

This is because the openvpn client can receive extra options from the openvpn server.
This is allowed by the option pull, which is present in NORDVPN configuration files.
NORDVPN is sending ping-restart option which override the ping-exit we provide on command line.

I just added the option --pull-filter ignore ping and now it works as expected.

Now my OPENVPN_OPTS variable is : `OPENVPN_OPTS=--inactive 3600 --ping 10 --ping-exit 60 --pull-filter ignore ping`

haugene:
Awesome. Good catch 👍. A bit sneaky of nordvpn to set ping-exit in it's default config and push ping-restart after connection. The ping-exit with restart strategy is a critical feature that hasn't then been working for everyone using nordvpn as far as I can understand. 

This is stuff that's important to the base operation and stability of the image. I have had half a mind to do some cleanup on this for a while. And I think with this as well, it's time.

So what I propose is to use sed and change all existing ovpn files like this:
- Remove all ping-restart and keepalive options (mutually exclusive with ping-exit)
- Add ping, ping-exit and inactive options so these are no longer necessary as environment variables
- All files with pull should set pull-filter

Then we should also add a note in the readme that these are the things we need to change in our ovpn files before adding them to the image. It's these options and it's the auth user-pass file and certificate paths.

What do you think? And am I forgetting something?

himred:
I agree 100% with you, and I don't see anything you forgotten.

The providers that use the pull option are: blackvpn, nordvpn, ovpn, tiger, vpnac, vpnbook, vpnht

> Add ping, ping-exit and inactive options so these are no longer necessary as environment variables

Yes, but if we do this, we should note somewhere in the README how to change the timeouts you will hardcode in the config files. using OPENVPN_OPTS, what do you think ?

Keep in mind a lot of users won't change their settings and thus will keep their "old" OPENVPN_OPTS variable as is. thus the most important is to update the providers files that have pull enabled to add pull-filter.

openvpn/start.sh
```diff
@@ -11,6 +11,15 @@ echo "Using OpenVPN provider: $OPENVPN_PROVIDER"
 
 if [ ! -z "$OPENVPN_CONFIG" ]
 then
+	n=$(echo "$OPENVPN_CONFIG" | wc -w)
```
alron: This whole block could be something like the following:
```
OPENVPN_CONFIG_ARRAY=( ${OPENVPN_CONFIG//,/ } )
if (( ${#OPENVPN_CONFIG_ARRAY[@]} > 1 )); then
     OPENVPN_CONFIG_RANDOM=$((RANDOM%${#OPENVPN_CONFIG_ARRAY[@]}))
     echo "${#OPENVPN_CONFIG_ARRAY[@]} servers found in OPENVPN_CONFIG, ${OPENVPN_CONFIG_ARRAY[${OPENVPN_CONFIG_RANDOM}]} chosen randomly"
     OPENVPN_CONFIG="${OPENVPN_CONFIG_ARRAY[${OPENVPN_CONFIG_RANDOM}]}"
 fi
```

Which removes all the calls to external applications (wc, awk, potentially echo, and probably [ and ]).

Instead of space separation, this block uses commas.

himred: Thanks alron,

This is all about coding style and code readibility.

When I read haugene code, he use tr in place of $(//), [ in place  of [[, ...
I consider my duty to add code in the author coding style, so he can maintain it easily.

IMHO, There is no reason to optimize 2 commands when they are not executed a zillion times in a loop or where performance is not an issue.
Using sed or ${//}, cut or ${#%%} is useful when performance is an issue, this is not the case here, and this is done at the expense of code readibility and portability.

I can code both styles, it's finnaly a matter of taste.

haugene:
Hi again. Sorry about the delay.

I think this looks good and will merge now. I agree that we'll just stick to the current solution because of code style and readability, but it was a clean solution from @alron without invoking other tools like awk etc.

Anyways. Seems like we have a plan for cleaning the ovpn files. I will create a new issue. At that point we can also remove the line in the readme detailing the pull-filter. Not sure it's clear for everyone why that is needed.

openvpn/start.sh
vSh1ny: Unfortunately, as some of .ovpn files for PIA use spaces in their names, it's not about coding style. IMHO, the delimiter shouldn't be a character eligible for a filename.
See #445.

alron: @vSh1ny #440 includes the comma delimiter block I suggested above, along with other firewall updates.

vSh1ny: @alron Whoops, didn't see that. Great PR, waiting for the merge. Thanks.

PR #440, opened 2018-02-25
Firewall feature changes & OPENVPN_CONFIG fix.

alron:
Added UFW_ALLOW_GW_NET. Changes firewall rules from using GW to the GW network when set to true.

Extended LOCAL_NETWORK to support comma seperated list of /CIDR.

Minor firewall cleanups. Removed specified TCP in range allow. No point, we don't specify tcp/udp anywhere else.

Formatting changes. 

Cleaned removed external [ ] calls. Use builtin [[ ]]. 

Use ${VAR,,} to lowercase isntead of TR.


haugene:
Looks good 👍 Thanks!
You're missing the UFW_ALLOW_GW_NET variable in Dockerfile.alpine, and there's now a merge conflict. Will merge once updated.

alron:
I totally forgot about the Dockerfile.alpine (or more so, I didn't see. whoops). I'll get that added.

If you want, I can also drop most of the BASH 3+ style/function changes I made. They're habit for me as I write a lot of KSH93 for work.  

Some are style like the OCD changing of $VARIABLE to ${VARIABLE} and the use of ${VAR_NAME} upper case for global vars, and ${camelCaseVars} for local scope. Others are functional and directly related to the way that BASH (and KSH) handle variables and expansion inside the if statements... and probably makes no difference here other than being one of my stupid habits. 

I'm used to working in mixed platform environments (Solaris/Linux/BSD) where using external application calls to things like sed/awk/grep/tr might not work as intended if you're expecting GNU style, or where you have to do platform verification and call applications differently. Most all of the time it is just easier to use the builtin manipulations and save a lot of trouble.

I've also checked all of the other shell scripts and didn't find anywhere that'd actually benefit from changes, otherwise I'd probably done something stupid like including them all in my pull. ^^;

Just let me know your preference. I totally don't want to step on your toes here. :)

himred:
@alron, I code shell scripts since 25+ years now and also teach it at university and I like your coding style.
It really use the shell features to avoid forks and commands compatibility issues, I wish my students code like you ;) 
This beeing said, for me, `[` does not use any external command, the main difference between the standard bash `[` and it's extended version `[[` is mainly regex support `[[ $var =~ ]]`
Even if we use `if test` no fork will occur as test is a shell builtin since years now.

```
# type [
[ is a shell builtin
type [[
[[ is a shell keyword
type test
test is a shell builtin
```

Regards from France.

alron:
@himred You are absolutely correct about the bult-in override. But I still try and stay away from the use of [ as it is still an application at the shell level. Some environments I've worked in ... didn't have versions new enough to override, and under heavy long term loads could give some extremely special out of order results. Totally not related to my use of them here, just history as to why I'm OCD about them. :)


From my local Ubuntu box. Similar results can be found on a Mac with /bin/[
```
$ /usr/bin/\[ --help | head -10
Usage: test EXPRESSION
  or:  test
  or:  [ EXPRESSION ]
  or:  [ ]
  or:  [ OPTION
Exit with the status determined by EXPRESSION.

      --help     display this help and exit
      --version  output version information and exit

```

```
$ ls -lh /usr/bin/\[
-rwxr-xr-x 1 root root 51K Mar  2  2017 /usr/bin/[
```

haugene:
Nah, I like the BASH 3+ style stuff you've added. In general I'm all for the builtin functions instead of relying on other tools that might behave differently depending on platform.

For the project as a whole I'm code style aware to the point where every .sh file shouldn't be in completely different styles so it's easy to reuse and understand across the project. But apart from that I like seeing more effective and clean ways to do things. I'm using bash scripts for work and have gotten decent at it, but it's just to script tools for convenience while developing Java-based systems. So I'm still very google-dependant while scripting.

So I say let's keep the style you've added and try to gradually make the rest of the project in the same style :)

alron:
@haugene I've updated the Dockerfile.alpine that was missed previously and resolved all of the conflicts.

Additionally, I've included a patch to fix the breakage caused by #438 splitting on space in the OPENVPN_CONFIG variable. As there are many providers configured with spaces, this is the least impacting way of providing multiple servers.
 
Example of existing PIA configs with spaces (using the new random):
```
- "OPENVPN_CONFIG=CA Vancouver,CA Toronto,CA Montreal"
```

haugene:
Awesome! Merging.

PR #442, opened 2018-02-28
Use same rpi-raspbian:stretch image for proxy

bharat555:
Use the same resin/rpi-raspbian:stretch image for proxy to not have to download another image and save space.


haugene:
Good, thanks! 👍

PR #448, opened 2018-03-11
Update README.md

jdk:
Fixes RSS container documentation


haugene:
Look at that. Yup, seems correct to me. Thanks for updating it!

PR #451, opened 2018-03-13
Add provider TunnelBear

wmoulton:
Adding support for TunnelBear to the Dev project, for review and approval. I'm not experienced with creating dockers, so if this could be merged, I would be thankful!


haugene:
Looks good! Thanks for contributing.

A couple of points to update:
- [ ] ovpn files names should be changed from "TunnelBear Germany" to "Germany" and so on
- [ ] Add the new provider to the list in README
- [ ] Squash your commits

After that I'll happily merge a new provider :)

haugene:
Changes look good. But there's 30 commits, and it would be nice if you could squash them first. Basically do a git reset to the HEAD of my dev branch and then commit all changes in one go.

wmoulton:
I tried to do it from the web interface, but was unable ;) I will do it as soon as I get home :)

I’ve looked up the method of how to do it - just have to do it now!

Can you tell I’m a newb? 

Thanks for your instruction- been very helpful!

William

PR #461, opened 2018-03-29
Added a list of SlickVPN configs

qoobaa:


h45rd:
I applaud your efforts.
But I would not split the locations by single gateways.
Some of them change or are unavailable every once in a while, but having all possible gateways for a location - even those not currently listed on the official website - in a single _$locations_.ovpn ensures a successful connection as long as any gateway is available for that location at all.

Locally I have been using a script to generate a config with every plausible gateway address and port for a given location and have openvpn iterate through until a successful connection is established.
Not clean enough for a public commit, but working nicely so far.

PR #477, opened 2018-04-13
Fix transmission-rss dockerfile for arm

lyz-code:
Currently the dockerfile of arm didn't work because:

* Ruby max version for jessie is 2.1
* This image needed to install the libc6-dev package
* There was a typo

All of them are fixed now


haugene:
Looks correct to me. Thanks for fixing!

PR #487, opened 2018-04-21
Proxy.sh config files

jobwat:
Here are the proxy.sh config files

`updateConfigs.sh` is just a syntax updater
In order to download the config files zip archive proxy.sh require to be logged in -- https://proxy.sh/configs

Thank you for sharing your project :)


haugene:
Looking good, thanks for contributing! 😃 

Changed base branch to dev and merging there first.

PR #489, opened 2018-04-22
Update config for NordVPN

throwawayyyyyyy:
1. downloaded https://downloads.nordcdn.com/configs/archives/servers/ovpn.zip
2. copied udp and tcp folders to nordvpn
3. ran adjustConfigs.sh.


skogsmaskin:
:+1:

haugene:
Hi, thanks for providing the update and sorry I've been slow.
Changes look good. I see your changing some ping options, I guess this is NordVPN changes.

There's an ongoing thread on issues with the container not restarting on disconnects, and the --ping, --ping-exit option combo has been the preferred way of making it exit and than let Docker restart the container with --restart=always.

But... It seems that doesn't always work, and lots of configs (including these) have the --pull option set as well, so the server can update the client with new configs. So I'll merge it, and we'll see if there comes and update when a strategy for disconnects is in place :)

PR #490, opened 2018-04-23
Updated PureVPN confifguration files

ChrisPetit:
Updated the PureVPN configuration files, since some servers and DNS records changed.


ChrisPetit:
I made an error in the config files...

haugene:
Hi, thanks for contributing and helping keeping the configs up to date 👍 

I have a couple of comments.
I see there's a few configs that have been removed, and a couple that has been updated. But most of the changes are adding a whitespace char to the end of the configs.

Could you remove that, so we see the diff more clearly and not all these files that hasn't really been changed? Also, you've removed the default.ovpn symlink. Please add a symlink called `default.ovpn` to one of the configs that can serve as default.

Thanks again!

haugene:
Still the same issue. Which editor are you using? It's removing newline at the end of the file. Not that it's a bad thing, it's just concealing the real change of configs with a bunch of "non-changes" that are just formatting.

ChrisPetit:
I'm using Visual Studio Code...I was looking into a solution for this. Does git diff --ignore-all-space | grep "+++" work for you?

haugene:
Ok. The formatting isn't that important, I'll merge it with the newline changes.
But your symlink isn't really a symlink. It's another ovpn file called default? Again, not a big deal but other providers just link to a config to use as default.

On a unix based system you create symlinks with `ln -s <actual file> default.ovpn`
And if you could squash the commits that would be great.

ChrisPetit:
This is the first time I'm trying to contribute and learning a lot...thanks! I found out that when comparing files adding ?w=1 to the end of the URL whitespace will be ignored: https://github.com/haugene/docker-transmission-openvpn/pull/490/files?w=1

I've tried to create the symlink, but gitkraken, says it cannot access the file then and refuses to upload...

haugene:
Yeah, it's possible to ignore it. But the change is still there.
I'll merge it and create the symlink, no worries!

haugene:
The new configs should be included on the dev tag once the builds are done: https://hub.docker.com/r/haugene/transmission-openvpn/builds/

If you can verify that everything is working as intended that would be great 😄

ChrisPetit:
Tried a couple of connections and they seem to all work...Again thanks for the help 👍  will check for any errors and let you know this weekend...

PR #503, opened 2018-05-08
Port Fowarding

jeffrangel1:
Just curious, is there a way to port forward from the container while in VPN? Using PIA exactly the same way the example is displayed however the "Peer Listening Port is closed". Any thoughts on how to get around this?


haugene:
Is this supposed to be a PR? As long as you're using PIA the port forwarding should be set up automatically. We've had some problems with this, but it's mostly related to the port forward API of PIA and which servers that support port forwarding. Recreating the container a couple of times should hopefully give an open port. Also seems that the API doesn't like to get spammed, so if it doesn't work after a few times then wait a bit before retrying.

These are the servers that support port forwarding:
https://www.privateinternetaccess.com/helpdesk/kb/articles/how-do-i-enable-port-forwarding-on-my-vpn

PR #541, opened 2018-06-21
Fix missing tinyproxy, optimize build

Noodle05:


haugene:
Looks good to me :) No automated builds here yet and I don't currently run an armhf device so can't test, but looks like an improvement so I'll merge.

Thanks for contributing!

PR #544, opened 2018-06-25
Updated newshosting configs

stuartst:
Updated newshosting configs as current versions are around 2 years old.

Updated using provided updateConfigs.sh script.


haugene:
Hey! Thanks for contributing and sorry for the delay.
For future PRs I would prefer the default.ovpn as a symlink to one of the others, but it's no difference in use so we do that in a future update.

I can't verify that it's working, but the changes looks good so I'll merge it to dev now :)
There's been a few PRs lately and I'll try to merge them before creating a new release and merging to master.

PR #547, opened 2018-06-26
Pass on arguments to pre and post start/stop hooks

haplo:
The *start.sh* and *stop.sh* scripts receive a set of arguments from *openvpn* when invoked from *tunnelUp.sh* and *tunnelDown.sh*. They are:

 * `$1`: network interface for the VPN (e.g. `tun0`)
 * `$2`: interface MTU (e.g. `1500`)
 * `$3`: link MTU (e.g. `1570`)
 * `$4`: local IP (e.g. `10.47.10.6`)
 * `$5`: remote IP if TUN (e.g. `10.47.10.5`), network mask if TAP
 * `$6`: "init" or "restart"

This commit makes these arguments available in *transmission-pre-start.sh*, *transmission-post-start.sh*, *transmission-pre-stop.sh* and *transmission-post-stop.sh* for those users who might need them.

### DIFF
--- a/transmission/start.sh
+++ b/transmission/start.sh
@@ -17,7 +17,7 @@ fi
 if [ -x /scripts/transmission-pre-start.sh ]
 then
    echo "Executing /scripts/transmission-pre-start.sh"
-   /scripts/transmission-pre-start.sh
+   /scripts/transmission-pre-start.sh "$*"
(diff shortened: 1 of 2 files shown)


haugene:
Thanks for contributing and sorry for the delay :)

PR #549, opened 2018-06-26
update readme (PIA port forwarding)

BorysekOndrej:
Hi,

the _PIA_ region _Netherlands_, which is mentioned several times in the readme, currently doesn't support port forwarding. This is in no way fault of this project, however, this project uses it in examples and says:

> And if you're using PIA as provider it will update Transmission hourly with assigned open port.

PR changes the quote above to

> And if you're using PIA as provider it will update Transmission hourly with assigned open port (if the port forwarding feature is available in the selected region).

and also does some small fixes of typos, which I noticed while reading through readme.

Thank you guys for all the work you've been doing on this project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ An example run command to get you going is provided below.
 Also worth mentioning.
 If you want to route web traffic through the same tunnel that Transmission is using there
 is a pre-installed Tinyproxy which will expose a proxy on port 8888 when enabled.
-And if you're using PIA as provider it will update Transmission hourly with assigned open port.
+And if you're using PIA as provider it will update Transmission hourly with assigned open port (if the port forwarding feature is available in the selected region).
 
 GL HF! And if you run into problems, please check the README twice and maybe try the gitter chat before opening an issue :)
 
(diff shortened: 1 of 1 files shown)


haugene:
Great! :)
Should we also change the default to a server that supports port forwarding? Suggestion to which one?

mirirald:
The list of servers with port forwarding is enable for PIA :

- CA Toronto
- CA Montreal
- CA Vancouver
- Czech Republic
- Spain
- Switzerland
- Sweden
- France
- Germany
- Romania
- Israel
- DE Berlin (new server)
- DE Frankfurt (new server)

Also I created an issue to add the new servers 

Good day

Miri

BorysekOndrej:
Hi,
thank you @mirirald for making the list of port forwarding enabled locations. Link to issue @mirirald created #553 about adding new PIA servers.

I'm not sure as for changing the default to location supporting port forwarding. In the readme example for port forwarding, yes, but in the rest, I think the Netherlands should stay.
The Netherlands is ([supposedly](https://www.privateinternetaccess.com/pages/network/)) capable of 548 Gbps bandwidth, while the biggest port forwarding enabled exit node (CA Toronto) of _only_ 125Gbps.  I suppose setting a default on such a big project like this could have a noticeable impact on the performance.

BorysekOndrej:
Config _Germany_ no longer exists and was split to _DE Berlin_ a _DE Frankfurt_. Should there be symbolic link for the removed germany.ovpn to one of the successors?

haugene:
Hey. Great stuff :)

I don't think we need to keep backward compatability with the removed germany.ovpn. It will fail fast when people update their image, and then they can change.

But we should probably change the default.ovpn symlink so it points to one of the port forwarding enabled servers?

PR #555, opened 2018-07-11
Updated all NordVPN config files from https://downloads.nordcdn.com/c…

dragfly:
Updated NordVPN config files and added a simple utility to automatically choose the best server for the country rather than sticking to a fixed one

PR #557, opened 2018-07-15
Annoying problem with NordVPN changing certificates

echaupynx:
To address the problem of changing NordVPN certificates I took the updateConfigs.sh file from PIA, copied and edited it to suit NordVPN. Also updated all the ovpn files using the same script and changed the passed settings to ping 10, ping-exit 60 instead of ping-restart, added inactive 3600. This was taken from the current .ovpn files.
Best regards
Echaup


haugene:
Hey, and thanks for contributing! I see that both this PR and #555 are against the master branch, while the dev branch have updates as well. You have both updated the configs, but also added some utilities for updating configs and automatically selecting server.

Could the two of you agree on the best solution, and one of you rebase against dev and update your PR? Then I'll happily merge some more updates/fixes, and do a release to the master branch.

Thanks for contributing!

PR #566, opened 2018-07-23
Update Dockerfile.alpine

nexusmaniac:
Could bump the Alpine version to 3.8 now that it's stable? :) In theory it'll pull the latest Transmission-daemon 2.94 too :D

Also, `curl` attempts to run as part of the port forward script for a PIA VPN connection... But we don't install curl :P

### DIFF
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -1,4 +1,4 @@
-FROM alpine:3.7
+FROM alpine:3.8
 # MAINTAINER Kristian Haugene
 
 VOLUME /data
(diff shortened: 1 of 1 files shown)


haugene:
Merged 👍 Thanks for contributing!
Let me know how the alpine/transmission update goes ;)

nexusmaniac:
Been running fine since you pushed the change to the Dev channel 😄 

Everything fully up to date, latest OpenVPN, Transmission, etc. Cheers dude! 👍

PR #589, opened 2018-08-27
Updated PureVPN configuration files

Othou:
Old configuration files has been moved to "old configs" folder


haugene:
Great! But do we want to keep the old configs? Are they still working, and will they be obsolete soon?
If we are to keep the old ones, please rename the folder to "old" as spaces in the path requires escaping the whitespace char etc.

Also. If we're keeping the old configs, they need to be changed so that they point to the old ca.crt as well. As far as I can see they're pointing to the main ca now.

merlinfrombelgium:
Great stuff @othou. I did find an error in the config file for Netherlands-udp: it has the FQDN for Norway. Please correct it to the correct name before committing. Looking forward to having a stable config thanks to your work!

haugene:
Thoughs on the old configs @merlinfrombelgium? Should we keep them in /old/ or just clean it out? Seems to me as there are enough options among the new servers IMHO. But I don't use this provider, so I don't have any strong feeling either way. Just that it will be cleaner ;)

Othou:
Don't mention it, I'm happy that I can contribute somehow for this awesome container :) I will get into this later today after work, maybe add some more servers I have

haugene:
Oh, and btw. See #590. We should rename configs with comma in the name, to avoid it being treated as a list of configs. If you those with comma to use - instead, then that bug will be corrected as well :)

Othou:
Alright. I removed old configs, fixed Netherlands-udp and added all other servers i got from PureVPN support. All done for now :)

haugene:
Cool! Looks good to me. Merging to dev so that people can test it.

rrd911:
I noticed in dev there are fewer gateways. are more being added?

Othou:
@rrd911 I added all gateways i got from PureVPN Support. If i understood correctly, they are in progress of making new gateways, so there might be more gateways later for this container.

rrd911:
There are more configurations for PureVPN based on the server page.

https://support.purevpn.com/vpn-servers

PR #599, opened 2018-09-15
Fix arg passing to start/stop script hooks

haplo:
When using `"$*"` all arguments ended up as `$1` in the hook scripts, i.e. `$2` and forward they were all blank. This fixes the code so it works as originally intended (see #547).


haugene:
Yeah, that's the syntax we want :) Thanks!

PR #605, opened 2018-09-19
Dev

Bschalck:
New config for for VPNBOOK with p2p allowed : 
 - DE4 
 - PL226


haugene:
Will do a merge from dev to master later, but for now the new vpnbook configs is available using the dev tag

PR #620, opened 2018-10-12
Create default.ovpn

cyberic79:
Add the default.ovpn for resolve the error : 
Options error: In [CMD-LINE]:1: Error opening configuration file: /etc/openvpn/tunnelbear/default.ovpn

### DIFF
--- a/openvpn/tunnelbear/default.ovpn
+++ b/openvpn/tunnelbear/default.ovpn
@@ -1,0 +1,1 @@
+France.ovpn


haugene:
Right, seems this provider is missing the symlink. Thanks for fixing, merging to dev now and will go to master and new release later.

haugene:
Btw, the default.ovpn files aren't just regular files. They are symlinks.
You create them by cd'ing to the same directory as the config files and doing `ln -s France.ovpn default.ovpn`. Fixed it now, pushing change to dev.

Issue #554, opened 2018-07-09
Adding New and Updating ALL VPN Provider Files


mrjackyliang:
I have seen so many issues related to supporting and updating various VPN provider files. Looking to keep a top list here, so everyone can see the providers that are pending updates and additions.

__If you have any updates or additional suggestions please @mrjackyliang or @haugene, so we can add it into this list.__

### Updates Pending
* FrootVPN (#234)
* Windscribe VPN (#569)
* Newshosting (#629)

### Additions Pending
* Invisible Browsing VPN (#502)
* ProtonVPN (#378)
* IronSocket (#287)
* VPN Unlimited (#259)
* CyberGhost (#597)
* Forcefield (#641)
* TweakNews VPN (#839)
* GooseVPN (#738)
* SurfShark VPN (#734)
* Air VPN (#817)
* LeVPN (#813)

lamboid:
any news on VPN Unlimited being added?

coolboiime:
@lamboid We're missing ovpn files! If you have those images, would you mind contributing? Thanks!

iamaburrama:
[en_configs.zip](https://github.com/haugene/docker-transmission-openvpn/files/2268220/en_configs.zip)

Easynews is my provider.  Attached ovpn files.  TY!

lamboid:
I thought the VPNunlimited ovpns are tied to our personal accounts. Am I wrong?

lamboid:
[VPNunlimited.zip](https://github.com/haugene/docker-transmission-openvpn/files/2342452/VPNunlimited.zip)

Took a look inside them - only certs and some config!
Attached are all the torrenting server ovpn files

haugene:
Hey @pejotll. Yes, if you can provide a zip file with ovpn files I can have a look at adding it.
Also seeing that @lamboid uploaded some config files, I'll check those as well.

Want to run through this issue and update/add what I can during the holdidays so we can close it :)

pejotll:
Hey @haugene, let me know if that's everything you need. 
[ProtonVPN_server_configs.zip](https://github.com/haugene/docker-transmission-openvpn/files/2690103/ProtonVPN_server_configs.zip)
[ProtonVPN_country_configs.zip](https://github.com/haugene/docker-transmission-openvpn/files/2690104/ProtonVPN_country_configs.zip)

markedphillips:
[...]

PR #624, opened 2018-10-17
Added trust.zone provider

Maximillion195:


haugene:
Hey. Please also add TrustZone in the list of providers in the README.

haugene:
Hello again. Sorry I'm doing this a bit scattered. Looking through it again I see a couple of more things.
This config has ping and ping-restart defined. We're recommending to use ping and ping-exit (documented [here](https://github.com/haugene/docker-transmission-openvpn#restart-container-if-connection-is-lost)).

The reason for ping-exit is that the openvpn restart has not been working well in the container and it just freezes. If we do ping-exit, openvpn should exit and the --restart=unless-stopped flag to Docker will make it restart the whole container instead of just openvpn.

Could you also remove those config lines, and I'll merge :)

Maximillion195:
Hey, not a problem mate. Makes sense :) I have removed those lines.

haugene:
Allright, merging. Thanks for contributing and let me know how testing it from the dev tag goes. Will do a merge to master and a new release - probably today :)

mrjackyliang:
Looks like this has been added, marking into #554!

Issue #41, opened 2015-11-18
Synology NAS instructions


timkelty:
After some tinkering I got this working on a Synology NAS (412+).
Just wanted to share in case anyone else was trying.

Here's what I had to do:

- Add `--dns 8.8.8.8 --dns 8.8.4.4` to the `docker run` command (as noted in README)
- Run the following script on startup.
    - In my case, this meant putting the script here (making sure it is executable): `/usr/syno/etc.defaults/rc.d/S01tun.sh`

```
#!/bin/sh
  
# Create the necessary file structure for /dev/net/tun
if ( [ ! -c /dev/net/tun ] ); then
  if ( [ ! -d /dev/net ] ); then
    mkdir -m 755 /dev/net
  fi
  mknod /dev/net/tun c 10 200
fi

# Load the tun module if not already loaded
if ( !(lsmod | grep -q "^tun\s") ); then
  insmod /lib/modules/tun.ko
fi
```

haugene:
Nice! Thanks for the input. I'm guessing this is related to issue https://github.com/haugene/docker-transmission-openvpn/issues/26

I'll add a reference to it from the README the next time I'm doing some changes. Maybe expand the known issues section to be a Known Issues/FAQ/Tips and Tricks kind of thing.

kornflex:
Hello,

In Transmission, I tried to download a torrent : Error adding "http://URL/xxxxx.torrent": gotMetadataFromURL: http error 503: Service Unavailable

I add --dns command in docker run:
docker run --dns 8.8.8.8 --dns 8.8.4.4 --privileged -d -v xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx

But, Synology doesn't want these parameters : --dns options non prises en charge

Any ideas ?

Thks

haugene:
The --dns options will be used to write the containers /etc/resolv.conf
You could probably make your own resolv.conf file on the host and mount it with -v /your/resolv.conf:/etc/resolv.conf

You should maybe instead mount something to /etc/resolvconf/resolv.conf.d/head that will be prepended to your /etc/resolv.conf when resolvconf generates it.

Or something like that. It might fix the issues. And if it don't I would just exec into the container (docker exec -it <container-name> bash) and change stuff manually and test using dig/ping until it resolves the hostnames. When you change the /etc/resolvconf/resolv.conf.d/head file you can make resolvconf reload the config based on your changes by running "resolvconf -u"

[...]

PR #627, opened 2018-10-18
Fix unpriviligied user to use tun device under docker

ducseb:
I've added a chmod to tun device to allow unprivileged docker containers to use tun device and send traffic through the vpn.
Related to this issue: https://github.com/haugene/docker-transmission-openvpn/issues/41


haugene:
Great, thanks for contributing 👍

PR #640, opened 2018-11-13
Update so original UI button works in 

Dulanic:
Added copy of index so that. Without index.original, the original ui button on web control UI does not work. I am not 100% on if this is best in the docker file or somewhere else, but seemed best and simplest method.


Dulanic:
Proof of test:
image
image
image

PR #644, opened 2018-11-18
Update NordVPN configs and update script

davestephens:
This PR updates the NordVPN configs and update script. The update script has been tweaked to set a default to a config that definitely exists, previously no check was made and the default symlink was pointing to a non-existent config.


davestephens:
Ping....

haugene:
Hey @davestephens, sorry about the delay.
I think I've already made similar changes on the dev branch in 14695cf975f91b2072217538e9ac1875b1c92aba. These should be merged to master I guess.

Will change base to dev and see if there are still a big diff.

davestephens:
No probs, thanks for checking out my PR. 

Aside from the out of date configs, my change to the update script automatically picks a valid config from the list rather than requiring one to be specified. The logging is also beefed up a bit. Appreciate that's not easy to see with so many changes! :)

haugene:
Allright, seems like good changes to the scipt then 👍 
These config-PR's are so lengthy. I usually just validate a handfull of configs and then make sure that all changes are in the same folder (provider). Will you update the PR to include only your script changes? And if there's still some config updates of course. But would be great if you'd rebase/squash the commits so we reduce the number of 10.000-files commits ;)

davestephens:
I did consider creating a PR that removes the config from the repo and makes it so that they are pulled and tweaked when the container is built - but annoyingly some providers hide their config away behind credentials which would mean everything was left a bit inconsistent :-/...

haugene:
Yeah, I've had a dream about that as well. But as you say, it will be inconsistent between providers. And it also places a build time dependency to all providers download-pages. If any of those are down (or goes out of business) the image won't build until it's back up or removed from the list. If we not ignore failing downloads, which would be a random thing as well.

I guess I'll clean the git tree at some point. We don't need revision history for the configs I guess.
And files being checked in means that people can verify what they're running. There is some trust in the picture here as well, the contents of the configs are important and should be explicitly visible.

davestephens:
The alternative is that we expect the user to provide configs to the container, the configs could then be manipulated for replacements before being loaded. This solves the trust issue and would clean the repo significantly, but would mean a little extra setup for the user.

It does mean however, that upgrading the container is less likely to break a working setup - as was the case for me which caused me to create this PR in the first place.

niXta1:
Could something like this be done for NordVPN? Choose where and what, then let NordVPN's API decide?
https://github.com/azinchen/nordvpn

haugene:
Hey again. Grabbed the updateConfig.sh changes and just committed them directly to dev. These NordVPN PR's are so huge so I'll skip an extra 10k file update as the repo is growing quickly.

Had some issues running it on my mac, but I guess that's due to the version of Bash or something else. The changes look good and as long as it runs on Linux that's the most important part. Thanks again, closing this now.

PR #648, opened 2018-11-21
Add transmission-cli package

### LINKED ISSUE #647: alpine build missing transmission-remote
The Alpine Dockerfile does not install the transmission-cli package which contains the transmission-remote binary which is called by /etc/openvpn/updatePort.sh

sedlund:
transmission-cli package contains transmission-remote which is called by /etc/openvpn/updatePort.sh

fixes #647


haugene:
Great, thanks for the fix. Must have been included in the daemon at some point - or has it never worked? Ah well... Sorry for the delay, but merging to dev now. Bundling up some more changes and making a new release.

Thanks again for contributing!

sedlund:
@haugene Unsure if it worked before.  My guess is no, as transmission-remote is usually part of a separate package in most distros.

PR #665, opened 2018-12-16
Fix builds on ARM arch but keeping ARM cross builds.

AdrienKuhn:
Fix #638.

@haugene The Dockerfile location in Docker Hub settings will need to be updated to `DockerfileCrossBuild.armhf` for all `armhf` tags

Issue #638, opened 2018-11-10
Build .armhf image on raspberry pi 


steveman1982:
Luckily :latest-armhf works for me. But before stumbling upon that solution I tried to build the image, which resulted in this:

docker build -t transmission-openvpn -f Dockerfile.armhf .
Sending build context to Docker daemon  51.68MB
Step 1/13 : FROM resin/rpi-raspbian:stretch
 ---> 21dc8fc1377f
Step 2/13 : MAINTAINER Kristian Haugene
 ---> Running in db295b976b88
Removing intermediate container db295b976b88
 ---> 96ba3f58f537
Step 3/13 : RUN [ "cross-build-start" ]
 ---> Running in 9be13a5e398d
standard_init_linux.go:190: exec user process caused "exec format error"
The command 'cross-build-start' returned a non-zero code: 1

Tried this on master and v2.4.

dutch1e:
I appear to be having the same issue with :latest-armhf on a Nanopi K2 running armbian. Docker pulls the image fine and creates the container, however the container is continually restarting with error 132, and no text in the container logs.

Output from the following commands here > https://pastebin.com/K7nveEcJ
docker ps -a
uname -a
cat /proc/cpuinfo

dutch1e:
Output of build from dockerfile:
docker build https://raw.githubusercontent.com/haugene/docker-transmission-openvpn/master/Dockerfile.armhf
Sending build context to Docker daemon  6.656kBs://raw.githubusercontent.com/haugene/docker-transmission-openvpn/master/Dockerfile.armhf  5.085kB
Step 1/13 : FROM resin/rpi-raspbian:stretch
 ---> 21dc8fc1377f
Step 2/13 : MAINTAINER Kristian Haugene
 ---> Running in 4afb53f40808
Removing intermediate container 4afb53f40808
 ---> cf4e6269b3b9
Step 3/13 : RUN [ "cross-build-start" ]
 ---> Running in c0d1700e63a7
standard_init_linux.go:190: exec user process caused "exec format error"
The command 'cross-build-start' returned a non-zero code: 1

haugene:
[...]

PR #674, opened 2018-12-30
Fix for ARM based devices (like Raspberry Pi)

NickB23:
Removed RUN [ "cross-build-start" ] & RUN [ "cross-build-end" ] from all .armhf files in order to work with ARM based devices

### DIFF
--- a/Dockerfile.armhf
+++ b/Dockerfile.armhf
@@ -1,8 +1,6 @@
 FROM resin/rpi-raspbian:stretch
 MAINTAINER Kristian Haugene
 
-RUN [ "cross-build-start" ]
-
 VOLUME /data
 VOLUME /config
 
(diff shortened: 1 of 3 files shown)


NickB23:
Good idea! I'll make the change

NickB23:
Made the change. All the following lines in .armhf files
`RUN [ "cross-build-start" ]`
 are changed into:

`#Uncomment the line below if you are building on a non-ARM (x86) device
#RUN [ "cross-build-start" ]`

and

`RUN [ "cross-build-end" ]`
is changed into

`#Uncomment the line below if you are building on a non-ARM (x86) device
#RUN [ "cross-build-end" ]`

haugene:
Hi guys. See #638 and the fix @ #665. This is merged to dev, and will be available on the master/latest tag very soon. I've just been a bit slow over the holidays.

We've created duplicates of the .armhf Dockerfiles as the Docker Hub Automated Builds still need it but it doesn't work when you actually build it on an armhf device.

Closing this for now and hope it solves your problems. Feel free to continue discussion if you have other suggestions.

Issue #229, opened 2017-04-02
update-resolv-conf outdated/openresolv necessary?


bspinner:
Hi, 
as I've mentioned in #225 , I had to make some modifications, so update-resolv-conf gets called.

While doing that, I've noticed that `update-resolv-conf` seems to depending on `openresolv`, which isn't installed by default in this image. Is `resolvconf` just misbehaving for me or is `openresolv` a missing dependency indeed?

Second, I wondered whether the update-resolv-conf script comes from [this](https://github.com/masterkorp/openvpn-update-resolv-conf) repo. Is that the case? If so, were modifications made or is the script in the mentioned repo just more up to date and should be merged into this repo?

haugene:
Hi,

Had a look at this now. And you're right that `openresolv` is missing in the image. As for the `update-resolv-conf`script, that is added during the installation of openvpn and not modified from this project.

The config files for PIA does not reference running of `update-resolv-conf` and the question then becomes if we should add your changes to the image so that they're run by all providers or make it possible to customize this using run parameters.

Would it be a possible solution to add installation of  `openresolv` to this image and making it possible to override the TRANSMISSION_CONTROL_OPTS set in openvpn/start.sh? Then you could start the container with something like `TRANSMISSION_CONTROL_OPTS=--up '/etc/openvpn/update-resolv-conf && /etc/transmission/start.sh'`?

bspinner:
> Would it be a possible solution to add installation of openresolv to this image and making it possible to override the `TRANSMISSION_CONTROL_OPTS` set in `openvpn/start.sh`? Then you could start the container with something like `TRANSMISSION_CONTROL_OPTS=--up '/etc/openvpn/update-resolv-conf && /etc/transmission/start.sh'`?

That should do it.

haugene:
@h45rd. Thoughts? The oldest open issue. Not sure what to do with it.
Are you still using this container @bspinner? Any input two years later? ;)

h45rd:
Hmm.. not sure what would be best on a global scale.
Personally, I prefer setting a third party DNS myself to leaving DNS resolution up to the VPN provider. Both from a security and a "container maintainability" standpoint.

During the search for a possible new VPN provider in the last couple of days, I tested several with this container.
And man let me tell you.. they all do their own thing when it comes to pushing options to the clients. Some push nothing, others push a plethora of usless stuff, like IPv6 routes and gateways that are not working, etc.
[...]

PR #676, opened 2019-01-07
Updated all Mullvad configs from scratch

jfessard:
I noticed that mullvad_us-ca_udp wasn't working for me and port numbers were off by 1 for some reason for a bunch of them, so I went ahead and grabbed all possible configs, renamed them to keep the convention, and ran them through adjustConfigs.sh as required. There was still those up/down lines that wanted to call `update-resolv-conf` that were causing issues, they were removed. I hope that was the right way to do it.

It is a massive pull request. I basically emptied out the mullvad folder and added everything new. Whitespace is as it is in the original config files.

Note that it seems that the configs have dropped comp-lzo, which apparently is on the way to being deprecated.
They bundle a CRL, but don't use it or specify it in the configs. (I see it was used in the old files)
Some endpoints were deprecated as well (the generic Canada one for instance).


haugene:
Hey @jfessard. This looks good to me, thanks a lot for contributing and helping keeping configs up to date 👍 

As for the update-resolv-conf, I'm not sure. I'll merge this, but I'm not sure that we want to remove that line for all providers (in adjustConfigs.sh). I might remove that line later.

Thanks again, and if you could verify that it works when it's all built that would be great.

jfessard:
Thanks for the merge, happy to contribute! 

### Building the container
Actually I did try and succeeded building the container locally before I submitted the PR, I'm actually running it right now. (until it gets merged to `master` I guess. What's the timeline in general for those?)

### With update-resolv-conf
I tried adding `up /etc/openvpn/update-resolv-conf` and the matching down command (since it was already in the image already anyway, and it's the exact same file as what mullvad provides). 

It still connected, but from the logs, `update-resolv-conf` got run instead of `tunnetUp.sh` which messes up the init since it never starts transmission. We should probably combine the two, I'm worried not running it is actually a bad thing (DNS leaks and such).

```
Thu Jan 10 23:39:02 2019 Multiple --up scripts defined.  The previously configured script is overridden.
Thu Jan 10 23:39:02 2019 Multiple --down scripts defined.  The previously configured script is overridden.
[...]
Thu Jan 10 23:39:08 2019 /etc/openvpn/update-resolv-conf tun0 1500 1552 X.X.X.X 255.255.0.0 init

```

haugene:
Yeah, it's the oldest open issue around: #229. I haven't dug into to the issue enough to know if we can find a good solution for all providers. Would very much appreciate some help finding and implementing a fix.

It could be that we just add update-resolv-conf in our up/down scripts, then it will be for all providers. Or we can dynamically evaluate the ovpn file before starting and if there's a update-resolv-conf option then we add it (although it sounds a bit messy).

Up for anything really :) And as for the merge to master, I did it yesterday as well. Changes should be on the latest tag. The timeline for it varies - sometimes I'm too slow. But I like having a dev branch that I can merge into less critically, and do numbered releases and tags on master so that people can revert back if an error is introduced. Pros and cons of having that pit stop I guess.

Issue #673, opened 2018-12-28
Ivacy certificates (and/or configuration) are outdated


dieselburner:
[Here ](https://support.ivacy.com/wp-content/uploads/2018/09/OpenVPN-Configs.zip)is location of new OpenVPN configuration files (as per [this link](https://support.ivacy.com/kb/openvpn-files-windows-routers-ios-linux-and-mac/)). Sorry, too busy atm to bake a pr.

haugene:
Hey. Thanks for the heads up and link. Updated them now, can you test if it works?
Will be available on the dev tag soon, and I think I'll just merge it to master later today so it's available on the latest tag as well. Getting stuff done after the holidays :)

haugene:
#554

PR #677, opened 2019-01-10
Updated IVACY directory

nravanelli:
Updated the IVACY configurations/certifications.

First pull request on github, but I hope I did it right....


haugene:
Hey @nravanelli. First of all, thanks for wanting to contribute to the project - I appreciate it!

I've haven't done anything with the project for a couple of weeks and sat down to go through all the notification emails now. I started with the oldest and there was an issue (#673) on this from two weeks ago that I then did first. And now I see that you were on it.

Anyways. This should now be fixed, and it would be great if you could test it. It's on the dev tag of the image, but I'll merge it to master/latest tag soon.

As for your PR. It looks pretty good, but there are a couple of things that would have to be fixed.
So you know for the next time, these three lines in the old configs:
```
ca /etc/openvpn/ivacy/ca.crt
tls-auth /etc/openvpn/ivacy/Wdc.key 1
auth-user-pass /config/openvpn-credentials.txt
```
are customisations for the configs to work in this container. We have a lot of providers and they are put in their own folders. Therefore the ca.crt and Wdc.key references need to point there.
Also, when you use OPENVPN_USERNAME and OPENVPN_PASSWORD variables the credentials are stored in a file /config/openvpn-credentials.txt. This needs to be passed to the auth-user-pass option so it doesn't ask for a username or password on the command line when starting.

I hope my update works, and if not please follow up in #673

PR #680, opened 2019-01-11
Monitor DNS resolution

sstassin:
Hi Haugene, as we have discussed about this;.. i v'e included a simple script that :

- every 10 minutes
- do a simple curl (with header only) to htts:\\www.google.com
- if failed try to do this 3 times with a delay of 1 minute
- if all the tries fails, it kill the openvpn client (causing a restart if the container was started with this option)
Feel free to integrate my first GitHub pullrequest on GitHub :)  and i am happy if it could help.
I have also noticed that using OpenDNS server are working better than the Google DNS
Regards

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,7 @@ RUN apt-get update \
     && tar -xvf src.tar.gz -C /opt/transmission-ui/transmission-web-control/ \
     && rm src.tar.gz \
     && git clone git://github.com/endor/kettu.git /opt/transmission-ui/kettu \
-    && apt-get install -y tinyproxy telnet \
+    && apt-get install -y tinyproxy telnet nano \
     && wget https://github.com/Yelp/dumb-init/releases/download/v1.2.0/dumb-init_1.2.0_amd64.deb \
     && dpkg -i dumb-init_1.2.0_amd64.deb \
     && rm -rf dumb-init_1.2.0_amd64.deb \
@@ -34,7 +34,8 @@ RUN apt-get update \
     && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* \
     && groupmod -g 1000 users \
     && useradd -u 911 -U -d /config -s /bin/false abc \
-    && usermod -G users abc
+    && usermod -G users abc 
+ #   && echo '*/10 * * * * root /etc/openvpn/checkdns.sh' >> /etc/crontab
 
 ADD openvpn/ /etc/openvpn/
 ADD transmission/ /etc/transmission/
@@ -130,6 +131,19 @@ ENV OPENVPN_USERNAME=**None** \
     DROP_DEFAULT_ROUTE= \
     WEBPROXY_ENABLED=false \
     WEBPROXY_PORT=8888
+
+# Exporting the checkdns.sh as cronttab
+# https://stackoverflow.com/questions/37458287/how-to-run-a-cron-job-inside-a-docker-container
+# Add crontab file in the cron directory
+ADD openvpn/checkdns-crontab /etc/cron.d/checkdns-crontab
+# Give execution rights on the cron job
+RUN chmod 0644 /etc/cron.d/checkdns-crontab
+# Apply cron job
+RUN crontab /etc/cron.d/checkdns-crontab
+
+# Create the log file to be able to run tail
+RUN touch /var/log/cron.log
+
 
 # Expose port and run
 EXPOSE 9091
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,7 @@ RUN apt-get update \
     && tar -xvf src.tar.gz -C /opt/transmission-ui/transmission-web-control/ \
     && rm src.tar.gz \
     && git clone git://github.com/endor/kettu.git /opt/transmission-ui/kettu \
-    && apt-get install -y tinyproxy telnet nano \
+    && apt-get install -y tinyproxy telnet \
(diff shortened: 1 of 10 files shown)


Dockerfile
```diff
@@ -129,7 +129,12 @@ ENV OPENVPN_USERNAME=**None** \
     TRANSMISSION_WEB_HOME= \
     DROP_DEFAULT_ROUTE= \
     WEBPROXY_ENABLED=false \
-    WEBPROXY_PORT=8888
+    WEBPROXY_PORT=8888 \
+    MONITORDNS_ENABLED=true 
+
+#Making the checkdns script executables
+RUN chmod 755 /etc/openvpn/monitor-checkdns.sh
+RUN chmod 755 /etc/openvpn/checkdns.sh
```
jfessard: It's weird that the `chmod` would have to be here, where is that command run for all the other scripts? (like `start.sh`?)

README.md
```diff
@@ -214,6 +214,16 @@ Remember to add a port binding for your selected (or default) port when starting
...
 |`WEBPROXY_PORT` | Sets the listening port | `WEBPROXY_PORT=8888` |
 
+### DNS Monitoring configuration options
+
+This container also can monitor if DNS resolution is OK throught the VPN connection, to avoid some DNS issues
+with some Providers. If the www.google.com could'nt be resolved, the openvn client is killed (true by default)
+
+| Variable | Function | Example |
+|----------|----------|-------|
+|`MONITORDNS_ENABLED` | Enables the DNS resolution monitoring | `MONITORDNS_ENABLED='true`|
```
jfessard: Stray character
```suggestion
|`MONITORDNS_ENABLED` | Enables the DNS resolution monitoring | `MONITORDNS_ENABLED=true`|
```
I like the var table idea!

haugene:
I agree with @jfessard's comments. The commits should be squashed, and chmod should be done to the file before it's added. So it's not a part of the image build.

That being said. Docker has support for native health checks. I just read a bit about them, but it seems they can be a good fit for our need. https://docs.docker.com/engine/reference/builder/#healthcheck

Basically we could make a health-check script that does DNS checks to begin with, but could also be expanded. From docker's side there is no automatic handling if the check fails (we want to exit). It was in the original PR https://github.com/moby/moby/pull/22719, but got removed for a future solution with restart-policy.

For now we could have "HEALTH_CHECK_ENABLED" and "HEALTH_CHECK_EXIT_ON_UNHEALTHY" variables that will decide if the tests should be run, and if an error should mark the container unhealthy or exit when it fails. This could be achieved by killing the process.

Then we would have an all-in-one solution that also kills the process, health check run as root so that should be OK. And users have the possibility to run for example this image https://github.com/willfarrell/docker-autoheal to handle the un-healthy containers.

We should also make the checks run through the tunnel device. And could we do with `dig`or `ping` instead of curl?

Thoughts?

h45rd:
Thanks for the hint about docker health checks.
Been reading up on them and sounds like they could be really useful for this type of container.

Have not been following the repo too closely in a while, so please bare with me if the answer might be self evident :)
But why is there a need to separately test for DNS resolution during runtime?
Are there instances it would break before OpenVPNs two-tiered inactivity disconnect triggers anyway?


On the topic of monitoring the container:
Since my provider's firewall sometimes randomly closes forwarded ports, I have been toying with a hacky "port watcher" script to check and restart the container on such occasions.
Think this kind of functionality would make a nice addition to a container-wide health check.
But wasn't yet able to implement in a non-hacky, globally applicable way that covers all test cases :/
Might open a new topic for brainstorming regarding this, if there is any interest.

Dockerfile
sstassin: > I agree with @jfessard's comments. The commits should be squashed, and chmod should be done to the file before it's added. So it's not a part of the image build.
> 
> That being said. Docker has support for native health checks. I just read a bit about them, but it seems they can be a good fit for our need. https://docs.docker.com/engine/reference/builder/#healthcheck
> 
> Basically we could make a health-check script that does DNS checks to begin with, but could also be expanded. From docker's side there is no automatic handling if the check fails (we want to exit). It was in the original PR [moby/moby#22719](https://github.com/moby/moby/pull/22719), but got removed for a future solution with restart-policy.
> 
> For now we could have "HEALTH_CHECK_ENABLED" and "HEALTH_CHECK_EXIT_ON_UNHEALTHY" variables that will decide if the tests should be run, and if an error should mark the container unhealthy or exit when it fails. This could be achieved by killing the process.
> 
> Then we would have an all-in-one solution that also kills the process, health check run as root so that should be OK. And users have the possibility to run for example this image https://github.com/willfarrell/docker-autoheal to handle the un-healthy containers.
> 
> We should also make the checks run through the tunnel device. And could we do with `dig`or `ping` instead of curl?
> 
> Thoughts?

Ok, i agree with that, i will do it but give me some times, i am actually looking for a new job :)

sstassin:
> chmod should be done to the file before it's added. So it's not a part of the image build.
I am sorry but i don't know how to do this ?? Directly on the GitHub repo ?
(Sorry i am a noob in Docker Container maker).... but improving my self :)

haugene:
I think this PR is largely superseded by #708. It does one more thing though.
There have been cases where the ping-exit doesn't seem to work in forcing Docker to restart the container. This might be due to missing pull-filter in the config and these arguments getting overridden. But adding this just for the `kill` statement might be overkill (awesome pun intended).

So I think we will close this for now, and instead I propose for another flag to be set so that the health check script can kill openvpn if it errors for too long. It runs as root and will have the permission.

Issue #575, opened 2018-08-10
Proxy not working on Raspbian


IvanGolovko3k:
When running on Raspbian the proxy won't work because the `tinyproxy.conf` cannot be found.

```
STARTING TINYPROXY
Setting tinyproxy port to 8888
sed: can't read /etc/tinyproxy.conf: No such file or directory
sed: can't read /etc/tinyproxy.conf: No such file or directory
Starting Tinyproxy lightweight HTTP proxy daemon: tinyproxy.
Tinyproxy startup script complete.
```

I inspected the running container and noticed that the location is not in `/etc/tinyproxy.conf` but in `/etc/tinyproxy/tinyproxy.conf`.

I modified the `/tinyproxy/start.sh` and changed the location to `PROXY_CONF='/etc/tinyproxy/tinyproxy.conf'`. 

I rebuild the image, started the container and now it's working.

```
STARTING TINYPROXY
Setting tinyproxy port to 8888
Starting Tinyproxy lightweight HTTP proxy daemon: tinyproxy.
Tinyproxy startup script complete.
```

I'm running Raspbian GNU/Linux 9 (stretch).

Since I'm not too familiar with shell I'll leave the fixing to someone else.

nicolas-albert:
Same here, I have copied the `/opt/tinyproxy` on my Pi, fixed and run it using a `-v <myfolder>:/opt/tinyproxy`.

haugene:
This should now be fixed. #732 
Sorry it took a while, but people post a lot of issues and I'm mainly focusing on the regular ubuntu version of the image. Anyways. Closing this now.

PS: It's currently only available on the dev tag. will be pushed to latest soon. There's also been some build issues on Docker Hub for the armhf image, this is also fixed on dev. So `dev-armhf` is the one to use for now.

PR #681, opened 2019-01-13
tinyproxy conf

dipietro-salvatore:
wrong tinyproxy conf file path

### DIFF
--- a/tinyproxy/start.sh
+++ b/tinyproxy/start.sh
@@ -3,7 +3,7 @@
 # Source our persisted env variables from container startup
 . /etc/transmission/environment-variables.sh
 
-PROXY_CONF='/etc/tinyproxy.conf'
+PROXY_CONF='/etc/tinyproxy/tinyproxy.conf'
 
 set_port()
 {


haugene:
Hey @dipietro-salvatore and thanks for wanting to contribute to this project.

Just wanted to verify this change before merging and it doesn't seem to me that this is the wrong config path:
```
➜  ~ docker pull haugene/transmission-openvpn
Using default tag: latest
latest: Pulling from haugene/transmission-openvpn
Digest: sha256:160a1faf3dbe2ed3471da90590c0547b065d1b337c87aef10c4fda7b48d4919f
Status: Image is up to date for haugene/transmission-openvpn:latest
➜  ~ docker run --rm -it haugene/transmission-openvpn bash
root@06790072633a:/# ls -la /etc | grep tinyproxy
-rw-r--r-- 1 root root                10193 Feb 22  2010 tinyproxy.conf
root@06790072633a:/#
```

From what I can see, in the latest version of this image. There is no /etc/tinyproxy/ folder and the tinyproxy.conf is located directly under /etc/. Are you seeing something else on your system?

dipietro-salvatore:
Probably I was using an old version of it.
The main problem was that the tinyproxy server was not configured correctly
and I need to change the file manually every time.
Can you just check that in the new version the tinyproxy configuration is
set correctly and that the "Allow" rules are not commented?

Thanks

haugene:
I see. I'll have a look this weekend 👍

Xyv:
Hi,
I have the same problem, this seems to be related to issue https://github.com/haugene/docker-transmission-openvpn/issues/575.
@dipietro-salvatore are you running this on raspbian?
@haugene Could we pass the path to the tinyproxy.conf file as an env variable when running the image?

dipietro-salvatore:
yes, I am!

h45rd:
The host OS should not matter.
But the the ARM docker image used on a raspberry is based on resin/rpi-raspbian:stretch instead of ubuntu.
Guess you ran into a package anomaly that requires a slight patch to the _tinyproxy/start.sh_ to check for the location of tinyproxy.conf.

In the meantime you could just docker-mount an external tinyproxy.conf to the right location for a quick fix, e.g.: `-v /docker/tinyproxy.conf:/etc/tinyproxy/tinyproxy.conf`

haugene:
Hey again. So my last message was a bit optimistic and I did not get the time.
Don't think I would have seen the connection to #575 either, so thanks @Xyv!

The problem is then that the conf file is placed in another location on armhf. So while this PR in its current state would fix it for armhf it would break on the default ubuntu based image as well as on alpine. I suggest that the tinyproxy/start.sh script runs a check for if the file exists at the ubuntu/alpine location first, then checks the armhf path and exits with a good error message if it can't find it.

Seems from #575 that the init.d start script on raspbian is in the same place, so we could leave that?
In the meantime, @h45rd solution of mounting the config file will do the trick 👍

Xyv:
Indeed, in the meantime mounting the config file to the right location works!
You do have to edit it manually before mounting though as the `sed -i` command used in `start.sh` does not work within docker (you can't change a mounted file's inode).

neil123454:
Hello, could someone help me with this. I am on raspberry pi with OMV. I have used the docker GUI to pull the latest arm images. I have tried and failed to use the command line to remount the file.

h45rd:
Need more information in order to help.
Failed how? What did you try? What happened? Any error message?

neil123454:
On OMV I cant even find the place in the GUI to run the command

h45rd:
Never used OMV with the docker plugin, so I am not much use there.
The [section on bind mounts](https://forum.openmediavault.org/index.php/Thread/11296-OpenMediaVault-Docker-GUI-Plugin-Usage/) in the plugin guide seems to suggest that you can add them, though. Screenshots only for registered users, so I can't give you the exact form field.

What the sample command I gave above does, is bind mount a local copy (on the host) of tinyproxy.conf to where the application expects it to be within the container.
You should be able to replicate this with the OMV GUI.

If you need more detailed instructions on how to accomplish this using OMV, your chances for a conclusive reply are better on the OMV board, though.

neil123454:
Ok so i found a random tinyproxy config file online but am unable to get it to work. 


STARTING TRANSMISSION
NO PORT UPDATER FOR THIS PROVIDER
STARTING TINYPROXY
Setting tinyproxy port to 8888
sed: cannot rename /etc/sedg5sCpl: Device or resource busy
sed: cannot rename /etc/sedDxR8Zp: Device or resource busy
Starting Tinyproxy lightweight HTTP proxy daemon: tinyproxy.
Tinyproxy startup script complete.
Tue Mar  5 19:49:59 2019 Initialization Sequence Completed

I get this error

h45rd:
I would suggest you just use the tinyproxy.conf from within the container (see [docker cp](https://docs.docker.com/engine/reference/commandline/cp/)) instead of a random one :)
Did you manually edit the file as suggested by Xyv in his last post? Then you can ignore the error and tinyproxy should work as per your configuration.


The issue with the startup script would be a quick fix. But I would prefer a more elegant solution than hardcoding tests for various possible config file paths. Is `whereis` available on all supported platforms?

neil123454:
I have tried to remap as follows
Container path: /etc/tinyproxy.conf
Host path: /etc/tinyproxy/tinyproxy.conf

I am now getting this error:

STARTING TRANSMISSION
NO PORT UPDATER FOR THIS PROVIDER
STARTING TINYPROXY
Setting tinyproxy port to 8888
sed: couldn't edit /etc/tinyproxy.conf: not a regular file
sed: couldn't edit /etc/tinyproxy.conf: not a regular file
Starting Tinyproxy lightweight HTTP proxy daemon: tinyproxy.
Tinyproxy startup script complete.
RTNETLINK answers: File exists
Tue Mar  5 20:34:50 2019 ERROR: Linux route add command failed: external program exited with error status: 2
Tue Mar  5 20:34:50 2019 Initialization Sequence Completed

I haven’t been able to understand what XvY was saying or how to do it. I am still new to Linux.

Neil

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

neil123454:
I'm not sure what you mean about whereis

h45rd:
Okay, let me explain then :)
The tinyproxy [start script](https://github.com/haugene/docker-transmission-openvpn/blob/master/tinyproxy/start.sh) has two `sed` calls that are used to adjust tinyproxy.conf, see lines 28 and 38.
Those do not work in your case because your tinyproxy.conf is a bind mount and sed can't edit it in place (-i).

So to get a working tinyproxy.conf, you need to manually replicate (on the host) the edits the script would make via sed.
The first call sets the tinyproxy listening port to the one specified via $WEBPROXY_PORT.
The second allows incoming connections from all IPs by commenting out all lines that start with `Allow`

For further reference just have a look at the tinyproxy.conf itself and `man tinyproxy`. It is well documented.

neil123454:
OK so how would i be able to edit to start.sh so that it used the correct location for the tinyproxy file

haugene:
@neil123454 Seems from your logs that the sed statement fails, and this is the one that changes the config to allow for connections from all hosts. I just committed a fix to the dev branch that should make tinyproxy work out of the box on both armhf and amd64 installations of tinyproxy.

Closing this PR, and follow up the config issue in #732

PR #697, opened 2019-02-04
Adds pre-start for openvpn

wilmardo:
I run this image on Kubernetes and I would like to create a tun device inside the container on startup, like this:

```bash
#!/bin/bash
mkdir -p /dev/net
mknod /dev/net/tun c 10 200
chmod 600 /dev/net/tun
```

Now I override the command of the container but it would be nice if this can be handled as the transmission scripts, like this:
```yaml
command: ["/bin/bash"]
args: ["-c", "/scripts/openvpn-pre-start.sh && dumb-init /etc/openvpn/start.sh"]  
```
I only added the pre-start because that one seems the most relevant to me but I can add the post-start, pre-stop etc. if necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,7 @@ Once /scripts is mounted you'll need to write your custom code in the following 
 
 | Script | Function |
 |----------|----------|
+|/scripts/openvpn-pre-start.sh | This shell script will be executed before openvpn start |
 |/scripts/transmission-pre-start.sh | This shell script will be executed before transmission start |
 |/scripts/transmission-post-start.sh | This shell script will be executed after transmission start |
 |/scripts/transmission-pre-stop.sh | This shell script will be executed before transmission stop |
--- a/openvpn/start.sh
+++ b/openvpn/start.sh
@@ -13,6 +13,14 @@ fi
 
 echo "Using OpenVPN provider: ${OPENVPN_PROVIDER}"
 
+# If openvpn-pre-start.sh exists, run it
+if [ -x /scripts/openvpn-pre-start.sh ]
+then
+   echo "Executing /scripts/openvpn-pre-start.sh"
+   /scripts/openvpn-pre-start.sh "$@"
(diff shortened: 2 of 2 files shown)


haugene:
Interesting. Creating the tun device inside the container alleviates the need to mount the device? You still need elevated permissions like NET_ADMIN for the container?

I will gladly merge the pre-start functionality, but I'm thinking it could be a good idea to include the tun creation script as well. We could set is as an "opt in" feature, but if it works well that will be standard quick I guess. It's just a hassle mounting the device from the host.

For my ubuntu/debian servers I've never had an issue, but there's been some issues on different NAS servers.

wilmardo:
> Interesting. Creating the tun device inside the container alleviates the need to mount the device? You still need elevated permissions like NET_ADMIN for the container?

The elevated permissions are still needed indeed since it has to create the tunnel.

 > I will gladly merge the pre-start functionality, but I'm thinking it could be a good idea to include the tun creation script as well. We could set is as an "opt in" feature, but if it works well that will be standard quick I guess. It's just a hassle mounting the device from the host.

Should I drop the pre-start and create a tun creation script as opt in? Or should I expand this to all the pre/post option transmission has and create and opt-in separately?

haugene:
An openvpn-pre-start.sh seems like a good addition to me in any case. So I'll merge this now and then if you would add the opt-in separately that would be great 👍 

As for the other pre/post scripts transmission has. I'm not sure if they would make sense for openvpn. The openvpn/start.sh script ends up with `exec openvpn` and there isn't really anything post that. If you agree I'll just merge this as is?

wilmardo:
> As for the other pre/post scripts transmission has. I'm not sure if they would make sense for openvpn.

Me neither so LGTM!
Hopefully this weekend I will look into the dev/tun device creation opt-in :)

haugene:
Great! Thanks again 👍

PR #703, opened 2019-02-09
Adding ProtonVPN configurations

cat24max:
I added all ProtonVPN servers that allow P2P traffic.
The configs work for me and the image is in "production" on 2 servers without any issue.


cat24max:
Nevermind, me being stupid

PR #704, opened 2019-02-09
Adding ProtonVPN configurations

cat24max:
I added all ProtonVPN servers that allow P2P traffic.
The configs work for me and the image is in "production" on 2 servers without any issue.


haugene:
Looks good! One note is that I see proton is using the "pull" option so that config can be pulled dynamically from the VPN server. This tends to override the --ping and --ping-exit commands. So the pull-filter could be added, also described in the readme. Considering adding it to all providers who has pull. But we'll see.

Have planned to add this for a while, so thanks a lot for helping out 👍  Merging to master now and adding a default.ovpn symlink as I have for all providers (in case no openvpn_config variable is given). Just a FYI for next time.

PR #708, opened 2019-02-14
Added a health check

btimbermont:
Sometimes transmission is unable to connect to the internet because the VPN connection shut down. The container then just sits there not doing anything, without it being noticed. A simple container restart fixes this of course, but the main problem is it going unnoticed.

Therefore I added a docker health check. It simply pings to see if the network is still up. I made a script instead of directly calling ping, because the ping command uses exit code 2 in some circumstances, which you shouldn't do according to Docker documentation. The script simply catches the ping exit code and returns either 1 or 0, to make the exit code up to spec with docker guidelines.
This also makes it easier to change or extend the way the health check is done, if need be.

The default host is google.com, this can be changed in the ENV settings. I added that new ENV variable to the readme.md.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ RUN apt-get update \
     && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* \
     && groupmod -g 1000 users \
     && useradd -u 911 -U -d /config -s /bin/false abc \
-    && usermod -G users abc \
+    && usermod -G users abc
 
 ADD openvpn/ /etc/openvpn/
 ADD transmission/ /etc/transmission/
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,11 +34,14 @@ RUN apt-get update \
     && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* \
     && groupmod -g 1000 users \
     && useradd -u 911 -U -d /config -s /bin/false abc \
-    && usermod -G users abc
+    && usermod -G users abc \
 
 ADD openvpn/ /etc/openvpn/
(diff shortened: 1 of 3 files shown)


h45rd:
While I really like the idea of monitoring / health checking the container (also cf. #680). I believe OpenVPNs [built-in connectivity check](https://openvpn.net/community-resources/reference-manual-for-openvpn-2-4/) as proposed in the [container readme](https://github.com/haugene/docker-transmission-openvpn#restart-container-if-connection-is-lost) to be the more elegant solution for a simple "online test".

Or are there instances I am missing where a simple ping probe would be superior?
Maybe as a second fail-safe in case OpenVPN itself stalls?

btimbermont:
Oh wow, turns out this was a PEBCAK because I totally missed that in the readme. I had this problem, saw there was no heath check and made the change. The existing method is actually exactly what I personally want.

The only reason to use my change over the existing one would be wanting to monitor connection problems. The existing solution simply restarts the container, I imagine not all users want that and want to know a problem exists, rather than the container just restarting, possibly over and over.

Maybe a good solution would be to have both a health check and a separate built in option to restart the container. I'll leave it up to haugene to decide what to do with it.

haugene:
Hey guys. Been thinking about this one and the current value of this script is probably not high as we have the other ping-exit method to actually handle error situations.

But I believe that we can expand these health checks and make them quite useful for debugging errors when more inexperienced docker users creates issues. It could for example curl transmission to see if it's running. Asking for the health check logs could then be useful.

Merging it for now, as it can't hurt. But if we can't find a good use case for it we might remove it later I guess.

haugene:
And thanks for contributing @btimbermont 😄 👍

PR #711, opened 2019-02-16
Updated config files for Frootvpn

lukecdavidson:
Updated the config files for US, Canada and Sweden. The remote domain was old and no longer worked. Also removed the french config file as frootvpn doesn't allow torrenting over the French endpoint making it useless to this project.


openvpn/froot/ca-east.frootvpn.ovpn
```diff
@@ -1,18 +1,22 @@
...
 dev tun
-proto udp
 resolv-retry infinite
-remote-random
 nobind
 persist-key
-persist-tun
-auth-user-pass /config/openvpn-credentials.txt
-ns-cert-type server
+auth-user-pass
```
haugene: This line should not replace `auth-user-pass /config/openvpn-credentials.txt`
The username and password supplied as enviroment variables are persisted to file and needs to be read so that OpenVPN don't prompt for credentials on startup.

haugene:
Great! Thanks for helping to keep the project up to date, and sorry for my late reply.
I added a comment to one of the files, but it goes for the others as well. The auth-user-pass line needs to stay as it is for our authentication mechanism to work. Apart from that it looks good.

Will merge when that gets updated :)

lukecdavidson:
Thanks! I have updated the configs and reverted the auth-user-pass line.

haugene:
Looking good! Thanks again :)

PR #720, opened 2019-02-23
Update ivpn configs

hongtron:
Adds config files for several new locations, and updates some settings in existing ones based on configs [downloaded from IVPN](https://www.ivpn.net/releases/config/ivpn-openvpn-config.zip) on 2/23/19 (e.g. [Canada.ovpn](https://github.com/haugene/docker-transmission-openvpn/pull/720/files#diff-4b104a56477b07c7b3eb4c283d994a41L15n)).

Generated by pulling the configs as linked above, and running `adjustConfigs.sh`. Also preserved the legacy filename format with spaces (the current versions use underscores in place of spaces) for backwards compatibility.

### DIFF
--- a/openvpn/ivpn/Australia.ovpn
+++ b/openvpn/ivpn/Australia.ovpn
@@ -1,7 +1,7 @@
 client
 dev tun
 proto udp 
-remote gb.gw.ivpn.net 2049
+remote au-nsw.gw.ivpn.net 2049
 auth-user-pass /config/openvpn-credentials.txt
 
 resolv-retry infinite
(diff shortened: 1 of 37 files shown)


haugene:
Thanks for helping out keeping the configs up to date 👍 This looks good to me.
Just one comment before I merge. I support the new naming conventions with underscore instead of space, and don't mind breaking backward compatibility on that small things.

So if you want we can change the naming, if not then I'll just merge :)

hongtron:
Nice, I like the new naming too - I'll push the changes in the next day or two.

hongtron:
@haugene should be ready to go now!

haugene:
Great stuff, thanks for contributing! 👍

PR #722, opened 2019-02-26
Create new CREATE_TUN_DEVICE variable

wilmardo:
As stated in #697 an option to enable creation of the /dev/net/tun device inside the container. This should mitigate the need to pass the device from the host to the container.
I create the device as the Linux kernel documentation states [here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/networking/tuntap.txt?id=HEAD#n36).
This also enables the container to be used on Kubernetes, the `--device` parameter isn't available (only block and character devices).

I purposefully haven't written that much documentation because I don't know how to proceed. If this works like it should on NAS devices then most of that documentation can be deleted. If this totally mitigates the purpose of attaching a device the whole readme needs cleaning up.

In short please test this new approach! I build and pushed the latest and latest-alpine to Docker Hub:
https://hub.docker.com/r/lansible/transmission-openvpn/tags

(also removed trailing spaces from the readme, couldn't resist ;))

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,9 +78,9 @@ If you provide a list, a file will be randomly chosen in the list, this is usefu
 ```
 -e "OPENVPN_CONFIG=ipvanish-AT-Vienna-vie-c02,ipvanish-FR-Paris-par-a01,ipvanish-DE-Frankfurt-fra-a01"
 ```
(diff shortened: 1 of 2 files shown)


wilmardo:
I just tested the image myself and it works nicely on my Kubernetes cluster :)

haugene:
Awesome. Really looking forward to testing this. Feels like such a waste to have been running for years with mounted devices ;) Not that it's a big hassle for most linux distros, but the NAS gang will get a simpler life. Thanks again!

PR #726, opened 2019-03-03
Add ability to specify the system architecture for Dockerize in the alpine image

gourleyman:


haugene:
Good addition. Merging, and adding for ubuntu version as well.
Thanks for contributing!

Issue #655, opened 2018-12-04
NordVPN API integration


niXta1:
I'd love to see the NordVPN API integrated like here: https://github.com/azinchen/nordvpn

tomhumphrey:
Would love to see this implemented as well

haugene:
Anyone tested this?

haugene:
This seems to be quite stable now. Thanks again to @Troubadoure for implementing this feature. Closing.

PR #728, opened 2019-03-05
Integrated with NordVPN api to select recommend server based on criteria

Troubadoure:
Included update script on start.sh to ensure latest server configs are being used. Also, created new script titled NordVPN_Server_Selector.sh based on @bubuntux [NordVPN Docker Client](https://github.com/bubuntux/nordvpn) to select the recommended server from NordVPN's api

As requested in issue [655](https://github.com/haugene/docker-transmission-openvpn/issues/655)


haugene:
Great stuff! This is going to be a really nice feature for NordVPN. As you know, their configs change frequently and it's been a real maintenance issue.

A couple of minor comments/questions. First, would you mind squashing your commits? I like keeping a clean history when possible. Second, any reason why you're sourcing the updateConfigs.sh script instead of calling it from openvpn/start.sh? And third, do you think we might as well delete the existing checked in configs and make the script run whenever NordVPN is chosen? The configs will gradually be outdated and this should be the preferred way of getting them?

Last comment. I'm thinking of a better way to handle the environment variables. Unsure if the NORDVPN ones should be in the Dockerfile or if we should try to keep it clean from provider-specific properties. But I'll do something about that later anyways, so no problem.

Troubadoure:
I've squashed the commits. 

updateConfigs.sh is sourced as I was having issues when just calling the script as it would continue before the download/unzip of the configs had completed. There may be a better way to force the wait for this thought I don't know.

I think so, they just get replaced on start anyway - I initially started with the idea that all configs that where publicly available would be automatically downloaded on select of the provider - and any vpns that didn't provide their configs publicly could be added manually by the user instead of them having to be maintained in this container. This way you could have a single folder for any config, which could either be exposed to the host or populated by a provider specific script. However, I ran out of time to do it. My main focus was the NordVPN part and I will probably re-visit other vpns at a later date when I get some time

I agree with this - it should be provider agnostic, however for the time being the script needed to expose some variables and I wanted it to be clear that these where only for NordVPN.

haugene:
Ok, I'll merge it to dev now and see if I can have a closer look within a couple of days.
The container started with only one, and later a few, providers. Now there's so many that maintaining them is getting to be a real headache. So I agree, something needs to change.

There's the "download on startup" alternative, but it could also be downloaded at build time so we don't hammer the providers when containers hit restart-loops.

Anyways. Thanks again for helping out with a great contribution and we'll see where the project ends up 😄

PR #741, opened 2019-03-17
Added sweden ovpn (TCP) version as downloaded from Anonine: https://a…

gitsly:
This is just the raw downloaded Sweden - Stockholm.opvn (TCP version), I don't really know what the contents of such a file is but...

### DIFF
--- a/openvpn/anonine/Sweden.ovpn
+++ b/openvpn/anonine/Sweden.ovpn
@@ -1,19 +1,26 @@
 client
-dev tun0
-proto udp 
+dev tun
+proto tcp
 ; Cert
-ns-cert-type server
-cipher BF-CBC
+remote-cert-tls server
+cipher AES-256-CBC
 ;Host
 resolv-retry infinite
 ;auth
-auth-user-pass /config/openvpn-credentials.txt
+auth-user-pass
+auth-nocache
+keepalive 10 30
 
-remote se-ovpn.anonine.net 443
+remote se-ipsec-private.anonine.net 1195
 persist-key
-persist-tun
+persist-remote-ip
 nobind
+
+sndbuf 0
+rcvbuf 0
+
 comp-lzo
+
 verb 2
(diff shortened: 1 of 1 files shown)


gitsly:
When trying with this one straight off I get (not really expecting it to work but)
transmission_1  | Using OpenVPN provider: ANONINE
transmission_1  | Starting OpenVPN using config Sweden.ovpn
transmission_1  | Setting OPENVPN credentials...
transmission_1  | adding route to local network 192.168.1.0/24 via 172.21.0.1 dev eth0
transmission_1  | Options error: --keepalive conflicts with --ping, --ping-exit, or --ping-restart.  If you use --keepalive, you don't need any of the other --ping directives.
transmission_1  | Use --help for more information.

haugene:
Yeah. Looks like there are changes here that could be the cause of your problems.
Some of your changes would also break it, I'll leave a comment on those lines and then just merge it manually.

openvpn/anonine/Sweden.ovpn
```diff
@@ -1,19 +1,26 @@
...
+proto tcp
 ; Cert
-ns-cert-type server
-cipher BF-CBC
+remote-cert-tls server
+cipher AES-256-CBC
 ;Host
 resolv-retry infinite
 ;auth
-auth-user-pass /config/openvpn-credentials.txt
```
haugene: This needs to be left as it is. It points to the file in the container that has the credentials

openvpn/anonine/Sweden.ovpn
```diff
@@ -1,19 +1,26 @@
...
+auth-user-pass
+auth-nocache
+keepalive 10 30
```
haugene: Keepalive will not work. Or. We're recommending running with ping-exit that makes openvpn exit when there are connection issues. That way docker can restart the container when something hangs or fails.

PR #743, opened 2019-03-18
Anonine update

gitsly:
So I guess this would be a working solution for updated Anonine files. 
I just downloaded the TCP version from Anonine and setup 
auth-user-pass /config/openvpn-credentials.txt
and 
removed keep alive setting (which is provided by image configuration to what I understand)


haugene:
There was some merge conflicts so I'll merge it manually. Also removing the spaces in config names because it can lead to having to quote variables etc. Apart from that it looks great 😄 

Thanks for contributing!

gitsly:
No problems. Tnx for maintaining a nice project.

PR #747, opened 2019-03-23
Surfshark VPN

Fylax:
I added all the config files required for surfshark vpn


haugene:
Looks good to me. I'm just going to remove some of the ping options in the config.
We're recommending to use ping and ping-exit option (https://github.com/haugene/docker-transmission-openvpn#restart-container-if-connection-is-lost), as we rather want openvpn to exit and Docker to restart the container in such cases.

Thanks for contributing to the image 👍

Issue #748, opened 2019-03-23
Mullvad config files need to be replaced


carjed:
**Describe the problem**
Looks like the old Mullvad configuration files will no longer work and need to be replaced, per the blog post on 2019-03-22:
https://www.mullvad.net/en/blog/2019/3/22/suddenly-cant-connect-mullvad/

p-eh:
Where they replaced, or did you find a way around it? In the same boat myself here.

haugene:
I think he closed it again because of this line in the blog-post:
> Any files downloaded after 19 November 2018 are fine

And the configs in this container were updated 7th of January, so they should be fine. Have you pulled the lates image?

carjed:
yep, exactly what @haugene said—pulling the latest image is what fixed it for me.

p-eh:
I assumed it was a certificate issue, but shortly after realized that it was due to no having a password. Added just the letter `m` as the password parameter and it worked. Ran into some ipv6 issues too, but used the following as `extra parameters` in docker and it fixed it all. 

`--sysctl net.ipv6.conf.all.disable_ipv6=0`

haugene:
@p-eh @carjed, just merged #751 which removes the old certificate from the ca.crt.
If one of you could pull the image again and verify it's working properly over the next couple of days that'd be great. Then Mullvad should be stable for a while going forward 😄

carjed:
pulled latest image and everything seems to be working ok for me—thanks!

PR #751, opened 2019-03-24
Update ca.crt for Mullvad

Aweponken:
Mullvad has updated their ca certificate and the old file does not work for me.
Reason for changes can be found here: https://www.mullvad.net/sv/blog/2019/3/22/suddenly-cant-connect-mullvad/


haugene:
See #748, have you tried pulling the latest version of the image?
Just want to confirm that it's working now as well? Need to know which changes we make and why.

Looks like there were two certificates for a while and that they have now finished migrating to the new one. This is the latest cert downloaded from Mullvad? Seems they're removing the old cert so it makes sense.

Aweponken:
I did try the latest image and it didn't work for me. I replaced the cert in the image and then it worked for me. They have completed the shift of certificates now so I guess it's appropriate to remove the old certificate. Yes, the cert is the latest downloaded from Mullvad.

haugene:
Ok, great. Merging it straight to master and re-tagging the 2.7 release I did earler today. Thanks for contributing!

PR #756, opened 2019-03-26
Fetch transmission-web-control from its original repository

bperel:


haugene:
Elegant solution, thanks for the fix 👍

bperel:
I realized afterwards that I forgot to apply the same changes on `Dockerfile.armhf`, `Dockerfile.alpine` and `DockerfileCrossBuild.armhf`. Can you do it or do you prefer to reopen this PR and I'll add the changes soon?

haugene:
If you wouldn't mind, that would be great 👍If not I'll have a look this weekend.

PR #757, opened 2019-03-26
NordVPN On Demand

Troubadoure:
Apologies for two pull requests in one day - After resolving the issue with the default.ovpn symlink, it was bugging me that each time the container ran it would download all of the NordVPN configs. After a bit of research I found a url where I could download just one ovpn file. A bit of rework and now only downloads the applicable config based off of NordVPNs api. No more 3000+ configs.

User can configure this using either NORDVPN_COUNTRY to connect to the best server in that country or a specific config using the OPENVPN_CONFIG parameter. If neither are specified config will be selected based off of NordVPN api and if both are specified it will connect to OPENVPN_CONFIG server first using the api to decide on a backup 'default.ovpn' config in case it fails in the country specified.

This may look a bit daunting with ~10k file changes but it is mainly the deletion of the NordVPN ovpn files. Other than that I've removed the NordVPN environmental variables from the Dockerfile - they can either be set by the user or are now automatically set in the startup script. I've also deleted the NordVPN selector script I introduced as during my rewrite I came to the conclusion there was no need for both the selector script and the updateConfigs.sh script but to keep everything in keeping I kept the updateConfigs.sh script and replaced the contents with my NordVPN_Selector script.

This is a lot faster then downloading all of the ovpn configs making the container snappier to boot


haugene:
Nah, the more the merrier 👍 😄 
Liking the cleaner way of doing it, avoiding unnecessary load of 8k configs and it's a step in the right direction getting rid of all the configs for NordVPN. Want to go this way for all providers. Download on startup and do the sed replacements we need for the container.

For other providers that aren't that dynamic with an API for "best location" we could consider only downloading new configs if they're over 24h old for example. But here it looks like a good solution.

Thanks again for a great day of contributions :)

PR #768, opened 2019-04-02
modifying config to match Torguard config generator

tinefin:
**Process:**
Generated config in Torguard client area web UI. This is the exact result file, only with `auth-user-pass` changed to `auth-user-pass /config/openvpn-credentials.txt`

**Status:**
Tested and works.

**Comments:**
cipher AES-256-CBC no longer supported it seems (I tried that), only AES-128-CBC.

Tried running `./adjustConfigs.sh torguard` but it spits out lines like `sed: 1: "torguard/Austria.ovpn": undefined label 'orguard/Austria.ovpn'` for every single ovpn file.

**TODO:**
I assume the ca cert should be moved. Not sure about the format, leaving for a future kind soul.

**TODO2:**
These warnings are generated when connecting:
```
WARNING: 'link-mtu' is used inconsistently, local='link-mtu 1602', remote='link-mtu 1569'
WARNING: 'tun-mtu' is used inconsistently, local='tun-mtu 1532', remote='tun-mtu 1500'
WARNING: 'comp-lzo' is present in local config but missing in remote config, local='comp-lzo'
```
- Attempted to set both link-mtu and tun-mtu, only one can be specified.
- Attempted to specify `tun-mtu` as in former config, but it doesn't remove the warning. Overridden somewhere else?
- Not sure where comp-lzo comes from.

It works, despite warnings.

### DIFF
--- a/openvpn/torguard/Denmark.ovpn
+++ b/openvpn/torguard/Denmark.ovpn
@@ -1,30 +1,39 @@
 client
 dev tun
-proto tcp
-remote den.torguardvpnaccess.com 995
+proto udp
+remote den.torguardvpnaccess.com 1912
+remote-cert-tls server
+auth SHA256
+key-direction 1
+setenv CLIENT_CERT 0
+<tls-auth>
+-----BEGIN OpenVPN Static key V1-----
+770e8de5fc56e0248cc7b5aab56be80d
+0e19cbf003c1b3ed68efbaf08613c3a1
+a019dac6a4b84f13a6198f73229ffc21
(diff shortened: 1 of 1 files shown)


haugene:
Hey @tinefin and sorry for the lack of response here.

As this just updates one ovpn file, is it a sign that all the others should be updated as well? I tried to run the updateConfigs.sh script that is written for this provider but I didn't end up with the same result.

Are you still active here? If I try to replace all configs, or download both tcp and udp versions could you test and verify it then?

haugene:
All TorGuard configs updated in #848. Closing this for now.

PR #772, opened 2019-04-05
Add option to enable ipv6 in container

ArjenR:
On my host (opensuse Kubic) it was necessary to explicitly enable ipv6 in the container. Otherwise the container startup would fail when the IPv6 address was set on the tun interface.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ $ docker run --cap-add=NET_ADMIN -d \
               --log-driver json-file \
               --log-opt max-size=10m \
               -p 9091:9091 \
+	      --sysctl net.ipv6.conf.all.disable_ipv6=0 \
               haugene/transmission-openvpn
 ```
 


haugene:
Hey. Sorry for the delayed answer here and good that you're pulling attention to some IPv6 troubles.
My only question is if this is host specific as you say - should it be the default option for all users?

Most users copy the example run command and modify it slightly. Will this be needed by the majority? Or should we maybe move it to another section of the README? Thoughts?

ArjenR:
Should it be default? I have little docker experience, so I do not know if this occurs more often. It could be my specific environment where ipv6 is disabled by default for containers because I chose some option in the past or because I use Kubic with Portage? No idea and can't recall. 

But to mention it in the README would be good. My intention was to help others wondering what to do to resolve this.  :-)

haugene:
I'm closing this now and we'll have to reconsider what should be default at a later time.

PR #776, opened 2019-04-09
feat(arm64): arm64 Dockerfiles

megglos:
Firstly thank you for this great image, I'm using it for a quite while now!

I recently switched to a NanoPi M4 running an aarch64 OS on which the armhf image causes the syslog getting flooded (10-20 per second) with deprecation warnings like this:
`kernel: [  987.971062] "transmission-da" (3870) uses deprecated CP15 Barrier instruction at 0xf6cef910`.
This is due to running armhf compiled binaries on aarch64.

I rebuilt the image based on the `balenalib/raspberrypi3-64:stretch` image and my syslog is neat now.
As of now the Dockerfiles are identical except for the base image. The usage of dockerize-armhf binaries also seems fine, at least I noticed no further warnings yet.

Would be great to get this merged for official arm64 images :)

### DIFF
--- a/Dockerfile.arm64
+++ b/Dockerfile.arm64
@@ -1,0 +1,129 @@
+FROM balenalib/raspberrypi3-64:stretch
+MAINTAINER Kristian Haugene
+
+VOLUME /data
+VOLUME /config
+
+# Update packages and install software
+RUN apt-get update \
+    && apt-get -y install transmission-cli transmission-common transmission-daemon \
+    && apt-get install -y dumb-init unzip openvpn curl ufw git tinyproxy jq \
+    && curl -L -o /tmp/release.zip https://github.com/Secretmapper/combustion/archive/release.zip \
+    && unzip /tmp/release.zip -d /opt/transmission-ui/ \
+    && rm /tmp/release.zip \
+    && git clone git://github.com/endor/kettu.git /opt/transmission-ui/kettu \
+    && mkdir /opt/transmission-ui/transmission-web-control \
+    && curl -L https://github.com/ronggang/twc-release/raw/master/src.tar.gz \
+     | tar -C /opt/transmission-ui/transmission-web-control/ -xzv \
+    && ln -s /usr/share/transmission/web/style /opt/transmission-ui/transmission-web-control \
+    && ln -s /usr/share/transmission/web/images /opt/transmission-ui/transmission-web-control \
+    && ln -s /usr/share/transmission/web/javascript /opt/transmission-ui/transmission-web-control \
+    && ln -s /usr/share/transmission/web/index.html /opt/transmission-ui/transmission-web-control/index.original.html \
+    && apt-get purge git unzip \
+    && apt-get autoremove --purge \
+    && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* \
+    && curl -L https://github.com/jwilder/dockerize/releases/download/v0.6.0/dockerize-linux-armhf-v0.6.0.tar.gz \
+     | tar -C /usr/local/bin -xzv \
+    && groupmod -g 1000 users \
+    && useradd -u 911 -U -d /config -s /bin/false abc \
+    && usermod -G users abc
+
+# Add configuration and scripts
+ADD openvpn/ /etc/openvpn/
+ADD transmission/ /etc/transmission/
+ADD tinyproxy /opt/tinyproxy/
+
+ENV OPENVPN_USERNAME=**None** \
+    OPENVPN_PASSWORD=**None** \
+    OPENVPN_PROVIDER=**None** \
+    GLOBAL_APPLY_PERMISSIONS=true \
+    TRANSMISSION_ALT_SPEED_DOWN=50 \
+    TRANSMISSION_ALT_SPEED_ENABLED=false \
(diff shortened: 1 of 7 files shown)


haugene:
Hey! Sorry for the delay here, but you know. Easter and everything.
Thanks for contributing and adding support for arm64. 

I'll have a look at this again in a few days. There's probably no way around. We should support arm64 as well, but it's starting to be a lot of different versions and architectures to be supported.

Just merged some changes to the armhf dockerfile so will just need to copy that into here when merging. And want to sleep on an alternative solution to avoid so much copy pasting ;)

megglos:
Hey @haugene  thanks for getting back on this!

As they are for now identical except for the base image maybe we could just use the same Dockerfile for now and make the base image an build arg with a default value (being armhf) and only configure the build to be executed another time with the 64 image as argument?

Celoxocis:
any update on this? would be nice to have arm64 build and run it on my odroid n2. =)

ndanyluk:
Also interested in any updates to run on my N2 :)

Have you thought about possibly using a tool like Travis CI to automate the builds?  Then you could have a folder for Dockerfiles and make the arch an environment variable in Travis

haugene:
Just can't seem to get enough time these days - but yes. There should be builds for more platforms, and the Docker hub is not ideal for managing it. So a CI tool could be a good way of going forward.

I had an early thought about building in a CI, but there was also something "safe" about having the image marked as an "automated build" on the hub. I figured that building and pushing from somewhere else would not give users the certainty that "what they see is what they get" when it comes to the .ovpn configs that are bundled.

That being said. An open build on a tool like Travis CI is also transparent, and the paranoid can always build it themselves. When it comes to Travis CI specifically I've become a bit sceptic as to it's future due to posts like this:
https://www.reddit.com/r/devops/comments/at3oyq/it_looks_like_ibera_is_gutting_travis_ci_just_a/
https://medium.com/@dirk.avery/r-i-p-travis-ci-347753c73775

But if anyone were to make a CircleCI build and submit it in a PR, that would be LOVELY 😄
Maybe we should create an issue. It could probably supersede a couple of others that target builds for various platforms.

Thoughts?

haugene:
Hello again guys. Just merged my own PR #850 that addresses the arm64 official builds.
Moved the arm builds to CircleCI: https://circleci.com/gh/haugene/docker-transmission-openvpn

Will have to do the same for proxy and rss-plugin, and considering doing all release builds there as well. But for now. @megglos, @Celoxocis, @ndanyluk: Could you test the new :latest-arm64 tag and see if it works? ^^

ndanyluk:
Giving it a look now :) Thanks for doing this!

So far the container appears to be running, but I noticed that the utilities you install for armhf/arm64 are different than the standard ubuntu image (particularly the rar/unrar commands) - I was able to add them to the debian image by adding debian non-free sources to /etc/apt/sources.list but since I was poking around it got me thinking, is there any reason now to use the balenalib images over ubuntu?

Would it be possible to base the arm64 image off the ubuntu image rather than the armhf image?  You would need to use arm64v8/ubuntu:16.04 rather than ubuntu:16.04 in your circleCI builds, but I think it would be almost identical from a Dockerfile perspective otherwise.  

I'm going to give it a shot building it locally and testing, but just figured I'd ask before I go to sleep for the night

haugene:
I'm very positive to any change that would make the versions closer to each other. Tools are probably added from feature requests and PRs and the arm images have just been forgotten or left out.
I don't have an arm-device to test on these days, so I haven't really made any changes to this image as I can't verify the changes.

That being said, the "cross-build" functionality is something that comes from balenalib images. Right? Because we want to build the arm images on x86 hardware we need this functionality.

ndanyluk:
Gotcha, that makes sense.  Actually I don't believe you need the balenalib docker images at all now that you're using CircleCI.  What Balenalib does to enable cross-building is they copy the qemu-user-static binary into their images at /usr/bin/ as the first step in the build.  They make sure to specifically build from the arm64v8 or arm32v7 versions of the ubuntu or debian images and that's their first step.  

I'll try to whip something up later today, but I think it should be pretty straightforward and using the same approach you used to enable arm64 we should be able to use the ubuntu base images and get to a common image structure

haugene:
Sweet! As I said I'm open to pretty much everything - the focus is ease of use and ease of maintenance. Apart from that it's grown to many distributions and hardware architectures without any strict rules and it could be time to clean up a bit.

And while we're at it. Been thinking about skipping the ubuntu image and just do alpine. Thoughts?
Most people don't care, and the ones who do usually prefers alpine because of the footprint.

And FYI. I'm taking a vacation from tomorrow and might be away for a couple of weeks. But then, if there comes a rainy day - vacation time is effective coding time ;)

edgd1er:
Balenalib communicate about the light weight of their images, and as they published many images for many linuxes distribution for many architecture, it's easier to move from x86 debian-build to armv7 debian-run. I think this opportunity is  kind of cool. At least, it should be easier for users to test and suggest modifications.
a Simple sed can transform: balenalib/amd64-debian:build  => balenalib/armv7hf-debian:run

That being said, I use docker-transmission-docker with openmediavault (debian based) ;)

megglos:
@haugene sorry for the late reply, the latest-arm64 images work fine for me, closing this PR as obsolete :) Thx for caring about arm64 images!

PR #777, opened 2019-04-11
Feature add ghostpath VPN

eskemojoe007:
GhostPath is just a resale of SlickVPN.  So I added the necessary files to make it usable here as well.


haugene:
Sorry for the delay, this looks good to me 👍 
Thanks for contributing!

PR #787, opened 2019-04-22
Adding ProtonVPN Country Configs

kyleinprogress:
ProtonVPN recently added country configs that allow you to select a country and they will pick the fastest server. Easier than swapping configs.


jyc66:
Will these configs pick p2p compatible server?

kyleinprogress:
Yes. When you go to download the server configs from ProtonVPN, you have 3 options. One of them is "Country Configs" and both Sweden and Singapore are 100% P2P compatible.

Screen Shot 2019-04-22 at 11 59 05 AM

jyc66:
I see, thanks for the explanation.

haugene:
Looks good 👍 Thanks for contributing!

PR #792, opened 2019-04-25
Updated freevpn openvpn config files

bandrea83:
i have updated freevpn config file that doesn't work anymore (read note on the freevpn website, april update)


bandrea83:
i also added automatic openvpn_password fill reading from the website

haugene:
Hey @bandrea83. Sorry for the late response, but your changes looks good to me 👍
Thanks for helping to keep configs up to date and adding a neat helper function for the passord.

I think we should get provider specific start scripts at some point, but for now adding it to start.sh is the way to go. Merging now, would be great if you could pull the image over the next couple of days to verify that it's working as expected.

I see there's some git pull issue from the hub right now so no new builds are done, but I'll have a look.

PR #799, opened 2019-04-30
Add AU Perth

JacisNonsense:
PIA now has an Australia Perth location


haugene:
Looks good to me 👍 
Thanks for contributing!

PR #810, opened 2019-05-11
Adding protonvpn config - Free plan

emirdora:
There is no available servers for protovpn on the "Free Plan". 

* Adding servers available (+9 JPN, USA and NL)
* Make nl-free-01 default instead of nl-01

Hope that works for you. I managed to have the config working for nl-free-02 using the custom mode for now, but other might like to have that.

Thanks for sharing this repo with us 👍

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.DS_Store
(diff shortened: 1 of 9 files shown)


emirdora:
Actually dropping this. It doesn't work (probably because of P2P not being supported on free plan). I use another VPN now and think this should not be merged as people using paid plan and default config would suddenly have a broken transmission. An it's hard to debug :(.

Closing this.

jyc66:
Yeah, p2p is blocked on the free plan

PR #819, opened 2019-05-29
Create ca.protonvpn.com.udp.ovpn

itlists:


haugene:
Hey. Thanks for contributing to keeping a wide and updated list of configs.

The configs needs to be slightly adjusted to work with this container. I'll merge both your PR's (#818 and #819) anyways and do the change myself. Will tag the commit with the PR numbers so you can see what needs to be done the next time you add a config.

haugene:
Also noticed that you had the `comp-lzo` option, while all other protonvpn configs set this to `comp-lzo no`. This is a deprecated flag and it seems that the latter is the recommended way to move to new compression, ref the manual: https://openvpn.net/community-resources/reference-manual-for-openvpn-2-4/

I therefore changed those as well for conformity with other configs and an assumption that this prepares for the deprecation of this option.

itlists:
Thanks very much! This is my first time using git and submitting changes. I took the ovpn files directly from protonvpn. I'll look at the changes you've made for future. Thanks again

PR #828, opened 2019-06-06
Update README.md

edgd1er:
Integrate a very simple NORDVPN API how to use.

For my example, response with filters[servers_groups][identifier]=p2p is empty, not sure if none is available in albania or the filter identifier is wrong.


edgd1er:
unneeded as the docs moved to gtihub pages

Issue #846, opened 2019-07-04
TorGuard is updating stuff. Probably need replacement configs. 


tralce:
Seems they're updating certificates and whatnot. How can I help?

![2019-07-04-150250_1167x300_escrotum](https://user-images.githubusercontent.com/11371119/60684615-ce852680-9e6c-11e9-928a-60ea50b4c730.png)

alescdb:
Same here, TorGuard not working anymore.
I've tried `updateConfigs.sh` but after updating I got this error :

```
transmission-openvpn    | adding route to local network 192.168.0.0/16 via 172.18.0.1 dev eth0
transmission-openvpn    | Options error: Unrecognized option or missing or extra parameter(s) in /etc/openvpn/torguard/default.ovpn:16: block-outside-dns (2.4.7)
transmission-openvpn    | Use --help for more information.
```

Side note : `updateConfigs.sh` is failing (`set -e`) because the command `dos2unix` is not installed in the container

alescdb:
So for those looking for a quick fix, exec bash inside the container with :
```
docker exec -it transmission-openvpn bash
```
then install dos2unix and an editor (here vim) :
```
root@f6c3061db22d:/# apt update && apt install -y vim dos2unix
```
Edit  : updateConfigs.sh
```
root@f6c3061db22d:/# cd /etc/openvpn/torguard/
root@f6c3061db22d:/# vi updateConfigs.sh
```
After the sed lines add :
```
sed -i "s/^block-outside-dns/#block-outside-dns/" *.ovpn
sed -i "s/^keepalive/#keepalive/" *.ovpn
```

save with `:wq` then run the script :
```
root@f6c3061db22d:/#  ./updateConfigs.sh
```

restart the container.

haugene:
Changes merged, the docker hub builds are sometimes slow but should be able to test and verify it soon @tralce. Thanks @alescdb for providing a fix 🥇 

This probably supersedes #768 by updating all TorGuard configs?

alescdb:
> This probably supersedes #768 by updating all TorGuard configs?

It seems so, since new config files use AES-256-GCM

alescdb:
@haugene I've disable the `keepalive 5 30` option in the config files because I was getting an error something like "ping is not compatible with keepalive", but I've just realized that my `OPENVPN_OPTS` contains `--inactive 3600 --ping 10 --ping-exit 60`, so maybe we should keep the `keepalive` in default configs, what do you think ?

[...]

PR #848, opened 2019-07-05
Update TorGuard updateConfigs.sh and ovpn files

alescdb:
#846 
Update new TorGuard ovpn files, and correct non-working `updateConfigs.sh`


haugene:
Looking good 👍 Thanks for contributing!

PR #850, opened 2019-07-07
Using CircleCI to build arm images instead of Docker hub

haugene:
Follow up of #776

Moving arm builds from Docker hub to CircleCI to have more flexibility in building with parameter and arguments. Using it for now to swap base image for armhf and arm64. This also removes the need for a separate CrossBuild dockerfile as CircleCI can `sed` the Dockerfile before building.

More builds will be moved. Still lacking builds for the proxy and rss-plugin before changes in #776 is fully implemented.


wilmardo:
@haugene Is it an idea to switch to one base image and use a manifest? I could try to implement something like this?

https://github.com/LANsible/docker-mosquitto/blob/master/.drone.yml

PR #859, opened 2019-07-18
Updated ProtonVPN config files

menteb:
I downloaded all the .ovpn config files from ProtonVPN. These include all Secure Core configs, Country configs and Server configs. All UPD configs. Can do the same for TCP if need be.


menteb:
I can update these every 3 to 6 months, depending on newly added servers if that would be helpful.

PR #876, opened 2019-08-12
New ProtonVPN Certificates

menteb:
This merge request includes all the current ProtonMail OpenVPN certificates with the correct link to the user credentials `auth-user-pass /config/openvpn-credentials.txt`.
The certificates included are all for UDP connections from the Secure Core, Country and individual Server configs.


haugene:
Hey @menteb, and thank you for helping out with keeping the configs up to date. I saw your comment in #859 and if you'd be willing to update ProtonVPN periodically then that's great! One less thing to worry about getting outdated 😄 

Just wondering how your process is here. The config looks good. I just thought it's strange that the PR only has additions and no deletions. I would guess that with a lot of new servers there is also some being removed?

How was your process here? You downloaded all the configs from proton and changed the auth-user-pass line. Did you delete all the existing configs before adding the new?

menteb:
Actually, the ones you currently have in the master branch are all still valid and didn't have any changes. As for the process, I downloaded the configs and did a find and replace so I could easily add the path to the config file. Did not have to delete/change anything. In my previous merge request, I hadn't added that path, so hence this new request. I can see whenever stuff changes on ProtonVPN's side and then ask for a merge again (e.g. when they add a new country).

menteb:
I added the new server configs for ProtonVPN servers in Estonia. Files have been altered to read the credentials from `/config/openvpn-credentials.txt`

haugene:
Sorry about the delay. Dropped the ball on this for a while. Trying to pick it up again :)

PR #883, opened 2019-08-22
Wait for transmission to become responsive, updates to PIA config files

nicknlsn:
I'm not sure if this stuff will be useful but I made these changes for my needs. Calls to transmission were timing out causing the container to stop, so I added code to wait for transmission to become responsive before setting the port.


haugene:
Hey, and thanks for submitting an improvement to the project!
I'm sorry for the late reply, but I haven't been able to prioritise this for a while.

I like the thought of testing for transmission to be live before running the port-operations.
Since we already have dockerize installed in container we could maybe use `dockerize -wait ...` for this?

There are also a bunch of .bak files committed. I guess these are the result of a  `sed -i.bak` command and shouldn't be there? If you update this and squash your commits I'll merge it.

nicknlsn:
No worries, I know how hard it can be to keep up with projects like this.

I got rid of the .bak files and squashed the commits. I saw the merge conflict with the new log statement and pulled from upstream to make that go away. I hope that works for you.

haugene:
This looks good, thanks for the quick update.
Working my way through PRs over the week and doing a larger merge to master with a release soon.

Thanks again for contributing! 👍

PR #895, opened 2019-09-12
fix firewall setting

susman:
Here's a fix for firewall setting function.

Some variables have to be interpolated within a loop, before function is called.

### DIFF
--- a/openvpn/start.sh
+++ b/openvpn/start.sh
@@ -110,7 +110,7 @@ TRANSMISSION_CONTROL_OPTS="--script-security 2 --up-delay --up /etc/openvpn/tunn
 
 ## If we use UFW or the LOCAL_NETWORK we need to grab network config info
 if [[ "${ENABLE_UFW,,}" == "true" ]] || [[ -n "${LOCAL_NETWORK-}" ]]; then
-  eval $(/sbin/ip r | awk '{if($5!="tun0"){print "GW="$3"\nINT="$5; exit}}')
+  eval $(/sbin/ip r s default | awk '{print "GW="$3"\nINT="$5; exit}')
   ## IF we use UFW_ALLOW_GW_NET along with ENABLE_UFW we need to know what our netmask CIDR is
   if [[ "${ENABLE_UFW,,}" == "true" ]] && [[ "${UFW_ALLOW_GW_NET,,}" == "true" ]]; then
     eval $(ip r l dev ${INT} | awk '{if($5=="link"){print "GW_CIDR="$1; exit}}')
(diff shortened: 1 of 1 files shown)


haugene:
Hey. Thanks for wanting to contribute to the project!
It's been a while since I've been into the ip tables part of this, can you say a bit more about the changes you're doing? Haven't ran your changes yet - and it might become clear to me then.

It seems as most of the changes are the way functions are called, and then it's the line that sets GW and INT. And when I do a quick test in the container I can't see why this should be right?

```bash
root@06f650be23d2:/# /sbin/ip r l m 0.0.0.0
default via 172.17.0.1 dev eth0

root@06f650be23d2:/# /sbin/ip r
default via 172.17.0.1 dev eth0
172.17.0.0/16 dev eth0  proto kernel  scope link  src 172.17.0.3
```

Why do we want to print the second line? Will this possibly be host dependant?

susman:
Hi @haugene,
Sorry, missed your reply.
I'm not sure what version of iproute2 you're running, in my container `/sbin/ip r l m 0.0.0.0` doesn't return useful output:

```bash
root@26d597eef483:/# cat /etc/issue
Ubuntu 19.04 \n \l

root@26d597eef483:/# apt search iproute
Sorting... Done
Full Text Search... Done
iproute2/now 4.18.0-1ubuntu2.1 arm64 [installed,local]
  networking and traffic control tools

root@26d597eef483:/# /sbin/ip r l m 0.0.0.0
Error: argument "0.0.0.0" is wrong: "metric" value is invalid

root@26d597eef483:/# /sbin/ip r | awk '{if($5!="tun0"){print "GW="$3"\nINT="$5; exit}}'
GW=172.18.0.1
INT=eth0
```
This indeed might be problematic on setups with several interfaces:
```bash
ip r s default | awk '{print "GW="$3"\nINT="$5; exit}'
```
this will always return default gw

I'll update this branch in a sec.

haugene:
Hello again, I'm tinkering with Alpine now and see that I also get an error on this line.
Just to confirm - you were running the alpine image? Does that mean it hasn't worked for a while in that version?...

Anyways. Thinking of moving away from managing ubuntu, alpine, armhf and arm64 versions of the image. One of the first two has to go if this is going to be maintainable. Think I'll move to alpine and your command came in handy 😄

susman:
Hi @haugene ,
I'm running Ubuntu 19.04 image
Probably has to do with newer iproute2 versions

haugene:
Yeah. I guess so. Not going to add those changes until we change for the 19.04 base image here I guess. The change is coming in alpine image now, and it's handy for the day we change the base. Closing for now.

Thanks for the head up!

PR #911, opened 2019-10-08
PrivateVPN server domain name change (October 2019)

gyrovague:
The PrivateVPN servers changed host names from .privatevpn.com to .pvdata.host, see https://privatevpn.com/serverlist/ .

 [ Please use squash commit when merging this pull request (I edited these config files online on github.com directly, so had to commit them individually). ]

### DIFF
--- a/openvpn/privatevpn/default.ovpn
+++ b/openvpn/privatevpn/default.ovpn
@@ -1,4 +1,4 @@
-remote se-sto.privatevpn.com 1194 udp
+remote se-sto.pvdata.host 1194 udp
 nobind
 dev tun
 
--- a/openvpn/privatevpn/london-uk.ovpn
+++ b/openvpn/privatevpn/london-uk.ovpn
@@ -1,4 +1,4 @@
-remote uk-lon2.privatevpn.com 1194 udp
+remote uk-lon2.pvdata.host 1194 udp
 nobind
 dev tun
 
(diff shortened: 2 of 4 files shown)


haugene:
Thanks for contributing! Merging to dev now, bulking up for a larger push to master.
Looking at the server list we're missing most of the configs. Is this something we should try to add? Are all of them supporting p2p?

gyrovague:
> Thanks for contributing! Merging to dev now, bulking up for a larger push to master.
> Looking at the server list we're missing most of the configs. Is this something we should try to add? Are all of them supporting p2p?

I don't know the definitive answer to either of those two questions, was merely trying to fix the ones that had been added by whoever added PrivateVPN support initially. (It worked fine until about two weeks ago.)

It would be nice if there was a clean way to override the servername in the ovpn config file via an env var passed to docker I guess  ( see here: https://privatevpn.com/support/getting-started/linux/linux/openvpn-cli  ), but that's maybe something not generic enough across all the VPNs. I think you mentioned somewhere you were working on a solution to something like this anyway?

haugene:
Yes, I will be looking into something like that. I've had a little break from this project both due to summer and work life demanding attention. Going into autumn I hope to get a new system in place for the config files.

The turn-key-solution is nice I guess, but maintaining config files is a constant endeavour. Time will tell where we end up 😄

PR #921, opened 2019-10-14
Fix FreeVPN config files.

rsalas:
Fix some problem in FreeVPN settings


haugene:
Hey, and sorry about the delay. This change seems reasonable, and looks like a more secure option when it comes to IP leaks on DNS etc. I'm guessing that's the reason you propose the change? Any other reasons or description you can give on this?

I have a feeling that this will be overridden the next time the configs are updated though. We should get a script to download and edit these configs for freevpn. Should this option be set for other providers as well you think?

rsalas:
Yes, I think it's a good idea to make a script that downloads the files and updates them with the best settings.

haugene:
I'm closing this as I think it's superseded by #945. Correct me if I'm wrong 😄

PR #933, opened 2019-10-24
Fix for NordVPN as UDP and TCP CDN changed

koenvanderlinden:
Fixing the NordVPN CND changes. Should fix downloading from wrong URL and getting a error like.
`Options error: Unrecognized option or missing or extra parameter(s) in /etc/openvpn/nordvpn/default.ovpn:1: html (2.4.7)`


haugene:
Thanks! Merging straight to master 👍

Issue #912, opened 2019-10-08
Send transmission logs to docker, not to disk


varac:
Currently, transmission logs to `/data/transmission-home/transmission.log`. It would be much better to send all logs to docker/stdout like openvpn does.

PR #938, opened 2019-10-31
Added option to log transmission to docker logs

pkishino:
this commit adds the option to address #912 which I think can be useful


wilmardo:
@pkishino This looks cool! Thanks for the PR!

One question though, do you have any reason to chose `/proc/1/fd/1` over `/dev/stdout`?

Also I think it is nicer to rename the variable from `DOCKER_LOG` to `STDOUT_LOG` since that is what it does.

pkishino:
Hey, I followed the example here:
https://docs.docker.com/config/containers/logging/
And also read somewhere else there could be some issues writing to stdout,
but can’t recall exactly what and it didn’t quite work for me in this
container. For me directly to proc has been working great since I created
the PR

haugene:
Cool feature! 👍 Thanks for contributing - and sorry about me being so late to the party 😄

PR #945, opened 2019-11-03
Create script freevpn

rsalas:


haugene:
Sorry about the delay and thank you for helping out with keeping the configs up to date 👍 
Great addition with the script. Will hopefully make the maintenance easier going forward 😄

PR #957, opened 2019-11-26
Add update script for usenet server vpn, and update profiles

jaumann:
*.ovpn files were generated by running included update script


haugene:
Hey @jaumann. Thanks for helping out and keeping configs up to date 👍 
Great addition with another update script. Always extra fun merging them 😄

PR #999, opened 2020-01-05
Add Current Ivacy Servers

Sean89us:
Signed-off-by: Sean Milbrath <Sean89us@gmail.com>

These are the OpenVPN servers currently listed on the [Ivacy servers page](https://support.ivacy.com/vpn-servers/). I left the old ones in place so I don't accidentally break someone's current setup, but not all of the current ovpn files work, so I wanted to add all of the new ones.

### DIFF
--- a/openvpn/ivacy/AustraliaTCP.ovpn
+++ b/openvpn/ivacy/AustraliaTCP.ovpn
@@ -1,0 +1,19 @@
+client
+dev tun
+proto tcp
+remote vlus-au1-ovpn-tcp.ivacy.net 80
+persist-key
+persist-tun
+ca /etc/openvpn/ivacy/ca.crt
+tls-auth /etc/openvpn/ivacy/Wdc.key 1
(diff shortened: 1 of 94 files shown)


Sean89us:
@haugene please review when you can, thanks!

haugene:
Hey @Sean89us, and sorry about that. It's been busy times and I've fallen behind.
This looks great. Thanks for helping out! 👍 

I'm a supporter of letting things fail fast. Meaning that I'd rather have the container say "that config file isn't here anymore" than it saying "resolv error: hostname not found" etc when a server is just gone. So if you want to update it and remove the old servers I'm all for it.

Another possible quick fix. Can it be that the `route 0.0.0.0 0.0.0.0` is causing the container to log `ERROR: Linux route add command failed: external program exited with error status: 2` on startup? I think we can remove it.

Merging to dev now and to master soon. Thanks again 😄

Issue #105, opened 2016-07-25
Please Add VPN Provider ExpressVPN


iAMarash:
Greetings. Appreciate if someone could add support for VPN provider ExpressVPN.

Thanks
Arash

benjamintillman:
Like AirVPN (see README.md), ExpressVPN generates .ovpn files per user. I've been trying to get it to work on and off, but I haven't had any luck yet.

For reference, I've attached a stripped .ovpn file that someone might be able to use as a reference to get things working.

[my_expressvpn_stripped.txt](https://github.com/haugene/docker-transmission-openvpn/files/392209/my_expressvpn_stripped.txt)

haugene:
Hmm, I think it would be quite messy to start passing keys and certs as parameters to the container.
As with #60, I recommend that you download your .ovpn file and use it as a custom provider. This is described in the [readme](https://github.com/haugene/docker-transmission-openvpn#using-a-custom-provider).

Let me know how it goes :)

Buiz:
Tried as you suggested but I get "Could not find OpenVPN provider: CUSTOM"?

haugene:
Then you haven't mounted the vpn config files to the right path inside the container. What does your docker run command look like?

Buiz:
ok. as I create the container (on QNAP's Container Station) I can only run
command 'dumb-init /etc/openvpn/start.sh' - I am not sure how or where to
add your suggested '-v
/usr/my_expressvpn_usa_-_new_york_udp.ovpn:/etc/openvpn/custom/default.ovpn'.
I tried just replacing the default.ovpn file after the container has been
created, but that didn't work either. Any advice?

2016-09-04 9:25 GMT-04:00 Kristian Haugene <notifications@github.com>:

haugene:
You should be able to configure volume mounting in the qnap gui. Maybe under advanced settings or something. That is really the simplest solution.

If not, you could build the container yourself, putting your config inside. Copying it into an existing container should also work if you restart the container afterwards. But your change will be gone when a new container is created. You could commit the change in the container and that way create and tag a new image.

Lots of choices here, but mounting the config is the simplest and most upgrade friendly approach.

Buiz:
I can only mount folders in QNAP's UI. So instead I cloned your repo on
github and added expressvpn as an option using the same file structure as
the other VPNs (folder name expressvpn, then a default.ovpn file that
contains the filename of the expressvpn.ovpn file). When I run it I get
following error:

[...]

PR #1002, opened 2020-01-13
Express VPN: all servers config files added

damsvann:
All ExpressVPN config files added and modified to fit the startup script. Tested and working as expected


vijayanandnandam:
Can this be merged?

haugene:
Yes, sorry about the delay. Always get too little time for this project these days.
Merging now, will be available on latest tag soon. Please update if things work or not 😄 

A bit surprised to see ExpressVPN added. We've had several threads in the isses where I thought that ExpressVPN as with AirVPN uses personal keys keys, ref #105 ? Have they changed?

PR #1026, opened 2020-01-26
Fix 2.92 memory issue in armhf image

fixl:
# Description

The way transmission is packaged in stretch leads to memory issues:
- https://github.com/transmission/transmission/issues/313#issuecomment-446005020
- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=865624

The way this manifests for me is that my raspberry pi keeps running out of memory and becomes unresponsive.

The was fixed in buster/transmission-2.94

I've built and pushed this to [my dockerhub repo](https://hub.docker.com/r/fixl/transmission-openvpn/tags) for testing.

### DIFF
--- a/Dockerfile.armhf
+++ b/Dockerfile.armhf
@@ -1,4 +1,4 @@
-ARG base_image=balenalib/raspberry-pi:stretch
+ARG base_image=balenalib/raspberry-pi:buster
 FROM $base_image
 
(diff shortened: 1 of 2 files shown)


haugene:
Thank you @fixl 👍 Merging to dev now and to master very soon.

PR #1009, opened 2020-01-17
Added HEALTHCHECK on the Dockerfile.armhf

sstassin:
I am using it since fews days, and it is working without any issue (see the screenshot, with one of my container is unhealthy).

It could also be interesting to explain that, according to the Docker Team's 
-  there is no plan to manage a restart, or stop of an unhealthy container.
- an unhealthy docker just run until it it stopped by a command
- this could be"easily done" by a bash command, that invoke a docker ps command to list all unhealthy containers and restart it (i could give you the code if you want)... scheduled in a cron job, you have a 24H/24H running container.

Regards and thank you for you great job.

Image 8

PR #1037, opened 2020-02-04
Add health check to armhf and alpine image

fixl:
# Description

Add `HEALTHCECK`s to armhf and alpine images as well.

This seems like this never made it from [this commit](https://github.com/haugene/docker-transmission-openvpn/commit/0abe1912f4bd41c25ad5b28e04f671b5c6f9238a) to the other two images.

This is similar to #1009


haugene:
Yep, you're right. Should have been added to the others as well.
Thanks for providing a fix and I'm sorry about the delay. Merging now.

PR #1046, opened 2020-02-12
Update default.ovpn

Cabur:
Choosing default for FASTESTVPN fails due to incorrect file name in file
Netherlands-UDP.openvpn should be Netherlands-UDP.ovpn


haugene:
You are correct 😄 Thanks for helping out!

PR #1056, opened 2020-02-23
fix: Quick, dirty hotfix to make NordVPN start.

Pete-Hamlin:
## Overview

So it looks like NordVPN have gone and changed all their API endpoints again (yay), so half the connection stuff flat out doesn't work any more (they even changed the download endpoint). Have managed to get the container up and running locally with the below changes.

This PR updates the endpoint for the download so that as long as the user provides an OPENVPN_CONFIG, then at least the container can grab that particular config and start up correctly. Current behaviour is that update script will curl the 404 page and then immediately fail (unsurprisingly) when it tries to use it as a ovpn config.

After doing some digging (https://github.com/gjedeer/nordvpn/issues/1), I discovered the https://nordvpn.com/api/server endpoint, but it looks to be very different to what we had before, and will require a bit more work to get up and running. I'll hopefully be able to take a look at it at a later date.

### API Status

As it stands for API endpoints (Used to select optimal profile):
https://api.nordvpn.com/v1/technologies - Still there :tada: 
https://api.nordvpn.com/v1/servers/groups - GONE :warning: 
https://api.nordvpn.com/v1/servers/countries - GONE :warning: 

### Footnote

Apologies, this is my first PR on this project. I did look for a contributing guide, but couldn't find one. If I've gone about this all the wrong way, please let me know I'll fix it. Just be great to have this awesome tool up and running again :smiley:

### DIFF
--- a/docs/nordvpn-script.md
+++ b/docs/nordvpn-script.md
@@ -1,5 +1,7 @@
 ## NORDVPN API
 
+UPDATE: 23/02/2020 - NordVPN have updated their API again, breaking much of the below. Currently the only way to get this part of things working again is to directly set your OPENVPN_CONFIG variable with the exact NordVPN config you wish to use (e.g. OPENVPN_CONFIG=nl127.nordvpn.com.udp1194.ovpn). This should hopefully be fixed once the niceties of https://nordvpn.com/api/server can be unravelled.
+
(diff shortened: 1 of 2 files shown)


pkishino:
Looks good

Pete-Hamlin:
So it appears those endpoints are working once again... 

So this PR may not be necessary at this point.

Issue #1061, opened 2020-02-28
PrivateVPN - update opvn files to include port forwarded servers


chris-deluca:
**Which provider?**
PrivateVPN

**Where are the configs?**
[PrivateVPN.zip](https://github.com/haugene/docker-transmission-openvpn/files/4268346/PrivateVPN.zip)

The attached zip file contains servers that support port forwarding.

PR #1064, opened 2020-03-02
#1061 - Add Port Forward ovpn files

chris-deluca:
1. These new files were provided to me by PrivateVPN support.  All ports are forwarded to the VPN connection - where most other ovpn connections do not allow this.
2. I followed the naming pattern established by gyrovague in the original check-in.  The original names of the files as delivered by support were: PrivateVPN-US-New York 4-TUN-1194.ovpn, PrivateVPN-SE-Stockholm-TUN-1194.ovpn, PrivateVPN-US-Los Angeles-TUN-1194.ovpn.


haugene:
Thanks a lot for contributing, and sorry about the delay on my part. I'm merging to dev now, merged another feature there earlier today as well. I'll wait a couple of days for a confirmation that it works or a warning that something is broken - and then merge to master.

Should be available on the dev tag in a couple of minutes...

PR #1066, opened 2020-03-03
Deprecate usenetserver for privado

jaumann:
Usenetserver has switched from Whitelabel VPN as their provider to Privado (non-whitelabled). This PR removes the usenetserver profiles, as usenerserver credentials no longer work with those servers, and adds a script to update the Privado profiles


haugene:
Great, thanks! 👍

PR #1071, opened 2020-03-11
Rename ca.crt.crt to ca.crt

mujambajecha:


haugene:
Looks like the purevpn configs are not referencing this file as they're using it inline. But I agree, it should be .crt 😄

PR #1072, opened 2020-03-12
Updating Proxpn files

bradsk88:
The current files are out of date and ProXPN has changed several of
their servers' IP addresses.

https://www.proxpn.com/updater/locations.html


haugene:
Thanks for helping out keeping configs up to date 👍 
Looks good to me.

kingkool68:
@bradsk88 Thanks so much for updating these. I rely on ProXPN and it took a lot of guessing to find a configuration that worked.

bradsk88:
:beers:

PR #1074, opened 2020-03-14
Updated the Surfshark Config to the Latest

mehulboricha:
Updated the Surfshark Config to the Latest

It can be downloaded from https://account.surfshark.com/api/v1/server/configurations 

Also added a script to rename all the Surfshark configs and remove 3 common lines in all the configs


haugene:
Looks good! Thanks for contributing @mehulboricha 
I've also seen that you've been active and helping out on issues, much appreciated!

mehulboricha:
@haugene I am glad that I was able to help. Also, if you need any assistance then do let me know, I am not an expert but I will try my best to get it done.

Thank you for creating such a useful repo. :)

Issue #1020, opened 2020-01-22
NordVPN Script not downloading ovpn file


timjlowe:
**Describe the problem**
The container is unable to connect to NordVPN as it fails whilst trying to download the ovpn file from NordVPN.  I have tried as many differernt NORDVPN config values as I can think of.

I tried running the updateConfigs.sh script outside of the container and it seemed to work, so it could just be that my NORDVPN env variables are not being passed (or formatted correctly) when run from docker-compose.

**Add your docker run command**
version: "2"
services:
  transmission-vpn:
    container_name: transmission-vpn
    image: haugene/transmission-openvpn
    cap_add:
      - NET_ADMIN
    devices:
      - /dev/net/tun
    restart: always
    ports:
      - "9091:9091"
      - "8888:8888"
    dns:
      - 45.90.28.29
    volumes:
      - /etc/localtime:/etc/localtime:ro
      - ${USERDIR}/docker/transmission-vpn:/data
      - ${USERDIR}/docker/shared:/shared
      - ${USERDIR}/Downloads:/data/watch
      - ${USERDIR}/Downloads/completed:/data/completed
      - ${USERDIR}/Downloads/incomplete:/data/incomplete
    environment:
      - OPENVPN_PROVIDER=NORDVPN
      - OPENVPN_USERNAME=***
      - OPENVPN_PASSWORD=***
      - OPENVPN_OPTS=--inactive 3600 --ping 10 --ping-exit 60
      - NORDVPN_COUNTRY=gb
      - NORDVPN_PROTOCOL=tcp
      - NORDVPN_CATEGORY=p2p
      - LOCAL_NETWORK=10.0.0.0/24
      - PUID=${PUID}
      - PGID=${PGID}
      - TZ=${TZ}
      - TRANSMISSION_RPC_AUTHENTICATION_REQUIRED=true
      - TRANSMISSION_RPC_HOST_WHITELIST="127.0.0.1,10.0.0.*.*"
      - TRANSMISSION_RPC_PASSWORD=webui_password
      - TRANSMISSION_RPC_USERNAME=webui_username
      - TRANSMISSION_UMASK=002
      - TRANSMISSION_RATIO_LIMIT=1.00
      - TRANSMISSION_RATIO_LIMIT_ENABLED=true

**Logs**
Using OpenVPN provider: NORDVPN
2020-01-22 15:12:32 Checking curl installation
2020-01-22 15:12:32 Removing existing configs
[...]

PR #1079, opened 2020-03-18
Nordvpn syntax fix

jsloan117:
Throwing this together real quick. I use VSCode & multiple linters when I write/create scripts/docker containers etc. Which led to this possible fix for #1020.


haugene:
Good catch, looks like two good modifications to me. Thanks for helping out and 🤞 that this fixes the issue. I see that the dev branch have fallen behind master and you branched from master. So I'll change the PR for master and merge it there. Then merge it all back to dev.

haugene:
And then I forgot 🤦‍♂ Ah well. Then it's on dev and can be tested with the dev-tag.
Probably more correct anyways....

PR #1085, opened 2020-03-22
Add docker compose example to README.md

dskvr:
`docker-compose` is standard, and following `linuxserver`'s excellent example, including it in the `README.md` enables a far more efficient quick start from Docker Hub.


haugene:
I agree, should be on the main page. Thanks for contributing!

PR #1093, opened 2020-03-26
Prettify NordVPN docs page

7596ff:
I was looking through the docs for these environment variables and glazed right past them due to them not being in a markdown table. So I made that table. I am unable to comment or format the rest of the docs because I am not aware of the implementation.

### DIFF
--- a/docs/nordvpn-script.md
+++ b/docs/nordvpn-script.md
@@ -1,16 +1,17 @@
 ## NORDVPN API
 
-The update script is based on the  NordVpn API. The API sends back the best recommended config file based on the filters given.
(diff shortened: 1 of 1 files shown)


haugene:
Had a look now and I agree they're easy to miss 👍 Thank you for pointing it out and even providing a fix!

As for the rest of the docs, it's written in plain markdown but using mkdocs to structure and build it to an html site.

If you have python/pip installed you should be able to run `pip3 install mkdocs mkdocs-material` and then from the project root `mkdocs serve`. The built site should then be available on `http://localhost:8000/` and be reloaded automatically on every change. The mkdocs.yml file is the core config and the rest is in the .md files.

PR #1104, opened 2020-03-30
Auto update freevpn

gpongelli:
- Using "me" as default FreeVPN domain
- launched updateFreeVPN.sh during start.sh
- added  updateFreeVPN.sh to a daily cron
- changed updateFreeVPN.sh to use OPENVPN_CONFIG domain
- fix issue on unzip into non existent folder
- searching and changing oven file names accordingly to latest bundle (April 2020) obtained on .me domain

Thank you for this awesome docker image.

### DIFF
--- a/openvpn/start.sh
+++ b/openvpn/start.sh
@@ -56,7 +56,16 @@ then
 elif [[ "${OPENVPN_PROVIDER^^}" = "FREEVPN" ]]
 then
     FREEVPN_DOMAIN=${OPENVPN_CONFIG%%-*}
-    export OPENVPN_PASSWORD=$(curl -s https://freevpn.${FREEVPN_DOMAIN:-"be"}/accounts/ | grep Password |  sed s/"^.*Password\:.... "/""/g | sed s/"<.*"/""/g)
+    export OPENVPN_PASSWORD=$(curl -s https://freevpn.${FREEVPN_DOMAIN:-"me"}/accounts/ | grep Password |  sed s/"^.*Password\:.... "/""/g | sed s/"<.*"/""/g)
+    
+    # Update FreeVPN certs
+    /etc/openvpn/updateFreeVPN.sh
+    
+    # Add daily update to cron
+    crontab -l > tempcron
+    echo "0 1 * * * /etc/openvpn/updateFreeVPN.sh" > tempcron
+    crontab tempcron
+    rm tempcron
 elif [[ "${OPENVPN_PROVIDER^^}" = "VPNBOOK" ]]
 then
     pwd_url=$(curl -s "https://www.vpnbook.com/freevpn" | grep -m2 "Password:" | tail -n1 | cut -d \" -f2)
--- a/openvpn/start.sh
+++ b/openvpn/start.sh
@@ -60,16 +60,6 @@ then
     
     # Update FreeVPN certs
     /etc/openvpn/updateFreeVPN.sh
-    
-    # Add daily update to cron
-    crontab -l > tempcron
-    
-    grep updateFreeVPN tempcron > /dev/null 2>&1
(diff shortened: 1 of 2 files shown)


openvpn/start.sh
```diff
@@ -56,7 +56,16 @@ then
...
+    export OPENVPN_PASSWORD=$(curl -s https://freevpn.${FREEVPN_DOMAIN:-"me"}/accounts/ | grep Password |  sed s/"^.*Password\:.... "/""/g | sed s/"<.*"/""/g)
+    
+    # Update FreeVPN certs
+    /etc/openvpn/updateFreeVPN.sh
+    
+    # Add daily update to cron
+    crontab -l > tempcron
+    echo "0 1 * * * /etc/openvpn/updateFreeVPN.sh" > tempcron
+    crontab tempcron
+    rm tempcron
```
haugene: The configs are set on container/OpenVPN startup, and this script is run on each start of the container.
Am I missing something, or will the cron job not make any difference here? 🤔

haugene:
Looks like a nice change that will make FreeVPN easier to use and keep up to date. Thanks for contributing 😄 

I just have one comment on the cron stuff as I don't see when that will have an effect. A new config is downloaded on container startup and OpenVPN uses that file to set up and connect. Will it then be evaluated again before container restart?

openvpn/start.sh
gpongelli: Hi, the cron is used to automatically call the update script every day at 1:00 AM.
So, this will catch the bundle changes when the container is running for many days.
I'm not sure about container persistency of crontab file, I can be wrong.
Eventually, a check can be added to avoid adding many times the row in crontab.

haugene: Yeah. But let's say that the container runs for a couple of weeks and then there's a change in the configs. Your script will fetch it and download the file and then what? It won't be reloaded while running, right?

gpongelli: Yes, every time the updateFreeVPN.sh script is executed the downloaded bundle files will be put under /etc/openvpn/freevpn , even if the downloaded file are the same of those already present into the container.
There is no reload, actually I've no idea on how to do it.

haugene: I don't think it's possible. I think OpenVPN must be restarted.
But that will happen automatically anyways. If you run with the --ping-exit options then the old server will not respond anymore (at some point) and the ping will fail and container will exit.
Then with the Docker restart flag (either "always" or "unless-stopped") the container will be restarted and then the start.sh will be triggered again and the bundle will be downloaded then.

So I don't think we need to add it to crontab. I don't think it will help.

gpongelli: ah ok, it can be the right way, I didn't know the ping-exit option.

meanwhile I've changed it to add the rule only one time, I'll remove all crontab soon.

gpongelli: Removed the crontab changes.

haugene:
Looks good to me 👍 Thanks for updating.
I'll merge it to dev now. If you wouldn't mind pulling and running the :dev image to see that everything works as expected that would be great. For this provider I could probably do it myself as well I guess? Don't need an account here?

gpongelli:
Yes, I've started trying FreeVPN yesterday and there's no need of any account.
Before I forget, actually the script will look at the ".me" domain, using the "Server 2" password (because it's the script's last match on the retrieved html page).

BTW, I've built and tested my branch locally when I've opened this PR, it worked.

Thank you.

gpongelli:
@haugene I've tested the dev image, the update script is not working.
The problem is because the script is actually called before the interface is up, so zip file is not downloaded.
Maybe it's better to move that call elsewhere, like at the end of start.sh or in "tunnelDown.sh", what do you think ?

PR #1113, opened 2020-03-31
FreeVPN Site and Certificate Changes

mattchis:
FreeVPN has changed its site and certificates.  

This update will correct the issue for when the container reaches out to FreeVPN's site to pull passwords it appears blank in the /config/openvpn-credentials.txt

FreeVPN has removed all of the previous OpenVPN servers and certificates.  I uploaded the new certificates for the new servers


haugene:
Will this be needed after #1104 and #1116 ?

pkishino:
@mattchis Please comment on haugenes reply as I also believe this is not needed any longer

PR #1116, opened 2020-04-01
Fix freevpn script

gpongelli:
Fix network issue for FreeVPN:

- added DNS environment variable; I've tried the "--dns" command line argument and in docker-compose file but they both don't work, the provider server remains unreachable. The only way to reach the server is to replace the content into resolve.conf

- get the password directly from webpage while it's reachable

Tested with this docker-compose file on my raspberry
```
version: '3.3'
services:
    transmission-openvpn:
        volumes:
            - '/storage/path/:/data'
            - '/etc/localtime:/etc/localtime:ro'
        environment:
            - CREATE_TUN_DEVICE=true
            - OPENVPN_PROVIDER=FREEVPN
            - OPENVPN_CONFIG=me-UDP40000    << it is the ovpn file name
            - OPENVPN_USERNAME=freevpn.me
            - DNS=<dns address>              << it works also with locally installed PiHole
            - WEBPROXY_ENABLED=false
            - LOCAL_NETWORK=192.168.0.0/16
            - OPENVPN_OPTS=--inactive 36000 --ping 100 --ping-exit 600 --tun-mtu 1500
        cap_add:
            - NET_ADMIN
        logging:
            driver: json-file
            options:
                max-size: 10m
        ports:
            - '9091:9091'
        image: haugene/transmission-openvpn:latest-armhf
        devices:
            - "/dev/net/tun"
        restart: always
``` 

Thanks.


haugene:
Appreciate the continued effort and contributions here 👏 👍 
I'm not sure about the global DNS environment variable though. I know this has been an issue for some users and I've hinted that they can mount a /etc/resolv.conf file if all else fails.
I want this to be Dockers responsibility and not tamper with the network setup that is configured on the container.

Either way the DNS environment variable will show up in all the GUI configuration tools on the NAS setups and people will expect it to set the DNS. Not only for FreeVPN.

Could you for now change the line
```
echo "nameserver ${DNS}" > /etc/resolv.conf
```
to
```
echo "nameserver 8.8.8.8" > /etc/resolv.conf
```
for now and then we can have another look at this in a separate issue/PR?

If relying on Docker to fix DNS isn't going to be good enough, then I guess we could have a "DNS_OVERRIDE" environment variable that echoes right into /etc/resolv.conf like here. But it should support multiple DNS servers and be executed in the main start.sh script.

gpongelli:
> If relying on Docker to fix DNS isn't going to be good enough, then I guess we could have a "DNS_OVERRIDE" environment variable that echoes right into /etc/resolv.conf like here. But it should support multiple DNS servers and be executed in the main start.sh script.

The point is that, relying on Docker to reach web server BEFORE opening the tunnel, it fails because the resolve.conf is this
```
nameserver 127.0.0.11
options ndots:0
```
This content let fail the initial download, so it doesn't work.



> Either way the DNS environment variable will show up in all the GUI configuration tools on the NAS setups and people will expect it to set the DNS. Not only for FreeVPN.
> 
> Could you for now change the line
> 
> ```
> echo "nameserver ${DNS}" > /etc/resolv.conf
> ```
> 
> to
> 
> ```
> echo "nameserver 8.8.8.8" > /etc/resolv.conf
> ```
> 
> for now and then we can have another look at this in a separate issue/PR?
> 

It's ok.

Issue #1117, opened 2020-04-03
DOCKER_LOG fails if transmission isn't running as root


t0xicCode:
DOCKER_LOG fails when transmission is not running as root because it tries to open `/proc/1/fd/1`, which is chown'd to `root`.

Essentially, dumb-init runs as root, which makes `/proc/1/fd/1` owned by root. If transmission is ran as any other user, it is unable to open that file, as it lacks the permissions.

```sh
docker run --cap-add=NET_ADMIN -e DOCKER_LOG=true -e PGID=1234 -e PUID=1234 --name "log-fail-demo" haugene/transmission-openvpn:latest
```

Relevant log section:
```
Transmission startup script complete.
Fri Apr  3 15:27:23 2020 Initialization Sequence Completed
Couldn't (re)open log file "/proc/1/fd/1": Permission denied
```

Running on an amd64 NAS.

pkishino:
Hi, thanks for the feedback, I’ll try and look into this when I have time. Should be possible to change so it’s usable by non-root

wilmardo:
@pkishino Sorry I didn't mean to snatch your contribution but I ran into this myself yesterday and fixed it ;)

pkishino:
@wilmardo  hey, thanks, looks good :) great work. I tried using stdout originally but had some issues, great if this works :)

wilmardo:
@pkishino If you would like to test, I build and pushed the alpine for testing:
https://hub.docker.com/r/lansibledump/transmission/tags

So just use `lansibledump/transmission:latest` instead of `haugene/transmission-openvpn`

pkishino:
This is working now I believe, please check and close

pkishino:
@t0xicCode Please close this as this was fixed in #1122

t0xicCode:
I can confirm that it was fixed. Thanks y’all

PR #1122, opened 2020-04-05
Fix logging to stdout when using a RUN_AS user

wilmardo:
### Before
```
Sun Apr  5 20:54:10 2020 Initialization Sequence Completed
Couldn't (re)open log file "/dev/stdout": Permission denied
```

### After

```
Sun Apr  5 20:56:18 2020 Initialization Sequence Completed
[2020-04-05 20:56:19.725] Transmission 2.94 (d8e60ee44f) started (session.c:740)
[2020-04-05 20:56:19.725] RPC Server Adding address to whitelist: 127.0.0.1 (rpc-server.c:971)
```

The `DOCKER_LOG` parameter wasn't working with a `PUID`/`GUID` since `/proc/1/fd/1` is owned by root:root.

I tried chowning that path but it wouldn't persist since it isn't designed for that. `/dev/stdout` is the userspace equivalent and is chownable :)

Couldn't resist and fixed the indenting in the file. Sorry for the noise, for just the fix check 8b70873

### DIFF
--- a/transmission/start.sh
+++ b/transmission/start.sh
@@ -61,7 +61,7 @@ if [[ "true" = "$DROP_DEFAULT_ROUTE" ]]; then
 fi
 
 if [[ "true" = "$DOCKER_LOG" ]]; then
-  LOGFILE=/proc/1/fd/1
+  LOGFILE=/dev/stdout
 else
   LOGFILE=${TRANSMISSION_HOME}/transmission.log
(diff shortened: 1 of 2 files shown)


wilmardo:
Fixes: #1117

haugene:
Great stuff, thanks guys 😄

PR #1124, opened 2020-04-06
Updated IVPN configs

vrntnn:
Updated IVPN configs to their latest versions:
https://www.ivpn.net/releases/config/ivpn-openvpn-config.zip


haugene:
Looks good 👍 
Thanks for helping out keeping configs up to date!

Issue #1130, opened 2020-04-10
Transmission settings are not persistent across container restarts


MeanderingCode:
**Describe the problem**
Transmission settings changed at runtime (or written into `settings.json`) are overwritten by environment variables at container start time. The environment variables are baked into the Dockerfile, even if not set at start time (by e.g. `docker-compose.yaml`).

This is because:
* The Dockerfiles have `TRANSMISSION_*` environment variables baked into the images
* The `openvpn/start.sh` script calls `dockerize -template /etc/transmission/environment-variables.tmpl:/etc/transmission/environment-variables.sh`, which writes a script to `export` all these envvars ~(redundant, since they are already set by the Docker engine in the container)~
* The `transmission/start.sh` script sources those variables ~(redundantly)~ and then overwrites `settings.json` with `dockerize -template /etc/transmission/settings.tmpl:${TRANSMISSION_HOME}/settings.json`

I understand there is some level of "feature, not a bug" involved in this, but I believe it to be a shortcoming. I have started a PR to address it, and will put it up as WIP for feedback while I test it.

**Add your docker run command**
Any

**Logs**
n/a

**Host system:**
Any

EDIT: Openvpn calls the script to start transmission without any envvars except its own, hence loading them from `environment-variables.sh`

antifarben:
I probably misunderstand this but when you configure it, [like documented via env variables](https://haugene.github.io/docker-transmission-openvpn/arguments/), everything is fine, right?

pkishino:
> I probably misunderstand this but when you configure it, [like documented via env variables](https://haugene.github.io/docker-transmission-openvpn/arguments/), everything is fine, right?

The point here is that when you change settings in the GUI, currently, after a restart, they are reset back to what it was.. this PR aims to allow for persistent settings across restarts.. so if you change something in the GUI, it would update settings.json and be read at next startup, I believe

MeanderingCode:
> this PR aims to allow for persistent settings across restarts.. so if you change something in the GUI, it would update settings.json and be read at next startup, I believe

That is correct. There is an issue in the WIP PR I haven't had time to address, but that is the goal.

pkishino:
Is the latest code committed? I’d be interested to take a look :)

MeanderingCode:
Since you asked, I put some time in :)

[...]

PR #1131, opened 2020-04-10
[WIP] Persistent Transmission settings

MeanderingCode:
Create Transmission defaults system instead of forcing environment variable use.

Remove baked in envvars for Transmission and put defaults in settings.tmpl.
This allows for persistent settings across container restarts and rebuilds
without forcing environment variables to be updated outside the container (e.g.
in docker-compose.yaml). Settings files can be portable, and preference changes
made at runtime will persist.

Some variables are used elsewhere, by openvpn/start.sh and the transmission
port update scripts for a couple VPN providers. The newly hard-coded
transmission/environment-variables.sh sources these in an order-of-priority
fashion: environment variable, settings file, and default.

Further, writing Docker-set environment variables to a script to source in other
processes is not necessary, as all procs spawned in the container have them set
already.

Addresses #1130


MeanderingCode:
Just an update: It turns out the sourcing of envvars in a shell script written out via dockerize is not redundant: openvpn explicitly executes up/down scripts in a way that carries no environment variables over.

I am going to update this with a new `transmission/environment-variables.tmpl` that will preserve environment variables but also load defaults and settings file values for when envvars are not set (a la `transmission/environment-variables.sh` from https://github.com/haugene/docker-transmission-openvpn/pull/1131/commits/fb81c1f17f5671ce35dbd7ee4f449bf3c22d143d#diff-5a82435864cdea1e4df45fe463abc674 )

MeanderingCode:
Latest commits re-introduce `transmission/environment-variables.tmpl` based on new defaults-settings-envvars work in this branch.

**This now works to build a settings.json from defaults, load a settings.json that is there, and always respect `TRANSMISSION_` environment variables when they are set.**

Unfortunately, transmission-daemon does not seem to write settings.json from runtime settings until graceful exit, and this container does not appear to issue `SIGTERM` to `transmission-daemon` and wait for exit when shutting down (tested with `docker stop` and `docker restart` commands). **In order to preserve settings changed at runtime, some work will need to be done to examine container process handling.** I will have to examine this, given PID 1 is the shell script which spawns openvpn which runs the script that starts transmission-daemon...

MeanderingCode:
Latest commit changes `transmission/stop.sh` to give transmission-daemon up to 2 seconds to shut down gracefully, allowing it time to write its settings to settings.json.

MeanderingCode:
@haugene This PR is ready for review.

pkishino:
Looks great, will try and test it today

pkishino:
hmm,cloned your repo, build with docker and launched with just basic details, but settings were not persisted across restart of container. could not see the settings change in settings.json file neither :/

MeanderingCode:
Hmm. Can you post the end of the log output when the container is shutting down? Just from "sending kill signal to transmission-daemon".

pkishino:
my bad..I built the wrong branch, tested the feature branch and works well :)

haugene:
I get the flow of most of this, and I like it 👍. I think this will be a good simplification for most people on how to manage the Transmission configuration. I'm not sure we need to keep backward compatibility with the environment-way of doing it, but it would probably cause less noise when it's rolled out.

About that, it seems to me that settings will be persisted to settings.json on the first start (when settings.json don't exist) but will not be updated later if settings are changed in environment variables? I see you're writing that it:
> always respect TRANSMISSION_ environment variables when they are set
but I think that the `-no-overwrite` in transmission/start.sh prevents this, or am I doing something wrong here?

Tested it by doing a `docker-compose down` to remove all volumes etc. Then added some random config `TRANSMISSION_ALT_SPEED_UP=1234` and ran `docker-compose up`. It was then loaded correctly. But if I first run `up` and then set the env variable, and then restart it - it doesn't change.

Because I'm using PIA and right now running without rpc user/pass I'm also getting:
```
openvpn         | head: cannot open '/config/transmission-credentials.txt' for reading: No such file or directory
openvpn         | tail: cannot open '/config/transmission-credentials.txt' for reading: No such file or directory
```
because the file needs to exist even if there is no username/password. but that's minor stuff.

I see your `DEFAULT_TR_...` environment variables, and agree that they are needed for some of the variables where we have to know ports for ufw. When it comes to the rpc username/password I've seen there is a `--peerport` option on transmission-daemon. So we might just skip that part of the scripts that call the rpc endpoint and instead run the port scripts before we start transmission and just pass it as an option.

MeanderingCode:
> I get the flow of most of this, and I like it +1. I think this will be a good simplification for most people on how to manage the Transmission configuration. I'm not sure we need to keep backward compatibility with the environment-way of doing it, but it would probably cause less noise when it's rolled out.

I like the idea of supporting envvars, just to keep in line with expected POSIX behavior and application behavior.

> About that, it seems to me that settings will be persisted to settings.json on the first start (when settings.json don't exist) but will not be updated later if settings are changed in environment variables? I see you're writing that it:
> 
> > always respect TRANSMISSION_ environment variables when they are set
>  but I think that the `-no-overwrite` in transmission/start.sh prevents this, or am I doing something wrong here?

It will not re-write the template out. Whether an option set in an envvar differing from something in `settings.json` is written out to `settings.json` when `transmission-daemon` exits (as settings changed at runtime are) is up to the `transmission-daemon` behavior.

Regardless, my testing of Transmission's behavior is that an environment variable will always take precedence over something read from `settings.json` at startup, so envvars will be "effective" at container start time.

> Tested it by doing a `docker-compose down` to remove all volumes etc. Then added some random config `TRANSMISSION_ALT_SPEED_UP=1234` and ran `docker-compose up`. It was then loaded correctly. But if I first run `up` and then set the env variable, and then restart it - it doesn't change.

This is a misunderstanding of Docker behavior. Docker will not change environment variables for existing containers (without hacking around it by writing into the json files dockerd keeps (`/var/lib/docker/containers/<container_id>/config.v2.json` by default). By design, Docker only sets all container properties (volumes, envvars, ports, etc) at container creation time.

> Because I'm using PIA and right now running without rpc user/pass I'm also getting:
> 
> ```
> openvpn         | head: cannot open '/config/transmission-credentials.txt' for reading: No such file or directory
> openvpn         | tail: cannot open '/config/transmission-credentials.txt' for reading: No such file or directory
> ```
>
> because the file needs to exist even if there is no username/password. but that's minor stuff.

Missed this, as I don't use it. The assumption I made in these changes was that one would have to set the username and password environment variables if you were to use a provider where those extra helper scripts were to be used. This should be changed to be less error prone / brittle. I'm open to suggestions, but will take a look at it myself in the coming days.

> I see your `DEFAULT_TR_...` environment variables, and agree that they are needed for some of the variables where we have to know ports for ufw. When it comes to the rpc username/password I've seen there is a `--peerport` option on transmission-daemon. So we might just skip that part of the scripts that call the rpc endpoint and instead run the port scripts before we start transmission and just pass it as an option.

I haven't looked into those scripts to know the logic they're trying to implement (the goal is for an open port for inbound connections through the VPN tunnel, yes?) If you have an idea for changing that setup, let me know. IIRC, there are a few providers which assign you a port after the tunnel is established. If you're suggesting setting the `TRANSMISSION_PEER_PORT` envvar at that point, before transmission starts, I like that. If there is a chance of that port changing during runtime, RPC commands might be the only way to inform transmission about that.

MeanderingCode:
Sorry about the forced pushes. Tried to get this out ASAP for your perusal.

https://github.com/haugene/docker-transmission-openvpn/pull/1131/commits/a11733eea2bd7d7c5cb0022b6f64d91ec77de5bb should address setting the peer port before startup of the transmission daemon, eliminating the need for user/pass or any RPC calls.

Note: It no longer blocks ports in UFW which were enabled during `/etc/openvpn/start.sh`. I do not think this is a concern, but let me know and I can build logic in to check those ports.

MeanderingCode:
Took a look at changing UFW settings and realized it would be better to move the whole thing to `transmission/start.sh`.

Testing appreciated!

pkishino:
Sorry, haven’t gotten around to test latest updates.. wanted to last week, but.. I’ll try to give it a go this coming week and give some feedback

MeanderingCode:
We should really have someone using PIA and Perfect Privacy providers test these changes.

pkishino:
@haugene @MeanderingCode 
What needs to be done before we can merge this? Just some provider specific testing?

MeanderingCode:
I have not tested the use of UFW settings with these changes (which are now impacted).

I would also feel more comfortable if someone using PIA and Perfect Privacy providers tested these changes.

MeanderingCode:
And just for discussion's sake, since i don't know we resolved it:

Because I changed the environment variable script template to `dockerize --no-overwrite`, environment variables set (e.g. in docker-compose) will *not* take precedence over values changed in settings.

I think i misrepresented that or did not make it clear in my initial description. Since Transmission only really reads `TRANSMISSION_HOME` from envvars, then envvars set for running containers will only really be used in the case that a settings file does not exist (mostly, first-run).

If there is interest in changing that behavior so that container creation or restart "resets" any setting for which there is an envvar explicitly set by the user, it would only be a little work to accomplish.

MeanderingCode:
Documentation on what environment variables Transmission reads: https://github.com/transmission/transmission/wiki/Environment-Variables

If we want to see all container starts write settings from environment variables (overwriting settings.json values directly written or from previous runtime settings changes) in order to align to fairly standard expectations of environment variable behavior, let me know.

Otherwise, I can see if PIA and Perfect Privacy offer trials, and test those and UFW settings, myself, in the coming week or two.

pkishino:
Thanks for keeping up the good work, myself am quite busy these days and I believe @haugene is enjoying the nice oslo weather ;) 
We will get to this sooner or later I am sure ;)

haugene:
Hey guys. First off, apologies for the late response here. But I agree with @pkishino - we want to merge this sooner, rather than later.

I believe that this could be merged to the dev branch soon. For a test period we would just merge other PRs straight to master and then keep dev up to date as best we can. I can test the port updater scheme for PIA, I have a subscription there.

Before we merge I think we should have a plan for the "priority of configs" or what to call it. I'm open to discuss it, but I wish to keep at least a partial backwards compatibility. In addition to compatibility we also need a way to set certain settings from our startup-scripts. I think that exporting an env-var could be a good way of doing this as a general solution.

Slightly into the future I want to get rid of many of these "if provider=nordvpn then ..." from the common scripts. I want to make many "hooks" where scripts would be triggered per provider, if they have one. Kind of like the pre/post startup scripts we support for mounting. At this point, the env-vars setting can be a generic contract of sorts where provider specific scripts can change the Transmission settings.

Anyways. A suggestion for a solution. Let's include a default settings.json in the image, which is not a template but a plain settings.json. If there is no settings.json in the TRANSMISSION_HOME folder then we use the default and copy it in place.
If there is a settings.json there, we use it as a base and merge it with whatever is given through the environment. So we'll change the template file so that if a variable is empty, we don't generate the config line.

If I only set TRANSMISSION_ALT_SPEED_ENABLED then the result of the template should be:
```
{
  "alt-speed-enabled": true
}
```

Then we can do `cat default-settings.json user-settings.json | jq -s add` and I think we have our result.
See output of this example:
```
$ echo '{"name": "Bob", "age": 40} {"name": "Bob", "age": 80}' | jq -s add
{
  "name": "Bob",
  "age": 80
}
```

Thoughts. Am I missing something?

Issue #1110, opened 2020-03-31
[Improvement] Automatic bundle refresh


gpongelli:
**Is your feature request related to a problem? Please describe.**
This improvement is related to my pull request #1104 : try to find a way to automatically being able to download new bundle, sed content if needed, move the file, restart OpenVPN.

**Describe the solution you'd like**
I think about something like:

- having an "update.sh" file under each provider folder

- call the "provider/update.sh", if present, when something bad happen (tunnel down? try to check if new bundle is available; if so download it and try to use new bundle, maybe this will resolve the issue)

- this will automatically solve the connection issue and surely decrease the issue/PR here about new bundles

- last point: the provider/update.sh shall be executed also on container startup, to address the case of docker image with older certificate bundle

**Describe alternatives you've considered**
It's the way updateFreeVPN.sh should work, but managed for every folder that contains the update script.

**Additional context**
Please also refer to my PR, seems something is not working due to my lack of knowledge on how the setup/teardown cycle is done.

gpongelli:
Just for further investigation, that I've noted on UpdateFreeVPN.sh, it's needed to have a connected interface to internet to do the upgrade (downloading new certificates) BEFORE opening the tunnel again (or for the first time).
Is this feasible ? If not, this improvement cannot be reached.

haugene:
Totally agree on the ultimate goal here. I've kind of tried this before as well. There's a script at `/etc/openvpn/updateConfigs.sh` that will search the provider directories for "updateConfigs.sh" script and execute them. It was intended as a build step in the image to reduce the amount of Issues and PRs of invalid provider configs. Triggering the provider scripts on startup would be a natural next step.

Some providers do not have a public URL where their .ovpn files can be retrieved. I'm not sure exactly why, but users needs to be logged in to get them. So this will not work for all providers. Still it needs to be extracted from this repo if it is to be maintainable.

I've been working on a feature to provide OpenVPN config using a URL and then the image would download it. The idea is that all less popular, or less open providers could be supported by just uploading the config to dropbox and sharing it with a public URL.

Have a look at this branch, and compare it to `dev` so you don't get all the provider stuff that is diff against master. https://github.com/haugene/docker-transmission-openvpn/tree/feature/external-config

Basically I want to split the per-provider behaviour to a separate script for each provider.
[...]

PR #1133, opened 2020-04-11
Automate PrivateVPN configs

arpaulnet:
The existing PrivateVPN configuration files do not span all of the PrivateVPN servers or options.  This update contains an `updateConfigs.sh` script along with new and updated `*.ovpn` files for PrivateVPN.

The script does the following:
1. Downloads/unpacks a zip file from PrivateVPN's support
2. Modifies the config files to match the previous OpenVPN options
3. Generates additional optional config files:
   - `tcp` = use `tcp` port `443`
   - `strong` = use `AES-256-CBC` cipher
   - `dedicated` = uses PrivateVPN's "dedicated ip" option (note: all ports are open, beware `WEBPROXY_ENABLED`)
4. Organizes the config file names to be relatively human-friendly
5. Creates links to maintain backward-compatibility with existing references to configuration files


haugene:
Great stuff, thanks for contributing 👍 
Always great to get update scripts for the providers and it's a good fit with #1110 initiative going on.

Merging now to dev, but will do a new release right away so it will be on :latest and release tag very soon. I guess you've already tested this locally, but if you could also verify it in the final image that would be great.

arpaulnet:
Thanks 😄. I spot-checked a handful of the connections in the final image, and they were working as expected.

doctorjei:
This commit seemed to have removed the port-forwarding update/fix from issue #1061. I think we need an addition to automatically generate the port-forward versions.

doctorjei:
I fixed the port-forwarding issue and added it to the PrivateVPN automation script: https://github.com/haugene/docker-transmission-openvpn/pull/1587

PR #1139, opened 2020-04-16
Add missing supported providers

chrismuellner:
Update documentation to include missing supported providers


haugene:
Yes indeed, those are missing. Thanks!

PR #1158, opened 2020-04-26
Updated ProtonVPN OVPN config files to add new servers

Radicitus:


casaper:
@pkishino It seems my container seazed to work since I pulled the new image containing this change with protonvpn.

What Might be the reason for this?

It either logs that it is unsuccessfully looking for a default.ovpn because I don't have a OPENVPN_CONFIG variable set.

I never had one set tho, and It used to just work before. And If I set one, I didn't manage to find out what the correct value for this would be?  
The name of the config files in the images `/etc/openvpn/protonvpn/*.ovpn` files?

The documentation sais something about an example with `UK Bristol` but assuming I live in Switzerland that doesn't really apply (Switzerland is the host country of protonvpn BTW)

So what do I need to do to get the vpn with protonvpn up and running again now?

casaper:
> @pkishino
> @Radicitus

> So what do I need to do to get the vpn with protonvpn up and running again now?

And I now 100% verified that it really must be that change that broke protonvpn to work for me with this image.

Using `image: haugene/transmission-openvpn:2.12-armhf` it works fine... 

For now I'm ok with this then, but perhaps we need to unbreak proton again, because I presume I'm not the only protonvpn user with that trouble?

pkishino:
this merge isnt part of any release yet.. 2.12 is veery old

pkishino:
you should specify which server you want to use in OPENVPN_CONFIG (see the protonvpn folder for examples.. such as CH-1 etc)
using CUSTOM should work if you fix the ovpn file (by default it wont work as it requires authentication..look at example in protonvpn folder)

casaper:
> you should specify which server you want to use in OPENVPN_CONFIG (see the protonvpn folder for examples.. such as CH-1 etc)

I see. Ok. I'll give that a try. 

But never the less, most servers of protonvpn won't allow P2P. I'ts a limited set only that will allow this.
I created a PR, only supplying the openvpn configs with the P2P flag:
https://github.com/haugene/docker-transmission-openvpn/pull/1170

> using CUSTOM should work if you fix the ovpn file (by default it wont work as it requires authentication..look at example in protonvpn folder)

Creating the https://github.com/haugene/docker-transmission-openvpn/pull/1170 I learned that mistake of mine in the diffs. I need to reference the config file the image generates from the variables provided in the docker compose or the docker run command.

casaper:
@pkishino 

But the default configs from proton vpn also differ at two other points:

protonvpn downloaded: 

```
comp-lzo no

script-security 2
up /etc/openvpn/update-resolv-conf
down /etc/openvpn/update-resolv-conf

```

The haugene/docker-transmission-openvpn configs have `compress` instead of the `comp-lzo no` and the lower section with script security and the up and down script definitions are not there.
Do I need to adjust these as well?

casaper:
@pkishino 

> see the protonvpn folder for examples.. such as CH-1 etc

So the script just takes the file name part before the first full stop as screaming snake then?

pkishino:
thats how it works for most providers I believe..regarding the details of
protonvpn, sorry, I do not know those..
and regarding your PR, that is fine, but again, this PR is merged in, yes,
but if you have pulled latest this is NOT part of that release.. so if
something is broken, it's not due to this PR

casaper:
@pkishino 

> thats how it works for most providers I believe..regarding the details of protonvpn, sorry, I do not know those.. and regarding your PR, that is fine, but again, this PR is merged in, yes, but if you have pulled latest this is NOT part of that release.. so if something is broken, it's not due to this PR

Sorry I missed that part. Yea, thats my bad. ;)

But besides me missing that fact:  
It makes no sense to add configs for VPN servers that don't allow p2p here, Right?

casaper:
@pkishino 
> you should specify which server you want to use in OPENVPN_CONFIG (see the protonvpn folder for examples.. such as CH-1 etc)

I now tried `CH-5` `CH-05` `CH 5` and `CH 05`. None works for me. They all lead to this:

```
transmission_1  | allowing 192.168.0.0/24 through the firewall to port 5060
transmission_1  | Rule added
transmission_1  | Options error: In [CMD-LINE]:1: Error opening configuration file: /etc/openvpn/protonvpn/default.ovpn
transmission_1  | Use --help for more information.
```

casaper:
But @pkishino  since you made me realize that I need to adapt the downloaded configs from proton vpn, you enabled me to run the thing with `PROVIDER=CUSTOM` which is also fine for me.

Thx. Sorry for bothering you with so many comments here.

Radicitus:
Hi there,

I can't really discern if the PR I made caused these issues, but I think I did make a mistake and might not have included any default configs when I updated all of the server configs that were added (which was a loooot) but instead only included individual global server configs. So you would need to specify specific servers to connect to instead of it choosing one for you, although I think you would get better performance by selecting servers in your region anyway.

casaper:
@Radicitus 
> I can't really discern if the PR I made caused these issues, but I think I did make a mistake and might not have included any default configs when I updated all of the server configs that were added (which was a loooot) but instead only included individual global server configs. 

As far as figured out your mr actually didn't, It will have been something else I presume.

> So you would need to specify specific servers to connect to instead of it choosing one for you, although I think you would get better performance by selecting servers in your region anyway.

Agree. It probably is, And since that fault I switched to using a custom config mounted into `/etc/openvpn/custom/*.ovpn` now, since I realized that there is not only p2p flagged servers, but also plus servers at protonvpn. Its only for the mid level subscriptions, which I have, so supposedly it runs faster now.
Actually speed doesn't really matter that much to me, but still won't hurt to have more of it anyway.

casaper:
@Radicitus 

Besides that, perhaps consider reviewing the PR https://github.com/haugene/docker-transmission-openvpn/pull/1158 of mine.

I've added only the protonvpn configs that are flagged p2p elegible by protonvpn.

I'm not sure if a server not being flagged p2p would mean that p2p won't work?  
Thats basically also why I'm mentioning this to you.

As you can see in this screenshot, only some of the swiss servers are with the p2p flag. None of the Taiwan servers has such a flag. Only Netherlands, Iceland, Singapore and Sweden servers all have that p2p flag.

image

Radicitus:
@casaper 

Sorry for the late reply, I was preoccupied with family matters.

I was wondering where you got the information for which servers were p2p flagged? From my experience, I've never noticed an issue with this but you seem to be right and only certain servers should be included. I have also seem posts about how ProtonVPN automatically reroutes your traffic to a p2p supported server if it detects you are torrenting, even if the server you are connected to doesn't support it. So maybe this isn't an issue after all?

I've also noticed that the ovpn files are slightly different for different operating systems, and I added all the MacOS ones. I haven't noticed any issues so it seems like a non-issue, but I wonder if this is causing problems for some people. It appears DNS leaks could be a problem when using configs not suitable for your OS.

The default config file is nowhere to be found on the ProtonVPN downloads page, so I wonder if anything can be done there, but as you have said with so many different tiers and p2p flags it may be more complex that it appears. I personally believe specifying servers is most beneficial, but does remove the ease of use factor that @RomainL972 seems to be having a problem with.

RomainL972:
I noticed it works with "OPENVPN_CONFIG=nl-01.protonvpn.com.udp" as a workaround while there isn't a default configuration

Issue #1163, opened 2020-04-29
Set timezone at run time


edgd1er:
**Is your feature request related to a problem? Please describe.**

The container is by default set to UTC time which is not my local time. As a result, scheduled alternate speed does not kick in at the expected hours.

**Describe the solution you'd like**

When The TZ variable is not null, define localtime with $TZ , and reconfigure tzdata, the package being already present in the image. The operation does not take long, it could be inserted in /etc/openvpn/start.sh before the last command.

```
[[ -n ${TZ} ]] && ln -sf /usr/share/zoneinfo/${TZ} /etc/localtime && dpkg-reconfigure -f non-interactive tzdata
``` 

**Describe alternatives you've considered**

mounting /etc/localtime is unfortunately not an option for me.

**Additional context**

haugene:
I'm pretty sure that this already works out of the box with the tzdata package?

Ran this now locally:
```
 ~ docker run --rm -it -e TZ=Europe/London haugene/transmission-openvpn bash -c "date"                                                           
Wed Apr 29 22:50:04 BST 2020
 
~ docker run --rm -it -e TZ=Europe/Paris haugene/transmission-openvpn bash -c "date"                                                            
Wed Apr 29 23:50:09 CEST 2020
```

Setting a valid TZ environment variable from [the list](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) of supported zones should do the trick.

edgd1er:
Well, I focus may be too much on the timezone ;) The time seems to be ok. With your example, the timezone stays UTC. 

if you run: `docker run --rm -it -e TZ=Europe/Paris haugene/transmission-openvpn bash -c "date;cat /etc/timezone;ln -sf /usr/share/zoneinfo/Europe/London /etc/localtime;dpkg-reconfigure tzdata; cat /etc/timezone" ` you will see that the timezone is then properly set.

the result is:
```
Thu Apr 30 00:11:30 CEST 2020
Etc/UTC

Current default time zone: 'Europe/London'
Local time is now:      Wed Apr 29 23:11:31 BST 2020.
Universal Time is now:  Wed Apr 29 22:11:31 UTC 2020.

Europe/London
```

[...]

PR #1164, opened 2020-04-29
setup timezone at run time

edgd1er:
merge proposal for #1163 

define timezone at container startup.


edgd1er:
as discussed in #1163 , this is not necessary.

PR #1170, opened 2020-05-02
Fix/protonvpn configs resolve #1171

### LINKED ISSUE #1171: ProtonVPN only with configs elegible for p2p
**Which provider?**
ProtonVPN

**Where are the configs?**
> Most providers supply a .zip file containing all configs and certificates. Please provide a link to the .ovpn config bundle

ProtonVPN offers the files individually in an authenticated member section of their page, as far as I know.

I discovered that this image contains all the proton VPN configs available. The problem is, that only a fraction of the servers will allow p2p. So I think it doesn't make sense to carry all these configs, that will most likely not work well or not at all in this image, since this image is specifically for p2p.

casaper:
~~Probably the PR https://github.com/haugene/docker-transmission-openvpn/pull/1158 broke proton vpn connections. At least for me it did.~~

Since I found out that #1158 is not yet in the latest image builds that assumption must be wrong. I'ts something else that broke the before working config for me. Will have to find out what that was.

This PR is an effort to get that problem fixed again.

For a start, proton vpn only flags some of its servers in some of its countries as p2p elegible. I removed all the vpn configs not flaged p2p and added server and country configs with the p2p flag.

Most countriy servers of proton vpn don't seem to allow p2p, only the ones I add in this PR.

fixes #1171

> Please bare with me. I'm very used to GitLab, and thus have no clue how to link an issue to a PR in GitHub. It seems the interface isn't self explenatory to me. Sorry about that.


casaper:
@pkishino Do I need to use `compress` or the downloaded `comp-lzo no` from protonvpns original here?

openvpn/protonvpn/ch.protonvpn.com.tcp.ovpn
```diff
@@ -48,7 +48,7 @@ persist-tun
 reneg-sec 0
 
 remote-cert-tls server
-auth-user-pass /config/openvpn-credentials.txt
+auth-user-pass
```
pkishino, line 51: why are you removing the credentials file? This is created by the container when run to provide the details..

pkishino: if you can explain why this was removed, then we cna look at merging this

pkishino: @casaper can you please comment?

casaper: oh

@pkishino  Sorry, I kind of lost track here. 

No, that change was accidental. Must have been a glitch when I reverted these changes and this was probably forgotten. 

So you're perfectly right. This change mustn't be merged, because it would render the config broken if used.

I'm sorry, I kinda forgot about having placed this PR  here.

I do not insist in keeping it.

Someone stated the config also works fine for her/him even with servers not deemed p2p eligible by protonvpns page, so I guess we can safely drop my PR here.

Sorry to have kept you waiting.

PR #1185, opened 2020-05-11
Removed space within parsed VPNBOOK password.

colinwebber:
VPNBOOK password is provided as an image. api.ocr.space/parse/image API erroneously sees a space between certain character combinations. Since passwords generally do not contain spaces, this extra awk expression removes the space.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,7 @@ workflows:
 jobs:
   build-ubuntu-image:
     environment:
-      IMAGE_NAME: haugene/transmission-openvpn
+      IMAGE_NAME: colinwebber/transmission-openvpn
     docker:
       - image: circleci/buildpack-deps:stretch
     steps:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -53,7 +53,7 @@ workflows:
 jobs:
   build-ubuntu-image:
     environment:
-      IMAGE_NAME: haugene/transmission-openvpn
+      IMAGE_NAME: colinwebber/transmission-openvpn
```
pkishino, line 56: this file should not be part of the commit as this will link to the wrong image

pkishino: Ah, so , if you remove the commit b97d817 then it will be fine

colinwebber: Apologies, I did that for testing purposes but committed to the same branch. I'm going to create a new pull request to avoid any remnants.

PR #1187, opened 2020-05-12
Removed space within parsed VPNBOOK password.

colinwebber:
VPNBOOK password is provided as a picture/image. The http://api.ocr.space/parse/image service erroneously sees a space between certain character combinations. Since passwords generally do not contain spaces, this extra awk expression removes the space.


pkishino:
This looks good to me

PR #1188, opened 2020-05-12
Giganews Provider Update 

steamGerbil:
The Giganews provider was not working for me. I have downloaded the latest ovpn files and replaced the old ones. These are now alsmost identical to the VyprVPN items, the only difference being the server endpoint


.circleci/config.yml
```diff
@@ -53,7 +53,7 @@ workflows:
 jobs:
   build-ubuntu-image:
     environment:
-      IMAGE_NAME: haugene/transmission-openvpn
+      IMAGE_NAME: steamgerbil/transmission-openvpn
```
pkishino, line 56: This is not right, please rebase and squash your commits and remove these circle config as they are pointing at the wrong image

steamGerbil: Sorry I thought I was on a cloned repo and not the fork, I've reverted those changes so only config changes are in. I'll rebase and push

steamGerbil:
All changes reverted to config only and the commits have been rebased and squashed

steamGerbil:
Saw that default.ovpn wasn't a symbolic link, corrected it

PR #1195, opened 2020-05-21
Add automated forwarded port configuration for PrivateVPN

caseyfw:
Had to MITM attack the Mac PrivateVPN client to sniff out the mechanism it uses to fetch the forwarded port you are assigned on connection. Despite this, the implementation was straight forward as it's just an unauthenticated fetch to their API with your local tunnel IP in the query string.

I built the image locally and span it up with `OPENVPN_PROVIDER=PRIVATEVPN` and `OPENVPN_CONFIG=SE Stockholm` and it configured the port correctly:

```
-------------------------------------
Transmission will run as
-------------------------------------
User name:   abc
User uid:    1001
User gid:    1001
-------------------------------------

STARTING TRANSMISSION
CONFIGURING PORT FORWARDING
Thu May 21 22:31:54 2020 /sbin/ip route add 193.180.119.2/32 via 172.17.0.1
Thu May 21 22:31:54 2020 /sbin/ip route add 0.0.0.0/1 via 10.35.15.1
Thu May 21 22:31:54 2020 /sbin/ip route add 128.0.0.0/1 via 10.35.15.1
Thu May 21 22:31:54 2020 Initialization Sequence Completed
Got new port 53125 from PrivateVPN API
transmission auth required
localhost:9091/transmission/rpc/ responded: "success"
Checking port...
Port is open: Yes
```

Thanks again for making this, it's a fantastic packaged solution. Let me know if there's any amendments you would like, cheers!

### DIFF
--- a/transmission/updatePrivateVPNPort.sh
+++ b/transmission/updatePrivateVPNPort.sh
@@ -1,0 +1,84 @@
+#! /bin/bash
+
+# Source our persisted env variables from container startup
+. /etc/transmission/environment-variables.sh
+
+# Settings
+TRANSMISSION_PASSWD_FILE=/config/transmission-credentials.txt
+
+transmission_username=$(head -1 $TRANSMISSION_PASSWD_FILE)
+transmission_passwd=$(tail -1 $TRANSMISSION_PASSWD_FILE)
+transmission_settings_file=${TRANSMISSION_HOME}/settings.json
+
+#
+# Fetch forwarded port from PrivateVPN API
+#
+
+# Get the port
+tun_ip=$(ip address show dev tun0 | grep 'inet\b' | awk '{print $2}' | cut -d/ -f1)
+pvpn_get_port_url="https://xu515.pvdatanet.com/v3/mac/port?ip%5B%5D=$tun_ip"
+pvpn_response=$(curl -s -f "$pvpn_get_port_url")
+pvpn_curl_exit_code=$?
+
+if [[ -z "$pvpn_response" ]]; then
+    echo "PrivateVPN port forward API returned a bad response"
+fi
+
+# Check for curl error (curl will fail on HTTP errors with -f flag)
+if [[ ${pvpn_curl_exit_code} -ne 0 ]]; then
+    echo "curl encountered an error looking up forwarded port: $pvpn_curl_exit_code"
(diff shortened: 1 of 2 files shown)


pkishino:
I’ll try to test this this weekend, I use privatevpn but with a custom config

caseyfw:
@pkishino how did you go?

pkishino:
Ok, a few comments. Just tried it out now by building with alpine..
first of all, alpine grep does not support grep -P option.. so that is a problem
if I build it with the normal Dockerfile, then it works
second, if I stop and start the container, it doesnt run again..but this is a separate issue
third, interestingly, on some servers where I have "custom" configs I got from support, working with PFW, it says that it does not support this, most likely to do with outdated configs though..
also, i noticed that the configs need to be updated..

pkishino:
but regarding this commit, it will break in alpine, @haugene I believe you want to keep alpine support, right?
If that can be sorted, then I'm happy to approve

caseyfw:
Nah, don't approve - I'll fix it to work with alpine 👍

caseyfw:
Ok, fixed - I had a look at other mechanisms for getting an interface's IP4 address without negative lookahead and arrived at a combination of grep/awk/cut shown here: https://askubuntu.com/a/560466/918834

```
➜ docker run -it --rm alpine                 
/ # ip address show dev eth0 | grep 'inet\b' | awk '{print $2}' | cut -d/ -f1
172.17.0.2
```

transmission/updatePrivateVPNPort.sh
```diff
@@ -0,0 +1,84 @@
...
+
+# Check for errors in curl response
+error=$(echo "$pvpn_response" | grep -o "\"Not supported\"")
+if [[ ! -z "$error" ]]; then
+    echo "PrivateVPN API returned an error: $error - not all PrivateVPN servers support port forwarding. Try 'SE Stockholm'."
+    exit
+fi
+
+# Get new port, check if empty
+new_port=$(echo "$pvpn_response" | grep -oP "Port \K[0-9]*")
```
pkishino, line 41: This -P also needs reworked please ;0

pkishino:
is this still working for you? I think they have changed the server/api again..

PR #1208, opened 2020-05-25
Default UTP false

### LINKED ISSUE #978: default TRANSMISSION_UTP_ENABLED to false
**Is your feature request related to a problem? Please describe.**
*really* hurts bandwidth 

**Describe the solution you'd like**
by using false as default

**Describe alternatives you've considered**
none

**Additional context**
I can only use 10% of my bandwidth when uTP is enabled. Wihtout it I reach my full bandwidth. And I don't seem to be alone if google is any indication.

pkishino:
fixes #978


haugene:
Looks good. You're really burning through issues these days. Thanks a lot for your effort!

Adding a reference to #1131. We need to make sure we keep the good defaults like this when we move to persistent settings.

PR #1224, opened 2020-06-02
Updated healthcheck to include openvpn and transmission process

### LINKED ISSUE #1189: Restart OpenVPN within the Container
Hello,

I have an issue with ProXPN where the VPN connectivity works for a while (number of days), then randomly loses internet connectivity. The VPN is showing as connected, however there is no internet connectivity. Restarting the container fixes the issue.

As this issue happens at a random time, I'd like to create a bashscript in a cronjob to check that the container has external internet connectivity, and if it doesn't to restart openvpn.

I've been playing around in the container, and can't get to grips with how the openvpn connection is handled, it doesn't seem to be running as a service... How can you restart the openvpn connection with the same settings as when you created the container, within the container itself?

Thanks,

pkishino:
container will be marked unhealthy when openvpn or transmission process is not running.
it will be up to the user to handle what to do then, (recommend autoheal)
fix #1189

### DIFF
--- a/scripts/healthcheck.sh
+++ b/scripts/healthcheck.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
+
+#Network check
 # Ping uses both exit codes 1 and 2. Exit code 2 cannot be used for docker health checks,
 # therefore we use this script to catch error code 2
 HOST=${HEALTH_CHECK_HOST}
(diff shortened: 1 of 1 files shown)


skrenes:
Thanks for this fix. I was wondering, however, why mark itself as unhealthy instead of just killing itself (allowing the restart policy to take over)?

pkishino:
For standalone containers, Docker does not have native integration to restart the container on health check failure though we can achieve the same using Docker events and a script. Health check is better integrated with Swarm. With health check integrated to Swarm, when a container in a service is unhealthy, Swarm automatically shuts down the unhealthy container and starts a new container to maintain the container count as specified in the replica count of a service.

Restarting of unhealty container feature was in the original PR (https://github.com/moby/moby/pull/22719), but was removed after a discussion and considered to be done later as enhancement of RestartPolicy.

At this moment you can use this workaround to automatically restarting unhealty containers: https://hub.docker.com/r/willfarrell/autoheal/

PR #1227, opened 2020-06-02
Update healthcheck.sh

DaftHonk:
`ps` does not return the process without `aux` or at a minimum `ax` as an option:

````
root@7b0bdc3ed3ad:/etc/scripts# ps
  PID TTY          TIME CMD
  225 pts/0    00:00:00 bash
  405 pts/0    00:00:00 ps
root@7b0bdc3ed3ad:/etc/scripts# ps aux
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root         1  0.0  0.0    208     4 ?        Ss   11:49   0:00 dumb-init /etc/openvpn/start.sh
root         6  4.0  0.0  12144  5028 ?        Ss   11:49   1:33 openvpn --script-security 2 --up-delay --up /etc/openvpn/tunnelUp.sh --down /etc/openvpn/tunnelDown.sh --inactive 3600 --ping 10 --ping-exit 60 --config /etc...
abc         53  1.7  0.0 610152 21328 ?        Ssl  11:49   0:39 /usr/bin/transmission-daemon -g /data --logfile /data/transmission.log
root       225  0.0  0.0   4216  2700 pts/0    Ss   12:08   0:00 bash
root       406  0.0  0.0   5856  1880 pts/0    R+   12:27   0:00 ps aux
````

PR #1228, opened 2020-06-02
healthcheck

clambin:
PR fixes status check for transmission & openvpn daemons in healthcheck.

### DIFF
--- a/scripts/healthcheck.sh
+++ b/scripts/healthcheck.sh
@@ -23,8 +23,8 @@ echo "Network is up"
 
 #Service check
 #Expected output is 2 for both checks, 1 for process and 1 for grep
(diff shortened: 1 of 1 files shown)


DaftHonk:
I submitted a similar PR (#1227) but used aux instead of -ef as the option. im option is used, it is important that it does not list any dead or zombie processes.

clambin:
yeah, saw that after I submitted mine.  not sure I understand the reference to zombie processes, but 'ps -ef' and 'ps aux' are basically the same thing.  so either will work.

pkishino:
hmm, in alpine image it works fine just using ps..I'll merge this for now as -ef also works

Issue #1166, opened 2020-05-01
Add a request at startup to check the public IP


diyfr:
To verify that the openvpn configuration is successful
Sample command
```shell
curl icanhazip.com
```

pkishino:
There already is a health check that is run and verified that you are connected. I don’t quite get what it is you want added.

ilike2burnthing:
That site returns your IP in plain text, so I'm assuming the idea here is to fetch this to ensure you don't have an IP leak

diyfr:
> That site returns your IP in plain text, so I'm assuming the idea here is to fetch this to ensure you don't have an IP leak
That's it ! :)

pkishino:
this would be fine, could you add a simple PR for this? Were you thinking to compare the result with something or?

albertcarrete:
What about a _Prohibit IP_ setting where an ip address can be statically added as a configuration option. Then a process checks the result of icanhazip.com (or similar) against that setting and prevents leakage if it reports a matches.

pkishino:
would love some comments on the PR linked

Issue #1134, opened 2020-04-12
Prevent data leakage when tunnel is down


jtrott:
I've knocked together a quick startup script that removes the default gateway before openvpn is started and instead puts in static routes to the openvpn server endpoints.
This prevents any data leaking out from transmission if the tunnel happens to come down.
The script is hard coded to proxpn but could easily be changed and made much simpler if it can be added to the start.sh, which already knows which random config it has picked.
Not sure how to do a pull, so just added it here in case it is useful for anyone.

```
#!/bin/sh

# grab the current router IP
ROUTER=`/sbin/ip route list | /bin/grep default | /bin/grep eth0 | /usr/bin/awk '{print $3}'`

# delete the default route
/sbin/ip route del default via $ROUTER dev eth0

# Loop through provided ovpn config files
for i in $(echo $OPENVPN_CONFIG | sed "s/,/ /g")
do
    # get the server address from the config file
    SERVER_ADDR=`/bin/grep remote /etc/openvpn/proxpn/${i}.ovpn | /usr/bin/awk '{print $2}' | /usr/bin/sort -fu`
    # set a static route for each server entry
    /sbin/ip route add $SERVER_ADDR via $ROUTER dev eth0
done
```

Cycl:
Is this different from setting this env variable: DROP_DEFAULT_ROUTE=true ([used in this script](https://github.com/haugene/docker-transmission-openvpn/blob/0b6ab30294f139f06eaf74cf4808ec8c9c7e9db1/transmission/start.sh#L60))?

Though that one seems to do it after the tunnel is up, but before transmission is started.

I do also see the static route that this script also adds too - though not sure if that is done by default or through the ufw commands I run.

jtrott:
That variable seems to try and drop the default route if transmission fails, I think.
I'd much rather lock down the container networking before transmission is even launched.

pkishino:
Lets continue discussion on PR linked, please take a look

Issue #1127, opened 2020-04-07
IP Leaks / Broken Killswitch / Tunnel Hardening


a-n-d-i:
If I firewall off port 1194 on my Docker Host (or somewhere else down the line), the "killswitch" is not working / the container is using my local internet connection if the vpn is disrupted.

To reproduce:

Block VPN Traffic. In my setup this one does the trick, yours may vary. Be aware of the docker host / ufw issue (major unreleated concern btw, docker by default ignores host ufw rules, see for example https://github.com/chaifeng/ufw-docker). 

`iptables -I DOCKER-USER 1 -p udp --dport 1194 -j DROP`

OR just remove the VPN route, easier than messing with iptables

`route del -net 0.0.0.0 gw 10.8.8.1 netmask 128.0.0.0 dev tun0`

start the container, open a shell in the container, install w3m or similar and connect some "showmyip" service. (if you do this on a running container, stuff freezes and I was to impatient to wait for timeouts).

I guess the main problem is that relying on openvpns "tunnel down" trigger as a killswitch does not cover the time period until the connection times out. Also (almost) any traffic should be sent through the vpn to protect against possible vulnerabilities in transmission. 

### ufw config
It's pretty open, the container is allowed to initiate connections to pretty much anything.

> Logging: on (low)
> Default: deny (incoming), **allow (outgoing),** deny (routed)
> New profiles: skip
> 
> To                         Action      From
> --                         ------      ----
> 51413                      ALLOW IN    Anywhere                  
> 8888                       ALLOW IN    Anywhere                  
> 9091                       ALLOW IN    172.24.0.1                
> 9091                       ALLOW IN    192.168.0.0/16  

This should be much more restrictive, like deny all outgoing except vpn and name/dns. 
see: https://gist.github.com/Necklaces/18b68e80bf929ef99312b2d90d0cded2

### routing table

> Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
> 0.0.0.0         10.8.8.1        128.0.0.0       UG    0      0        0 tun0
> **0.0.0.0         172.24.0.1**      0.0.0.0         UG    0      0        0 eth0
> [...]

There is a default route for routing stuff without using the vpn tunnel. This might be removed (also depending on name/dns service I guess)

ovizii:
[...]

PR #1231, opened 2020-06-03
WIP at adding iptest and ip route harding

pkishino:
Hey,
I started this today with some input from #1134 and #1166 but got stuck with getting transmission port check to work.. would be great if someone could take a look and improve this, mostly just hacked together, not too familiar with the ip routing stuff
also could include ideas from #1127


martinruenz:
You might find the 'kill switch' implementation of protonvpn interesting, which is based on iptables: https://github.com/ProtonVPN/linux-cli/blob/1190545a4293332824b28fe442f95f7be44b65d5/protonvpn_cli/connection.py#L764

sylphaenos:
Hey there, this PR was originally supposed to close several other **very** important security-related issues (#1134, #1127). Is this still planned?

pkishino:
@haugene  some of this stuff might no longer be needed based on latest but there are still Harding things that are needed

haugene:
Ok. We should have a sync then. Try to sum up the issues on potential DNS leak and route hardening and find a good way to introduce it where we try not to break things. I think most of the hardening should be a well documented opt-in feature to begin with. Try it out on those who are most interested in it and then change the defaults gradually.

sylphaenos:
Sorry to be a bother, but could we get this PR excluded from stale bot (or should I open another issue for security hardening?)

pkishino:
I think a new PR with updates following latest improvements would be best, haven’t had much time recently to dev..

PR #1232, opened 2020-06-03
Adds regexp support to rss

### LINKED ISSUE #1167: Add RSS feed with regex?
Hey All.

I am using the docker image in this repo :+1: 

I want to add an RSS feed with regex included to this docker setup, but I can't figure out how to set it up correctly 😕 

I looked at the documentation here for the RSS plugin.
https://haugene.github.io/docker-transmission-openvpn/rss-plugin

I can see there is the "-e" flag to add an URL, but I can't figure out how to do regex on this URL. I think this is possible because of the original repo for transmission-rss supports this with the "regexp: foo" tag.

Thanks in advance for the help

**Host system:**
Ubuntu 19.04 Server version, Docker version 18.09.7

pkishino:
adds support for  regexp, resolve #1167

### DIFF
--- a/plugins/rss/Dockerfile
+++ b/plugins/rss/Dockerfile
@@ -11,6 +11,7 @@ RUN apt-get update \
 ADD . /etc/transmission-rss
 
 ENV TRANSMISSION_DOWNLOAD_DIR=/data/completed \
-    RSS_URL=**None**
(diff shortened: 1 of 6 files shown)


haugene:
Sorry about the delay. Merging now.

PR #1296, opened 2020-07-17
updates for pia server changes

jerizm:
PIA made some server updates according to https://www.privateinternetaccess.com/helpdesk/news/posts/july-16-2020-important-updated-july-16-server-changes-and-related-issues

Also noticed us-newyorkcity is now us-nyc since it's the one I used. Not sure what other updates were unannounced.


wedemmoez:
I submitted the PR #1294 to update the US Texas to US Dallas, but didn't even think to check other regions. Good catch!

PR #1317, opened 2020-08-06
Naming community repo as community

haugene:
Installing ufw from the community repository worked, but we should probably call it community and not testing.


haugene:
Just merging this now to get it into dev and then start with removing the ubuntu based image. Tagged you so you could shout out if you disagreed.

Issue #1288, opened 2020-07-11
Dockerize failing with "Exec format error" in arm64 image


Mhmccabe:

pkishino:
Please post config, which example are you referring to?

Mhmccabe:
a variation of the one from the dockerhub page https://hub.docker.com/r/haugene/transmission-openvpn/


version: '3.3'
services:
    transmission-openvpn:
        volumes:
            - '/share/download/:/data'
            - '/etc/localtime:/etc/localtime:ro'
        environment:
            - CREATE_TUN_DEVICE=true
            - OPENVPN_PROVIDER=VYPRVPN
            - OPENVPN_CONFIG=Argentina-256
            - OPENVPN_USERNAME=***********
            - OPENVPN_PASSWORD=***********
            - WEBPROXY_ENABLED=false
            - LOCAL_NETWORK=192.168.0.0/16
            - PUID=1000 # id of the docker user
            - PGID=100 #Change to your PGID
            - OPENVPN_OPTS=--inactive 3600 --ping 10 --ping-exit 60 # Leave this as is
        cap_add:
            - NET_ADMIN
        ports:
            - '9091:9091'
            - '8888:8888' #Transmission Port
        image: haugene/transmission-openvpn:latest-arm64
        # container_name: vpn2
        # restart: unless-stopped

pkishino:
hmm,seems like there is an issue with dockerize generating settings from template
@haugene looks like dockerize not working ?

haugene:
`Exec format error` is an error message that I have seen relating to running on wrong cpu architecture before. Are you sure that you should run arm64 image and not armhf? Can you try `latest-armhf` instead?

Mhmccabe:
Hi, 
Thanks for taking the time to respond to me, its very much appreciated 

I can confirm, I am using the image haugene/transmission-openvpn:latest-arm64, the VPN connetion was being estblished, but the boot failed when it tried to configure transmission

I've ran it again with the image that you published yesterday, I am now getting the below


the VPN connection is established correcfly, I get the following when using latest:arm64
the CPU is an Arm Cortex a57

> Tue Jul 14 20:52:17 2020 OPTIONS IMPORT: data channel crypto options modified                                                                                                                                                                                                                                              
> Tue Jul 14 20:52:17 2020 Data Channel Encrypt: Cipher 'AES-256-GCM' initialized with 256 bit key                                                                                                                                                                                                                           
[...]

PR #1323, opened 2020-08-09
WIP: Persistent settings and drop dockerize

haugene:
Tried to continue the work of @MeanderingCode in #1131, but also allow for configuring using environment variables. Due to the issues with Dockerize (#1288) I wanted to try to remove it and use simple python scripts instead.

This is still a work-in-progress, but it would be great with some initial thoughts or testing.
I think there is a problem in that the default-settings.json is read into a dictionary and at that point there can be booleans, integers or strings. But when we update from the environment variables, they are always strings. It seems Transmission is quite liberal when it comes to parsing its config, but we should generate settings with the right data type I think.

#1131 also re-organises some code and moves logic from openvpn/start.sh to transmission/start.sh which is nice. I also have to update the "port update scripts". Interested to see how it plays on the arm versions. Anyways, I think it looks relatively clean so far. Curious about your thoughts.

@thomasborgen Any comments on my Python? We're basically just running two different apps so I want to persist selected environment variables on startup, and then they are sourced into the second script. In the second script I want to be able to base my settings on a default, but let them be overridden if a corresponding environment variable exists.


transmission/updateSettings.py
```diff
@@ -0,0 +1,37 @@
...
+transmission_settings = sys.argv[2]
+
+if not os.path.isfile(default_settings):
+    sys.exit('Invalid arguments, default settings file does not exist')
+
+# Define which file to base the config on
+if os.path.isfile(transmission_settings):
+    configuration_baseline=transmission_settings
+else:
+    configuration_baseline=default_settings
```
thomasborgen, line 20: Is there a case where you'd like both `default_settings` and `transmission_settings` to be read and then if theres anything in the `transmission_settings` file then those should overwrite the default settings?

transmission/updateSettings.py
```diff
@@ -0,0 +1,37 @@
+import os
+import sys
+import json
+
+
+# Verify script arguments
```
thomasborgen, line 6: `argparse` does this quite nicely. Gives you type checking, nice error messages and nicely formatted `help`

openvpn/persistEnvironment.py
```diff
@@ -0,0 +1,30 @@
+import os
+import sys
+import json
+
+
+# Verify script arguments
```
thomasborgen, line 6: argparse does this quite nicely. Gives you type checking, nice error messages and nicely formatted help

openvpn/persistEnvironment.py
```diff
@@ -0,0 +1,30 @@
...
+if len(sys.argv) != 2:
+    sys.exit(
+        'Invalid number of arguments. Usage:\n persistEnvironment.py /path/to/varibles-script.sh')
+
+envVarsScriptFile = sys.argv[1]
+
+wantedVariables = ['OPENVPN_PROVIDER', 'ENABLE_UFW', 'PUID', 'PGID', 'DROP_DEFAULT_ROUTE', 'GLOBAL_APPLY_PERMISSIONS', 'DOCKER_LOG']
```
thomasborgen, line 13: `frozenset`(immutable) or `set`s(unique) are nice for these kind of values.

openvpn/persistEnvironment.py
```diff
@@ -0,0 +1,30 @@
...
+
+variablesToPersist = {}
+
+for variable in os.environ:
+    if variable.startswith('TRANSMISSION_'):
+        variablesToPersist[variable] = os.environ.get(variable)
+    if variable.startswith('WEBPROXY_'):
```
thomasborgen, line 20: i'd probably make these `elif`s since all of these are mutually exclusive

PR #1324, opened 2020-08-09
Feature/persistent settings python

thomasborgen:
@haugene here the edits I would do.

If you are considering using more python I would also introduce a `pyproject.toml` to be used with [poetry](https://python-poetry.org/) for dependency management (yes just lots of dev linters! :D). This would also force that whatever py version you want to use would be used while developing.

As a last point I would also rename the files to use snake case instead of underscore.

You also have to decide what to do if an env var is set without a value or if the value is None. or empty string.

### DIFF
--- a/openvpn/persistEnvironment.py
+++ b/openvpn/persistEnvironment.py
@@ -1,30 +1,44 @@
+import argparse
 import os
-import sys
-import json
 
+parser = argparse.ArgumentParser(
+    description='Store env vars into runnable .sh file with export statements',
+)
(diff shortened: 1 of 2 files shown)


haugene:
Thanks a lot @thomasborgen, looks much better 👍 We'll see about the pyproject.toml at a later time.
About the "env var i set without a value" case. Thanks for the heads up, but I will ignore it for now and let it fail where it fails in that case. Setting it to an empty string should be legal, as there are some options where that is a valid setting.

PR #1339, opened 2020-08-23
Adding support for fetching configs from URL

haugene:
This PR restructures the start.sh script a bit. Separating the provider specific scripts from the main script and making it more "pluggable" for new provider specific logic.

Main feature is to accept a URL for config files. This is an alternative to the "manual provider" way of using your own config without coding. Hopefully this will make life easier for those that find it a bit confusing with the volume mounts.

I will update the documentation and enhance this feature a bit more as it will also be an alternative to the frequent provider config updates if more people use it.


haugene:
Been doing some testing so I'll merge it now. Any comments "after the fact" are also welcomed. Will bundle the changes up in dev and then take another look before going to master.

PR #1351, opened 2020-09-02
Update ARM64 image to use Buster as its base image

ndanyluk:
While investigating what seemed like a memory leak in the ARM64 docker image, I realized that the ARM64 docker image still uses Debian Stretch as its base image (limiting it to Transmission 2.92).  This PR corrects that by adjusting the CircleCI build instructions to build against Debian Buster

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,7 +118,7 @@ jobs:
     environment:
       IMAGE_NAME: haugene/transmission-openvpn
     docker:
-      - image: circleci/buildpack-deps:stretch
+      - image: circleci/buildpack-deps:buster
     steps:
       - checkout
       - setup_remote_docker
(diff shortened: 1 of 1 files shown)


haugene:
Hey @ndanyluk, sorry for the late reply. This looks good to me and thank you for pointing this out. I guess that means that we should also build the armhf image with buster as base image? https://github.com/haugene/docker-transmission-openvpn/blob/master/Dockerfile.armhf#L1

I don't think that the base image used by CircleCI should matter here, but it doesn't hurt so I'll merge it as is. Then I'll add the buster tag to armhf as well. I'm changing the target branch to `dev` so you can pull and test using the dev image after this.
Have been away from this project for a little while but I'm planning to push the new 3.0 release to master soon now.

Issue #1306, opened 2020-07-30
OpenVpn Connection Lost due to failed resolving host 


RickyXwang:
**Describe the problem**

The Vpn will timeout due to inactivity and tried to restart but will eventually fail in the loop of can not resolve the hostname to the VPN address.
It will recover after a manual restart of the docker container without having to make any other changes. The internet is online all the time with other services/container have no issue.
**Add your docker run command**

```
sudo docker run -d \
    --cap-add=NET_ADMIN \
    --device=/dev/net/tun \
    -v /volume1/networkconfig/resolv.conf:/etc/resolv.conf \
    -v /volume1/docker/transmission/data/:/data \
	-v /volume1/docker/transmission/config/:/etc/openvpn/custom/ \
	-v /volume1/hdhome:/hdhome \
	-e "TRANSMISSION_WEB_UI=transmission-web-control" \
    -e "OPENVPN_PROVIDER=CUSTOM" \
    -e "OPENVPN_USERNAME=*************" \
    -e "OPENVPN_PASSWORD=*************" \
    -e "LOCAL_NETWORK=192.168.0.0/24" \
    -e "OPENVPN_OPTS=--inactive 3600 --ping 10 --ping-exit 60" \
    -e "PGID=100" \
    -e "PUID=1028" \
    -p 9091:9091 \
    --sysctl net.ipv6.conf.all.disable_ipv6=0 \
    --name "transmission-openvpn-syno" \
    haugene/transmission-openvpn:latest
```
The command is modified from https://haugene.github.io/docker-transmission-openvpn/synology-nas/

**Logs**

The full logs are too long to be put in the comment, please see the logs file upload.
[trans.logs.txt](https://github.com/haugene/docker-transmission-openvpn/files/5000749/trans.logs.txt)
Here are some hight lights from the logs:
```
Wed Jul 22 02:41:39 2020 SIGUSR1[soft,ping-restart] received, process restarting
Wed Jul 22 02:41:39 2020 Restart pause, 5 second(s)
Wed Jul 22 02:41:44 2020 WARNING: --ns-cert-type is DEPRECATED.  Use --remote-cert-tls instead.
Wed Jul 22 02:41:44 2020 NOTE: the current --script-security setting may allow this configuration to call user-defined scripts
Wed Jul 22 02:41:44 2020 TCP/UDP: Preserving recently used remote address: [AF_INET]45.41.180.33:1195
[...]

PR #1373, opened 2020-09-21
Update HealthCheck for auto-recover TUN down

RickyXwang:
I added some code to try recovering from a hung TUN by resetting it.
See issue #1306


pkishino:
I like this, @haugene which branch should this go into?not master I think

fongd:
I would propose sending a `SIGINT` rather than `SIGUSR1` signal to OpenVPN. In my tests, `SIGINT` makes the container recover network connectivity faster and more reliably than a `SIGUSR1`.

RickyXwang:
Thanks for the feedback. I will do some more testing on that.

PR #1379, opened 2020-09-27
Add LOG_TO_STDOUT environment variable + documentation

kushagharahi:
The environment variable `DOCKER_LOG`, when set to `true` already logs Transmission logs to stdout. In my opinion this is a poor name, so I added a new variable `LOG_TO_STDOUT` and added documentation for it. I've left `DOCKER_LOG` in to ensure backwards compatibility. 

Regarding 3.x: I can open a second PR once this gets merged to remove `DOCKER_LOG` in the 3.x branch.

If the reviewers do not like the depreciation of `DOCKER_LOG` and introduction of `LOG_TO_STDOUT` then I can update this PR to just be documentation of the environment variable `DOCKER_LOG`.

PR #1380, opened 2020-09-29
Remove DOCKER_LOG in favor of LOG_TO_STDOUT

kushagharahi:
Context https://github.com/haugene/docker-transmission-openvpn/pull/1379


pkishino:
Can you please rebase this to remove the other health check commit ?

haugene:
I think this might also be due to master being commits before dev. I'll create a PR that merges master into dev and the result of that is probably what @kushagharahi should rebase on top of. I think we can keep most of the changes from master but since one Dockerfile (for alpine) is removed on dev and we'll deprecate DOCKER_LOG the last fixups has to be done there.

kushagharahi:
ok @haugene and @pkishino I've rebased off of the latest dev

Dockerfile
```diff
@@ -46,7 +46,7 @@ ENV OPENVPN_USERNAME=**None** \
     WEBPROXY_PORT=8888 \
     WEBPROXY_USERNAME= \
     WEBPROXY_PASSWORD= \
-    DOCKER_LOG=false \
+    LOG_TO_STDOUT=false \
```
kushagharahi, line 49: https://github.com/haugene/docker-transmission-openvpn/blob/master/Dockerfile#L140

I think this was removed in merge conflicts

haugene:
Looks good :+1: Thanks @kushagharahi

PR #1396, opened 2020-10-10
Merge master into dev

haugene:
Ref discussion in #1380. In order to deprecate DOCKER_LOG and introduce LOG_TO_STDOUT we will have to merge it on top of the master branch at some point and I think we might as well do it now.

This PR merges changes from master into dev. There are 4 commits to master that are not on dev. 3 of them has no conflicts, but the DOCKER_LOG/LOG_TO_STDOUT has conflicts as expected.

The conflicts are as follows:
```
	both modified:   Dockerfile
	deleted by us:   Dockerfile.alpine
	deleted by us:   transmission/environment-variables.tmpl
``` 

In this PR I've resolved the conflicts by just staying on the `dev` version for all the conflicting files. Two of them are deleted so that should be a no-brainer but we're then lacking all changes that should be in the Dockerfile. And that's where @kushagharahi comes in.

Sounds good? If so we'll merge this @pkishino @kushagharahi


pkishino:
sure, looks fine

PR #1406, opened 2020-10-19
VyprVPN Config Auto-Fetcher

jsloan117:
This script automatically pulls down the configs and renames them just like they are currently named in master. It will only download them (configs) if `/etc/openvpn/vyprvpn` is empty, trying not to download the zip file every time and hammering the provider's site.


haugene:
Looks great and seems to work like a charm when testing locally :+1:  I get to `AUTH_FAILED` but that's as expected as I don't have an account. Thanks for helping to clean up this image and make it more future proof :smile:

PR #1408, opened 2020-10-20
Update update-port.sh

mizzi0n:
a change to allow us to keep using port forwarding


haugene:
Great stuff, thanks @mizzi0n :tada: 
Let's follow up on this in #1334 and see the response from others testing it.

PR #1413, opened 2020-10-23
Coerce env var values to expected types by settings.json

tallzoologytortbatik:
@haugene 

This addresses a bug introduced by moving from the old settings templating mechanism to using updateSettings.py. 

Previously the settings template controlled the output type when env var values were written to settings.json by having some values quoted and not others. After the move to updateSettings.py that control was lost and all were written as strings. Subsequently Transmission would pick those up and reset them when there's a type mismatch. E.g. a string given instead of an int for max peers "100" would get reset to 0. No torrents would actually download after that. With this change we look at the baseline settings for the expected setting type and coerce the value of the corresponding env var to that type. Tested by building locally and running.

### DIFF
--- a/transmission/updateSettings.py
+++ b/transmission/updateSettings.py
@@ -57,6 +57,24 @@ for setting in settings_dict:
     setting_env_name = setting_as_env(setting)
     if setting_env_name in os.environ:
         env_value = os.environ.get(setting_env_name)
+
+        # Coerce env var values to the expected type in settings.json
+        if type(settings_dict[setting]) == bool:
+            env_value = env_value.lower() in ['True', 'true']
+        else:
+            setting_type = type(settings_dict[setting])
+            try:
+                env_value = setting_type(env_value)
+            except ValueError:
+                print(
+                    'Could not coerce {setting_env_name} value {env_value} to expected type {setting_type}'.format(
+                    setting_env_name=setting_env_name,
+                    env_value=env_value,
+                    setting_type=setting_type,
+                    ),
+                )
+                raise
+
         print(
             'Overriding {setting} because {env_name} is set to {value}'.format(
                 setting=setting,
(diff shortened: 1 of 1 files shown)


haugene:
Looks great @tallzoologytortbatik, thanks for helping out! :+1:
I'll merge it now so we can squash that bug for now and test that change. But I just wanted to ask what you think about not raising the error if we get `ValueError` but instead just log it as a warning and treat it as a string? That way Transmission can try to resolve it and it might reset it or it might not. Or is it better just to fail hard and let us handle that corner case when it occurs?

tallzoologytortbatik:
@haugene   I was debating that myself and here's the reasoning I went through. I already know that for multiple integer settings I've tried Transmission resets then to 0, so I'd speculate that all integer settings would reset to 0 if given strings. This error shouldn't occur for string settings and booleans are handled separately, so we only care for the integer case. If we don't raise an error that stops execution, Transmission would run but incorrectly as it won't respect what the user set in their environment variables. The user might not find out why Transmission isn't functioning correctly until they notice something isn't right and look at the log and find this particular message among others. Alternatively Transmission wouldn't run right from the start, which would prompt the user to look at the log and they'll see the last message which would be this failure. I think the latter is easier to discover and fix. In the single data point I have - myself - it definitely took a while before I focused my investigation on settings being applied incorrectly. My first thought was something being wrong with VPN config so I spent significant time digging into that first.

haugene:
Sounds like a good plan. I generally like the "fail fast" approach as opposed to letting things start with invisible or hard to catch errors hidden in some log somewhere. Let's keep it as it is until some other consideration comes along forcing us to re-evaluate :smile: :+1:

haugene:
Re this discussion. New case in #1450, feeling up for providing a fix? 😊

PR #1425, opened 2020-10-27
Switch to nextgen PIA servers

vtavernier:
Official announcement: https://www.privateinternetaccess.com/blog/private-internet-access-legacy-vpn-network-sunset-announcement-30-september

PIA started retiring their "legacy" VPN infrastructure. As far as I could test, the legacy servers stopped working today (or at the very least, *recently*). I patched the running version of the image I had with the "NextGen" version of their OpenVPN configuration files, and it seems to work again.

**Note: they now use the account's username and password instead of the specific PPTP/L2TP username and password**

### DIFF
--- a/openvpn/pia/updateConfigs.sh
+++ b/openvpn/pia/updateConfigs.sh
@@ -9,7 +9,7 @@ cd "${0%/*}"
 find . ! -name '*.sh' -delete
 
 baseURL="https://www.privateinternetaccess.com/openvpn/openvpn"
-extension=".zip"
+extension="-nextgen.zip"
 declare -a configsURLs=(    "" "-strong" "-tcp" "-strong-tcp" "-ip")
 declare -a configsFolders=( "" "strong"  "tcp"  "tcp-strong"  "ip")
 


vtavernier:
This seems to break the port forwarding script though, I'll update the PR with more info if PIA decides to publish some.

markchaik:
Currently on "latest" docker image (not dev) and as of today, can't connect to several PIA servers.  I suspect it's related to their decommissioning of several servers.  Will this make its way into v2, and into latest docker image?

zenxedo:
Had to switch over to dev docker image today because latest is not working anymore for PIA.

edwardreed81:
In addition to updating the openvpn configs, the port forwarding mechanism needs to be changed.

https://github.com/pia-foss/manual-connections/blob/master/port_forwarding.sh

That's the new script from PIA that seems to do the magic, but I'm not smart enough to implement that in this container.

haugene:
Hey guys. I think this conversation should be merged with #1334.

As for the changes proposed here. On the dev-branch the `updateConfigs.sh` script should actually be deprecated and replaced by `configure-openvpn.sh` that is an attempt to standardize how different providers are initialized on container startup.

Thanks for the links to the official scripts. I think our current script for nextgen port forwarding is loosely based on it, but I will check later if we can adopt it more directly. We're currently using a migration of what we had before and this https://github.com/fm407/PIA-NextGen-PortForwarding/blob/master/pia-pfSense.sh

PR #1433, opened 2020-10-29
Mullvad has new configs with new servers

XenGi:
I'm mostly using this but but all profiles need to be updated I guess.


AKDub:
Presumably this is why I now get "Supplied config mullvad_gb_mnc.ovpn could not be found." and the container wont start?

garret:
I also experienced this. Hopefully it will be included soon as I imagine Mullvad is among one of the most used VPNs.

haugene:
There is no place to get a complete list of Mullvad servers that is publicly available, right? Then we could make a script for this as with the other providers. Or a .zip file containing all of the configs?

Mullvad has almost 300 servers and I feel it's not a solution to update them one by one?

garret:
@haugene It seems you can download a .zip containing the configs but you need to have a login access ( https://mullvad.net/en/account/#/openvpn-config/?platform=linux ). I can try to help if I receive some guidance since I am not really an expert. 
Screenshot_2020-11-10  Mullvad VPN(1)

haugene:
Yep, that's what we need :+1: Seems you have to select countries/regions, can you select all? And then download the zip file. Default UDP is fine.

garret:
@haugene I attached the zip archive after selecting "All countries". I have removed my `mullvad_userpass.txt` file but left the `mullvad_ca.crt` and `update-resolv-conf`. Sorry could you confirm I did not attach any sensible information? 

[mullvad_config_linux.zip](https://github.com/haugene/docker-transmission-openvpn/files/5520351/mullvad_config_linux.zip)

haugene:
That looks good @garret, no sensitive information. I've now updated the image with the newest config and you should be able to test it if you pull and run the :dev tag.
Looking at my changes I see that I deleted many more files than I added here. I think the last update of mullvad was more thorough. Adding all different configs (udp53, tcp80 and tcp443) for each country. I'm not sure, but here it seems that all servers for a given country is included in a _all config. Might be good enough?

garret:
@haugene I tried de :dev tag and it does not work because I get the same error due to `tun-ipv6` present in all the config files. I wrote earlier https://github.com/haugene/docker-transmission-openvpn/pull/1433#issuecomment-724579521 how I solved that issue. Please let me know if I can help more with testing.

XenGi:
The `tun-ipv6` is only important if you want to have IPv6 traffic routed through the VPN. If you only have IPv4 in the docker container it should be fine to leave it out. Also I guess the default UDP config should be fine. Only people in restricted networks need the other configs.

XenGi:
I'm updating the pull request with all other variants.

XenGi:
Now it should be complete.

XenGi:
Forgot to fix the paths for the credentials and ca file. Let me know if I should change anything else. The last bit that is in there that wasn't before is the up/down scripts. Should I remove them?

haugene:
Hey. Thanks for fixing this @XenGi. You need to rebase with dev though. I've pushed changes and you have conflicts.
The up/down options can stay I guess. But they won't have any effect as we override them to start/stop Transmission. I've commented them out in my change - as a reminder that it can be hooked in at a later point.

garret:
FIY I don't know if it was updated in the end but I just re-downloaded the `:dev` and still have the `[..] dev tun0 RTNETLINK answers: Permission denied [..]` error by default.

XenGi:
I'll update the pull-request and base it to `dev`.

XenGi:
omg.. I have no idea what I did, but now everything seems to be broken.

haugene:
Yes. May be just as good to just create a new PR. Rebasing is a powerful tool, and can be easy enough. But if you're not very familiar with it then one error really send you on your way. Been there, many times :laughing:

XenGi:
I created a proper pull request. Hopefully correct this time. :D
https://github.com/haugene/docker-transmission-openvpn/pull/1509

Issue #454, opened 2018-03-16
Container restart on connection failure


haugene:
There has been some discussions on the container not restarting when the connection drops. Creating this issue to see how many are affected and find a solution.

In #438 it was pointed out that the openvpn server can override settings at the client if `pull` option is enabled. So all configs that have this option should maybe be updated?

Or we could write a separate checker that kills openvpn when it looses connection or dns?

himred:
I confirm, here when the remote server dies, the container does not restart.
I use custom healthcheck and the hcd daemon to restart it.

richardtheb:
I seem to be having this issue. The Docker container keeps running, and I can access the GUI locally as normal. But all of the trackers have failed connections. It seems to happen after I have left it running for some time. The only way I can get around it is to stop the docker container and restart it. This is with PureVPN. Not sure what else I can do to pin this one down, but ask if you need log files, etc.

AlexWootton:
I'm seeing this issue as well. Running v2.0 of this Docker image with Windscribe as the VPN provider. It runs fine for a few days and then OpenVPN fails reporting `RESOLVE: Cannot resolve host address: uk.windscribe.com:1194 (Temporary failure in name resolution)`. This is an error that I have faced when running OpenVPN locally so the easiest solution may be to restart the container if the VPN tunnel drops.

I am running the container with the environment variable `-e "OPENVPN_OPTS=--inactive 3600 --ping 10 --ping-exit 60"` but the container is not killed when the VPN tunnel drops.

Are there any useful logs that I can provide the next time this issue occurs?

mbarany:
Yes, this would be a nice addition. I currently have a cron on the host that just restarts the container on a daily basis, but that's more of a hack.

IBNobody:
I am running the docker on a Synology NAS, and I use Mullvad.  The `OPENVPN_OPTS` with the pull-filter is not restarting (same SIGUSR1 issue that kicked off this initiative). Having a second process as a kill switch would be ideal.

For reference, this is what I was setting `OPENVPN_OPTS` to...

` -e "OPENVPN_OPTS=--inactive 3600 --ping 10 --ping-exit 60 --pull-filter ignore \"ping\"" \`

coolboiime:
I'm also getting a ton of cannot resolve host addresses just like @AlexWootton. My old script used to just check if the port forwarding was communicating, if not, it would restart everything.

coolboiime:
[...]

PR #1459, opened 2020-11-04
Update healthcheck script

RickyXwang:
See #1306 
Update
1. Change ping -c 1 to -c 5, should avoid the issue with unstable network or light packet loss.
2. Move interface check to the front
3. Added date-time output to better log
4. If SIGUSR1 did not work use HUP
5. Added logic to check if the ping used tun dev


RickyXwang:
We can hold on to this for now. Making an update to address issues with multiple tun dev fond and DNS return ipv6 address instead of ipv4.

haugene:
Hey @RickyXwang. Thanks for addressing this issue, and for all your help on answering issues. Greatly appreciated!

As for this change. It looks good and is probably more stable than today's script so I'm tempted to just merge it. I'm then assuming that ping will give a 0 exit code as long as at least one response is given?

Another thing though. I was actually quite surprised to see that the health check script was killing openvpn. I must have forgotten that was there. I think it feels a bit wrong. A health check script should check the health and not do this kind of intrusive action, what do you think?

haugene:
Regarding my last comment about health check script being non-intrusive. Have a look at the last comments about using autoheal companion container in #454. What do you think about introducing this as the "official" solution to the problem for now?

RickyXwang:
I do agree that healthcheck should be just a health check, will move the restart out somewhere. For the autoheal container, it sounds like a good solution be seems like to me it hasn't been updated for a while (the latest version from 2 years ago). It also has some issues working with docker-compose as the label doesn't work in compose (I think quite a lot of folks do use compose here). I will keep looking and see if we got any other options or a more maintained project for this.

haugene:
Great! :+1: I added the label to the project so they should be there by default. It was also reported in the other issue that adding them locally was a struggle. If we end up proposing another companion-service here then we can adapt to it instead.

haugene:
Hey again. I just reverted the last change to the healthcheck script. I've been seeing the log message `event_wait : Interrupted system call (code=4)` in multiple issues and I'm suspecting that it's the health check script that is interrupting openvpn.
The container might not function anyways, as the health check reports it as being broken - but wanted to try and see if it resolved anything. So you'll have to rebase against that as well. But I'm all for a better health check script, so ping me when you feel this is ready to be merged.

RickyXwang:
I think, for now, I can make this new version of `healthcheck` to a recovery script and have it run separately, set a flag to enable it from the start script. 
We are facing lots of people having issues because
1. They are new and start to set things up
2. They changed some config that is not supposed to be changed or using the old configuration in the new image.
3. Some updates (We made or the VPN server-side change) broke the previous working configure and now require some change.

And If the script is trying to restart the OpenVPN because it failed to ping, it sometimes covered up why it is failing because all they see is it keeps trying to restart.
This recovery script should be set to disable by default and should only be enabled manually after the container is running stable.
This approach should also take care of the server push overwriting the docker ping-restart issue.

What do you think?

pkishino:
new PR merged #1805 that covers recovery etc.. is this one still needed/useful?

RickyXwang:
Hi @pkishino, I took a quick look at #1805, it appears to me that the recovery is implemented as if it failed ping after certain tries it will `kill 1` which will restart the container (Correct me if I am wrong). Which is not what I was trying to do. What I wanted is a way to recover without restart the container first, and only when that failed we restart the container. So that it can avoid having to reload everything from the hard drive to the memory, as well as keeping a better record of the seeding status. In case the Tracker went down for a period of time, I would like to keep the transmission running if possible so when the Tracker went back up it has a consistent record.
I am still running my version of healthcheck in my container and it's working great so far. I got busy from work and haven't had time to follow up on the dev. So if someone wants to pick up my change they are welcome to do so. Or if you still want my change I may be able to merge it when I have more time (two months later maybe :<) If not you can just close this for good.

PR #1469, opened 2020-11-06
Added Filters to ignore ipv6

Rainee4563:
Filters to ignore ipv6 to fix rnetlink permission denied

### DIFF
--- a/openvpn/mullvad/mullvad_us-ca_tcp443.ovpn
+++ b/openvpn/mullvad/mullvad_us-ca_tcp443.ovpn
@@ -41,10 +41,12 @@ ca /etc/openvpn/mullvad/ca.crt
 
 
 
-tun-ipv6
+
 script-security 2
 
-
+pull-filter ignore "dhcp-option DNS6"
+pull-filter ignore "tun-ipv6"
+pull-filter ignore "ifconfig-ipv6"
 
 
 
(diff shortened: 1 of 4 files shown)


haugene:
Allright, thanks for the hint. But Mullvad has ~280 servers so we need to update them all in that case.
I'll get back to this. Have to look more into how we should handle IPv6 in this image. It has been a growing topic as IPv6 gets more wide spread, but there is still no uniform policy in this image and we should get it. This is also a concern across all providers.

Rainee4563:
Didnt imtead to be mean in this just wanted to help... in anyway i could maybe just pushing this temp fix not sure

haugene:
No no. I'm sorry if I wasn't clear. It's great that you want to contribute and you're changes here are a hint at the larger solution :+1: 
I was just trying to say that I need to have a closer look at this and that I'll get back to it. It might be that we should add these lines to all the configs. The problem is just that if we don't add the routes, and there are IPv6 packets - will they then not go through the VPN tunnel and instead use the eth0 interface and your home LAN?

We need to make sure that we handle this on all levels and I just need to read up a bit on that. Right now I'm burning through small issues and trying to clear the way to work on some more cross-cutting concerns as this is an example of. So I'm just saying that I'll get back to it later :)

Rainee4563:
ahh okay then! just don't burn yourself out ok!

pkishino:
closing as incomplete

PR #1471, opened 2020-11-06
Prevent logging rpc-password to stdout

utkuozdemir:
This PR prevents the logging of the rpc-password, which is a sensitive information, to the stdout.
Before:
```
Generating settings.json for Transmission from environment and defaults /etc/transmission/default-settings.json
...
Overriding incomplete-dir because TRANSMISSION_INCOMPLETE_DIR is set to /data/incomplete
Overriding rpc-password because TRANSMISSION_RPC_PASSWORD is set to SuperSecretPassword1
Overriding rpc-port because TRANSMISSION_RPC_PORT is set to 9092
...
````

After:
```
Generating settings.json for Transmission from environment and defaults /etc/transmission/default-settings.json
...
Overriding incomplete-dir because TRANSMISSION_INCOMPLETE_DIR is set to /data/incomplete
Overriding rpc-password because TRANSMISSION_RPC_PASSWORD is set to [REDACTED]
Overriding rpc-port because TRANSMISSION_RPC_PORT is set to 9092
...
````

### DIFF
--- a/transmission/updateSettings.py
+++ b/transmission/updateSettings.py
@@ -54,9 +54,11 @@ def setting_as_env(setting: str) -> str:
 
 # For each setting, check if an environment variable is set to override it
 for setting in settings_dict:
+    setting_is_sensitive = setting == "rpc-password"
     setting_env_name = setting_as_env(setting)
     if setting_env_name in os.environ:
         env_value = os.environ.get(setting_env_name)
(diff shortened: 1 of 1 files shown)


haugene:
Agreed :+1: Thank you for contributing! :raised_hands:

PR #1482, opened 2020-11-08
Changes to make systemd docs work with 3.0

MattWThomas:
CREATE_TUN_DEVICE is now set to true by default and spaces in OPENVPN_CONFIG are no longer escaped.


haugene:
Thanks :+1:

PR #1488, opened 2020-11-09
Update documentation for timezone configuration option.

ksurl:
I was using the bind mount method for /etc/localtime as read only and was wondering if there was support for an environment variable instead. A few closed issues showed it is already there, but nothing on the documentation page for arguments mentions it.


haugene:
Great, thank you for adding that :+1: It is supported as you say - but has not been part of the docs.

PR #1490, opened 2020-11-10
set timezone at startup

ksurl:
Setting the TZ environment variable is not respected by transmission log timestamps. It does show the correct time when running `date` interactively in the container, but logs still show UTC timestamps. Various issues have been opened for this and the answer has been to bindmount the host /etc/localtime to force it work. 

Since tzdata is already installed in the image, I hope this is a more elegant solution: copying the timezone data to the correct path before transmission starts. If possible, it'd be ideal to place this at the start of the container so any ufw, openvpn or other logs will also show the correct timestamp, but I was unable to trace where /etc/transmission/start.sh is called from /etc/openvpn/start.sh so this a start.

### DIFF
--- a/openvpn/persistEnvironment.py
+++ b/openvpn/persistEnvironment.py
@@ -21,6 +21,7 @@ wanted_variables = {
     'GLOBAL_APPLY_PERMISSIONS',
     'LOG_TO_STDOUT',
     'DISABLE_PORT_UPDATER',
+    'TZ'
 }
 
 variables_to_persist = {}
--- a/transmission/start.sh
+++ b/transmission/start.sh
@@ -69,10 +69,6 @@ else
   LOGFILE=${TRANSMISSION_HOME}/transmission.log
 fi
 
-if [ ! -z ${TZ} ]; then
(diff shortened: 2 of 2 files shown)


ksurl:
It seems it is not working. Still seeing UTC without mounting /etc/localtime. Any insight on when it takes effect would be helpful

haugene:
Great initiative :+1: 
Transmission is started by the --up option to OpenVPN. We set it here: https://github.com/haugene/docker-transmission-openvpn/blob/master/openvpn/start.sh#L124
It is then the tunnelUp.sh that calls transmission/start.sh

Because of this handover these are not the same shell and OpenVPN doesn't automatically set all environment variables it's running with to its child processes. It's been a long time since I tackled that problem so I'm not sure if we could make it do it
But we've handled it by persisting a selected set of environment variables to a shell script and then sourcing it wherever the environment is needed.

So you would have to add TZ to that list as well. https://github.com/haugene/docker-transmission-openvpn/blob/master/openvpn/persistEnvironment.py#L15

ksurl:
seems to work now on a local build of the image 👍

haugene:
Great! Thanks for contributing :raised_hands:

PR #1491, opened 2020-11-10
Add Njalla VPN configuration

ianw:
This is for the VPN service provided by https://njal.la/


haugene:
This is the config exactly as it comes from Njal.la? I'm getting some errors when running it locally.
```
Linux ip -6 addr add failed: external program exited with error status: 2
```

Not sure why this happens here and not other providers

ianw:
@haugene Yes, that is just what I downloaded from them.  I also saw that; and had to use the --sysctl flag (like noted in https://github.com/haugene/docker-transmission-openvpn/tree/master/docs#make-it-work-on-synology-nas) to disable ipv6 as seems to be noted in a few other places, then it connected.  Is there a better way to do that?

haugene:
Ah, there we go. It works indeed. But it works **too well**. I get connected without providing any username/password. Njalla uses a certificate auth so the configs are personal. We've seen this before and it's a blocker when it comes to adding it as a provider here. You should invalidate this config and regenerate keys as soon as possible.

What you can do though, to use this container, is to either mount the file locally on your machine. Follow the instructions in the documentation about custom providers. Or you can upload the file to Dropbox or Google Drive and create a download link from there that you can add to `OPENVPN_CONFIG_URL` and the container will download it on start and use it. The config would then be available to anyone with a link, but the links are basically a hash and you would need to assess the security there.

About the sysctl stuff I'm not sure. I think so, but I also think I'll learn more about possible options soon as IPv6 is a repeating topic here - and getting more frequent year by year.

ianw:
@haugene oops; thanks.  I had a look and thought it looked the same as the other config files, but clearly my human ovpn parsing skills are lacking :)

haugene:
No worries :smile: Closing this then?

Issue #1362, opened 2020-09-12
/scripts/transmission-pre-stop.sh runs after tunnel down


YujiShen:
**Describe the problem**

I would like to paused all my torrents when I kill the container, because some trackers will not allow you to download from multiple IP addresses. If torrents are not paused correctly, the tracker will still have previous records and I have to wait several hours for it to clear.

I put `transmission-remote --torrent all --stop` in /scripts/transmission-pre-stop.sh, but it does not work. First, I suspect the torrents are not really paused after the command. So I wrote a more complex script to make sure all torrents state are "Stopped", I even add sleep after that. But it does not work when I kill the container (tracker still has record). It does work when I call the script directly in the container. 

So I think the /scripts/transmission-pre-stop.sh is running after VPN tunnel is down. Is there a way to run script before the tunnel is down (and before stop transmission)?

RickyXwang:
Hi, I might want something similar as well. Have a couple of questions for you:
Assuming your script works when running manually to have all torrent stopped. 
What triggers your script to run? Also, in which way you are killing the container?

YujiShen:
@RickyXwang Thanks for attention.
1. I went inside container and run the script `/scripts/transmission-pre-stop.sh`.
2. I kill container with docker-compose stop or docker-compose down.

haugene:
Well, sounds reasonable. We need an event to react to though. From what I see in the OpenVPN manual it might look like `down-pre` is an option we might want to try in that case. https://community.openvpn.net/openvpn/wiki/Openvpn24ManPage

clement-z:
I think the `--route-pre-down` option is more suited than `down-pre`:

```
--route-pre-down cmd
    Run command cmd before routes are removed upon disconnection. 
```

It is the behavior I would have expected from the container (and I think other users as well).

An easy fix is to change `--down tunnelDown.sh` to `--route-pre-down tunnelDown.sh` in [openvpn/start.sh#L124](https://github.com/haugene/docker-transmission-openvpn/blob/93d0141d928ec661411e2c6f0e079aeff2c018dd/openvpn/start.sh#L124)

I doubt this could be considered a breaking change, but 🤷

haugene:
I agree @clement-z. I think this will both solve @YujiShen's issue and be more in line with what to expect from the container.
Transmission should be stopped at the first sign of a shutdown coming.

[...]

PR #1502, opened 2020-11-12
Initial version of route-pre-down script

RickyXwang:
See enhancement request #1362 

Try to gracefully stop transmission when stopping the container.

To enable: add `--route-pre-down /etc/scripts/route-pre-down.sh` to `OPENVPN_OPTS`

According to https://wiki.archlinux.org/index.php/transmission#Transmission_daemon_and_CLI
`pkill -3 transmission-daemon` or `transmission-remote --exit` is the proper way to stop the transmission-daemon, however, during my test `pkill -3` does not work. I was only able to get the exit log with `transmission-remote --exit`

### DIFF
--- a/scripts/route-pre-down.sh
+++ b/scripts/route-pre-down.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+# redirect stdout/stderr to a file
+#exec &>>route-pre-down.log
+
+#Print Date
+NOW=$(date +"%Y-%m-%d %T")
+
+echo "${NOW}: route-pre-down script: Start "
+
+echo "Sending exit signal to transmission."
+
+transmission-remote --exit &
+
+wait
+
+echo "route-pre-down script: Done"
+
+
+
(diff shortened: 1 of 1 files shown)


scripts/route-pre-down.sh
```diff
@@ -0,0 +1,19 @@
...
+#exec &>>route-pre-down.log
+
+#Print Date
+NOW=$(date +"%Y-%m-%d %T")
+
+echo "${NOW}: route-pre-down script: Start "
+
+echo "Sending exit signal to transmission."
+
+transmission-remote --exit &
```
clement-z, line 12: This will fail if authentication is needed. You can reproduce the following steps to make it handle transmission being protected:

https://github.com/haugene/docker-transmission-openvpn/blob/93d0141d928ec661411e2c6f0e079aeff2c018dd/openvpn/pia/update-port.sh#L10-L13
https://github.com/haugene/docker-transmission-openvpn/blob/93d0141d928ec661411e2c6f0e079aeff2c018dd/openvpn/pia/update-port.sh#L100

Then `transmission-remote $myauth --exit`

RickyXwang: Thanks for pointing this out, I will make the change later.

pkishino: thinking to update with this change soon?

pkishino (changes requested):

pkishino (approved):

PR #1514, opened 2020-11-14
Add safeguard around transmission shutdown

### LINKED ISSUE #1513: It is not checked whether transmission actually exited
https://github.com/haugene/docker-transmission-openvpn/blob/93d0141d928ec661411e2c6f0e079aeff2c018dd/transmission/stop.sh#L11-L18

The `transmission/stop.sh` script does not actually check whether `transmission` has exited. This is probably a nitpick, but it would likely be better to safeguard it.

clement-z:
This is really a nitpicking PR, but it doesn't hurt IMO.

Commit 0ee67b1... sanitizes the `ps -p` call.
Commit e9e399c... ensures `transmission-daemon` is indeed killed after the 2 second timeout which was already present.

On my setup, exiting a normally running daemon is instant; I don't know what could make it hang for more than 2 seconds (I/O in the case of remote file systems transparently given to the container ?) but it might make sense to either increase it or to set it as an environment variable since it is killed using `kill -9`.

### DIFF
--- a/transmission/stop.sh
+++ b/transmission/stop.sh
@@ -12,16 +12,21 @@ echo "Sending kill signal (SIGTERM) to transmission-daemon"
 PID=$(pidof transmission-daemon)
 kill $PID
 
-# Give transmission-daemon time to shut down
-for i in {1..10}; do
+# Give transmission-daemon some time to shut down
+TRANSMISSION_TIMEOUT_SEC=10
+TIMEOUT_NITER=$((2 * $TRANSMISSION_TIMEOUT_SEC))
+for i in $(seq $TIMEOUT_NITER); do
     ps -p "$PID" &> /dev/null || break
-    sleep .2
+    [[ $i == 1 ]] && echo "Waiting ${TRANSMISSION_TIMEOUT_SEC}sec for transmission-daemon to die"
+    sleep .5
 done
 
 # Check whether transmission-daemon is still running
 if ! ps -p "$PID" &> /dev/null; then
     echo "Sending kill signal (SIGKILL) to transmission-daemon"
     kill -9 $PID
+else
+    echo "Successfuly closed transmission-daemon"
 fi
 
 # If transmission-post-stop.sh exists, run it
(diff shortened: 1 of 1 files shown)


clement-z:
fixes #1513

haugene:
Safeguards are good :+1: I can merge it as it stands, but do you want to increase the 2 second timeout?
Adding an environment variable is another options, but I don't think it will be a very much used one. And I'm generally not too fan of extracting config that "never changes". Flexibility is good though :man_shrugging: 
Could give it up to 5 seconds and say that it's kill -9 after that? Open for suggestions.

Thanks a lot for all your other contributions here the past time as well. Much appreciated!

clement-z:
I made it into a variable inside the script, but I don't think making it available as a default config option is needed. I follow your opinion: a variable which will (almost ?) never be modified should not pollute the parameter set.

If a longer timeout is needed, a user can either include its own shutdown script in `scripts/transmission-pre-stop.sh` (which can be the snippet from `transmission/stop.sh`) or rebuild the container after modifying the file. 
The first would just replicate the log messages + an invocation error for `kill` if `$PID` is empty.

I think both options are quite low effort, but I might be biased. 😄

clement-z:
I had inverted the `if` branch. Booted it up and it seems like it works OK now.

clement-z:
> a variable which will (almost ?) never be modified should not pollute the parameter set.

Although we could maybe do:

```bash
TRANSMISSION_TIMEOUT_SEC=${TRANSMISION_TIMEOUT_SEC:-10}
```

haugene:
> > a variable which will (almost ?) never be modified should not pollute the parameter set.
> 
> Although we could maybe do:
> 
> ```shell
> TRANSMISSION_TIMEOUT_SEC=${TRANSMISION_TIMEOUT_SEC:-10}
> ```

Yeah, why not. We don't have to add it to the documentation. But for those reading the script they'll see that they have a choice :+1: If you squash commits while you're at it that's great. If not, I don't mind too much :smile:

clement-z:
Ok I will rebase/squash and open a new PR.

Issue #833, opened 2019-06-11
No way to set folder permissions


aSaltyPirate:
**Is your feature request related to a problem? Please describe.**
My system requires the items downloaded and the folders they are in to be accessible to all members of the group. Currently one can set the umask but that only applies to the files inside the folders they're downloaded in and not the folders themselves.

**Describe the solution you'd like**
Either making the folders reflect the given umask in settings or simply allowing changing the permissions of the completed directory. As it is now the permissions of this directory are hardcoded to 640 and they revert to this on container restart.

evsasha:
Hi, I did not understand your request.

- What is your system?
- What parameters do you run the container with?
- What folders do you have not group permissions for?

I tested on CentOS and synology NAS and transmission user and group have full acces to data directories:

```
$ cd downloads/Anime/
$ ls -al
total 0
drwxrwx---+ 1 transmission transmission 1592 Jun  3 00:55 .
drwxrwx---+ 1 transmission transmission  432 Nov 12  2018 ..
drwxrwx---+ 1 transmission transmission 3268 Mar 15 11:43 XXX1
drwxrwx---+ 1 transmission transmission   12 Jun  1 21:16 XXX2
drwxrwx---+ 1 transmission transmission  562 Mar 15 12:02 XXX3
```

Show some examples please.

aSaltyPirate:
Huh, now it seems to be ignoring my umask variable completely.  I run with the following command 

```
docker run --name=transmission --cap-add=NET_ADMIN  -d \
	--label autoheal=true \
	-v /home/transmission/data:/data \
	-v /etc/localtime:/etc/localtime:ro \
	-v /home/transmission/vpns/AirVPN_America_UDP-443.ovpn:/etc/openvpn/custom/default.ovpn \
	-v /media/videos/downloads/completed:/completed \
	-v /media/videos/downloads/incomplete:/incomplete \
	-v /media/videos/downloads/dropbox:/dropbox \
	-e CREATE_TUN_DEVICE=true \
	-e TRANSMISSION_UMASK=2 \
	-e OPENVPN_PROVIDER=CUSTOM \
	-e TRANSMISSION_PEER_PORT=57014 \
	-e TRANSMISSION_RATIO_LIMIT_ENABLED=true \
	-e TRANSMISSION_RATIO_LIMIT=3 \
	-e TRANSMISSION_IDLE_SEEDING_LIMIT_ENABLED=true \
[...]

PR #1518, opened 2020-11-15
Apply Transmission umask setting to download directories

wriches:
Applies the `TRANSMISSION_UMASK` configuration to the download and incomplete directories. Solves the problem of hardcoded permissions in these directories.

The octal conversion requires changing the `userSetup.sh` script to be interpreted by `bash` instead of `sh`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,6 +38,7 @@ ENV OPENVPN_USERNAME=**None** \
     TRANSMISSION_DOWNLOAD_DIR=/data/completed \
     TRANSMISSION_INCOMPLETE_DIR=/data/incomplete \
     TRANSMISSION_WATCH_DIR=/data/watch \
+    TRANSMISSION_UMASK=2 \
     CREATE_TUN_DEVICE=true \
     ENABLE_UFW=false \
     UFW_ALLOW_GW_NET=false \
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,6 @@ ENV OPENVPN_USERNAME=**None** \
     TRANSMISSION_DOWNLOAD_DIR=/data/completed \
     TRANSMISSION_INCOMPLETE_DIR=/data/incomplete \
     TRANSMISSION_WATCH_DIR=/data/watch \
-    TRANSMISSION_UMASK=2 \
     CREATE_TUN_DEVICE=true \
     ENABLE_UFW=false \
     UFW_ALLOW_GW_NET=false \
--- a/transmission/userSetup.sh
+++ b/transmission/userSetup.sh
@@ -42,7 +42,7 @@ if [ -n "$PUID" ] && [ ! "$(id -u root)" -eq "$PUID" ]; then
             ${TRANSMISSION_WATCH_DIR}
 
         echo "Setting permissions for download and incomplete directories"
-        TRANSMISSION_UMASK_OCTAL=$(printf '%03g' $(printf '%o\n' $(jq .umask /config/settings.json)))
+        TRANSMISSION_UMASK_OCTAL=$(printf '%03g' $(printf '%o\n' $(jq .umask ${TRANSMISSION_HOME}/settings.json)))
         DIR_PERMS=$(printf '%o\n' $((0777 & ~TRANSMISSION_UMASK_OCTAL)))
         FILE_PERMS=$(printf '%o\n' $((0666 & ~TRANSMISSION_UMASK_OCTAL)))
         echo "Mask: ${TRANSMISSION_UMASK_OCTAL}"
--- a/transmission/userSetup.sh
+++ b/transmission/userSetup.sh
@@ -42,20 +42,17 @@ if [ -n "$PUID" ] && [ ! "$(id -u root)" -eq "$PUID" ]; then
             ${TRANSMISSION_WATCH_DIR}
 
         echo "Setting permissions for download and incomplete directories"
-        TRANSMISSION_UMASK_OCTAL=$(printf '%03g' $(printf '%o\n' ${TRANSMISSION_UMASK}))
+        TRANSMISSION_UMASK_OCTAL=$(printf '%03g' $(printf '%o\n' $(jq .umask /config/settings.json)))
         DIR_PERMS=$(printf '%o\n' $((0777 & ~TRANSMISSION_UMASK_OCTAL)))
         FILE_PERMS=$(printf '%o\n' $((0666 & ~TRANSMISSION_UMASK_OCTAL)))
         echo "Mask: ${TRANSMISSION_UMASK_OCTAL}"
         echo "Directories: ${DIR_PERMS}"
         echo "Files: ${FILE_PERMS}"
 
-        for THIS_DIR in ${TRANSMISSION_DOWNLOAD_DIR} ${TRANSMISSION_INCOMPLETE_DIR}
-        do
-            find ${THIS_DIR} -type d -print0 | xargs -0 \
-                chmod $(printf '%o\n' $((0777 & ~TRANSMISSION_UMASK_OCTAL)))
-            find ${THIS_DIR} -type f -print0 | xargs -0 \
-                chmod $(printf '%o\n' $((0666 & ~TRANSMISSION_UMASK_OCTAL)))
-        done
+        find "${TRANSMISSION_DOWNLOAD_DIR}" "${TRANSMISSION_INCOMPLETE_DIR}" -type d \
+        -exec chmod $(printf '%o\n' $((0777 & ~TRANSMISSION_UMASK_OCTAL))) {} +
+        find "${TRANSMISSION_DOWNLOAD_DIR}" "${TRANSMISSION_INCOMPLETE_DIR}" -type  f \
+        -exec chmod $(printf '%o\n' $((0666 & ~TRANSMISSION_UMASK_OCTAL))) {} +
 
         echo "Setting permission for watch directory (775) and its files (664)"
         chmod -R o=rX,ug=rwX \
--- a/transmission/userSetup.sh
+++ b/transmission/userSetup.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 # More/less taken from https://github.com/linuxserver/docker-baseimage-alpine/blob/3eb7146a55b7bff547905e0d3f71a26036448ae6/root/etc/cont-init.d/10-adduser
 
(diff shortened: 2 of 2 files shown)


wriches:
Addresses #833

clement-z:
###  `sh`/`bash`

The `umask` mask specification follows `chmod`'s and so there is still the need to convert to octal. I did not see why `bash` was actually required over `sh`, it seems to work on my machine with `sh`. However I am personally not against using `bash` since it is installed anyway.

### `chmod`ing

According to the coreutils [manpage](https://www.gnu.org/software/coreutils/manual/html_node/chmod-invocation.html), `chmod` can reset files to respect the default `umask` directly:

```bash
# Set file permissions of DIR and subsidiary files to
# be the umask default, assuming execute permissions for
# directories and for files already executable.
chmod -R a=,+rwX dir
```

I didn't know about this behaviour until now, but it seems cleaner to use it, rather than the `find` approach (which I assume has to be used since you cannot use the `X` flag).
I tested it out on alpine armhf (inside the transmission-openvpn container) and it **does not behave as expected from the coreutils documentation**. The behaviour is consistent however with the [POSIX specification for `chmod`](https://man7.org/linux/man-pages/man1/chmod.1p.html), as well as the rest of the coreutils documentation. I will report it tomorrow.

So it seems to me that running `umask ${OCTAL_UMASK}; chmod a=,+rwX dir` is strictly identical to your solution, as the executable bit of already executable files is **not** preserved.

This however *could be* considered **a breaking change** from the original behaviour. The possibility of something breaking in a torrent download folder is remote, but for example assume you have extracted some software directly in the `data/completed` folder: if you try to use it again after your method, the system will complain. This does not happen with the current version.

A corrected `chmod` rule which has the intended behaviour would be:

```bash
chmod -R a-rw+X,go-x,+rwX dir
```

I think it reads well but still:

- `a-rw+X`: remove `rw` and set `x` permissions for everyone is someone had it
- `go-x,`: remove `x` bits of `group` and `other`
- (at this point the file is either `--x------` if it was executable or `---------` otherwise)
- `+rwX `: try to add `rwxrwxrwx` to directories and executable files and `rw-rw-rw-` to normal files.

### Cleaner method

So to sum it up, a cleaner method would therefore be along the lines of:

```bash
TRANSMISSION_UMASK_OCTAL=$(printf '%04o' ${TRANSMISSION_UMASK})
echo "Setting default permissions for download and incomplete directories (umask = ${TRANSMISSION_UMASK_OCTAL})"

# Set file permissions of transmission folders and 
# subsidiary files to follow ${TRANSMISSION_UMASK},
# assuming execute permissions for directories and
# for files already executable.
( # scope umask alteration
umask ${TRANSMISSION_UMASK_OCTAL}
chmod -R a-rw+X,go-x,+rwX ${TRANSMISSION_DOWNLOAD_DIR} ${TRANSMISSION_INCOMPLETE_DIR}
)
```

The parentheses scope the `umask` alteration inside a subshell to avoid unintended consequences in the rest of the script (but currently there is no conflict). It works in both `sh` and `bash`.

Could you implement something like this and see if it works on your side?

wriches:
@clement-z Thanks, excellent feedback.

### `sh`/`bash`
I encountered some errors with `sh` when performing the bitwise subtraction, but I can't recall the exact issue and can't seem to replicate it right now. Besides, if we can use the umask directly, as in your example, there is no need for the bitwise subtraction anyway.

### `chmod`ing

The main reason I opted for setting the permissions with separate `chmod` commands for files and directories is that the log output already included a statement of setting permissions to `755` and `644`, as well as for other directories. So to maintain this logging, the mode had to be calculated.

It's certainly overly verbose though, and I don't like using `find` and `xargs` for such a simple function either, even though they are available in the base Alpine image!

**@Maintainers:**  Do you want to log the permissions set, or just log the umask?

> I tested it out on alpine armhf (inside the transmission-openvpn container) and it **does not behave as expected from the coreutils documentation**. The behaviour is consistent however with the [POSIX specification for `chmod`](https://man7.org/linux/man-pages/man1/chmod.1p.html), as well as the rest of the coreutils documentation. I will report it tomorrow.

Yes, I'm seeing the same behaviour - it doesn't respect existing file execute bits.

> A corrected `chmod` rule which has the intended behaviour would be:
> 
> ```shell
> chmod -R a-rw+X,go-x,+rwX dir
> ```

Unfortunately I don't think this works, since the `X` in the final clause (`+rwX`) will apply to **all** of `u`, `g` and `o` as long as there is an execute bit set on at least one. As stated in the `man`:

```
The perm symbol X shall represent the execute/search portion of the
file mode bits if the file is a directory or if the current
(unmodified) file mode bits have at least one of the execute bits
(S_IXUSR, S_IXGRP, or S_IXOTH) set. It shall be ignored if the file
is not a directory and none of the execute bits are set in the
current file mode bits.
```
Whereas to avoid breaking changes, we of course only want maintain the execute bit on the `who` that already has it. An example:
```
$ ls -la test1/
total 0
drwxrwxr-- 1 will users 10 Nov 16 06:09 .
drwxr-xr-x 1 will users 28 Nov 16 06:09 ..
-rwxrwxr-- 1 will users  0 Nov 16 04:05 test2
$ umask 022
$ chmod -R a-rw+X,go-x,+rwX test1
$ ls -la test1/
total 0
drwxr-xr-x 1 will users 10 Nov 16 06:09 .
drwxr-xr-x 1 will users 28 Nov 16 06:10 ..
-rwxr-xr-x 1 will users  0 Nov 16 04:05 test2
```
It seems like there should be a simpler solution to this, but some guidance from the maintainers would I think be welcome here - as they may not even want to maintain execute bits on the download directories.

> ### Cleaner method
> So to sum it up, a cleaner method would therefore be along the lines of:
> 
> ```shell
> TRANSMISSION_UMASK_OCTAL=$(printf '%04o' ${TRANSMISSION_UMASK})
> ```

Nice! That's cleaner, I didn't realise you could zerofill and change base in a single `printf` format. I've been doing that in two steps for years!

clement-z:
> Unfortunately I don't think this works, since the X in the final clause (+rwX) will apply to all of u, g and o as long as there is an execute bit set on at least one. 

Your understanding of `X` is correct, but AFAIK this is **the expected behaviour**! The original code also did this (`go+rX`):

https://github.com/haugene/docker-transmission-openvpn/blob/93d0141d928ec661411e2c6f0e079aeff2c018dd/transmission/userSetup.sh#L45-L47

The difference is that `chmod -R a-rw+X,go-x,+rwX dir` is applying the `umask` rather than automatically setting the `x` bit for everyone. So for example if `g` or `o` should not have the `x` bit set by default on executable files and directories, set the corresponding bits in the `umask` (`umask $((umask | 0011))`).

wriches:
> Your understanding of `X` is correct, but AFAIK this is **the expected behaviour**!

I see. I think this behaviour is quite risky. In your previous comment you mentioned the risk of this PR _removing_ execute bits, but the code as it stands actually _adds_ execute bits. IMO that's much riskier.

I think it's far from transparent to the average user that if the umask is left at the default of `002`, then an execute bit could be added to their file's group. Typically, people automatically associate `002` with file permissions of `664` and directory permissions of `775`. Yet if they have an executable lying around with `700` permissions, then with the current code or your proposed code, an execute bit would be added to both `group` and `other`.

clement-z:
I can understand why you would think that, but it's a fallacy. If `g`/`o` have read access to the file, they can make an executable copy of it. The only place where it really matters is in `setuid` cases.

Arguably, we could decide to clear the `setuid` (and eventually `setgid` bits). Removing `setuid` can be done with:

```
chmod -R u-s,a-rw+X,go-x,+rwX dir
```

Dockerfile
```diff
@@ -38,6 +38,7 @@ ENV OPENVPN_USERNAME=**None** \
     TRANSMISSION_DOWNLOAD_DIR=/data/completed \
     TRANSMISSION_INCOMPLETE_DIR=/data/incomplete \
     TRANSMISSION_WATCH_DIR=/data/watch \
+    TRANSMISSION_UMASK=2 \
```
clement-z, line 41: Is it needed? I'm nitpicking but:

The default transmission `umask` is already defined in the `settings.json`, which is modified if needed to respect environment variables **before** `userSetup.sh` is run. So we should use this instead IMO. For example with something like:

```
TRANSMISSION_UMASK=$(python3 -c 'import json; f = open("/config/settings.json"); settings = json.load(f); f.close(); print(settings["umask"])')
```

transmission/userSetup.sh
```diff
@@ -41,10 +41,21 @@ if [ -n "$PUID" ] && [ ! "$(id -u root)" -eq "$PUID" ]; then
             ${TRANSMISSION_INCOMPLETE_DIR} \
             ${TRANSMISSION_WATCH_DIR}
 
-        echo "Setting permission for files (644) and directories (755)"
-        chmod -R go=rX,u=rwX \
-            ${TRANSMISSION_DOWNLOAD_DIR} \
-            ${TRANSMISSION_INCOMPLETE_DIR} \
+        echo "Setting permissions for download and incomplete directories"
+        TRANSMISSION_UMASK_OCTAL=$(printf '%03g' $(printf '%o\n' ${TRANSMISSION_UMASK}))
```
clement-z, line 45: Adapt depending on the use/definition of `TRANSMISSION_UMASK`.

clement-z (changes requested):
`TRANSMISSION_UMASK` does not need to **also** be a default environment variable, a default value is already defined in `default-settings.json`, and it's better to keep it only in one spot.

clement-z:
Hello @wriches, can I ask whether you are still working on this PR?

transmission/userSetup.sh
```diff
@@ -41,10 +41,21 @@ if [ -n "$PUID" ] && [ ! "$(id -u root)" -eq "$PUID" ]; then
...
+        DIR_PERMS=$(printf '%o\n' $((0777 & ~TRANSMISSION_UMASK_OCTAL)))
+        FILE_PERMS=$(printf '%o\n' $((0666 & ~TRANSMISSION_UMASK_OCTAL)))
+        echo "Mask: ${TRANSMISSION_UMASK_OCTAL}"
+        echo "Directories: ${DIR_PERMS}"
+        echo "Files: ${FILE_PERMS}"
+
+        for THIS_DIR in ${TRANSMISSION_DOWNLOAD_DIR} ${TRANSMISSION_INCOMPLETE_DIR}
```
clement-z, line 52: Doesn't handle spaces in directories names.

clement-z (changes requested):

clement-z:
Hello @wriches, if you are not working on this anymore, I will follow up in another PR based off your commits.

pkishino:
@clement-z hey, mind checking (and closing this if not needed) as well as the other dns leak one I just made?

bubbaiOS:
Curious if the above fix will be applied to current (legacy) branch? Agreed it needs to be fixed in dev using new Json config, but folks with the legacy setup still run into permission issues on startup without this fix. Otherwise we're either rolling our local copy and changing userSetup.ah ourselves, or it breaks at each startup/reboot and we have to manually go back in and change perms.

pkishino:
@clement-z @haugene I think this can be closed as no update..

amosshapira:
> @clement-z Thanks, excellent feedback.
> > I tested it out on alpine armhf (inside the transmission-openvpn container) and it **does not behave as expected from the coreutils documentation**. The behaviour is consistent however with the [POSIX specification for `chmod`](https://man7.org/linux/man-pages/man1/chmod.1p.html), as well as the rest of the coreutils documentation. I will report it tomorrow.
> 
> Yes, I'm seeing the same behaviour - it doesn't respect existing file execute bits.

The docker image isn't using coreutils - it's using BusyBox:

```
bash-5.1# chmod --version
chmod: unrecognized option: version
BusyBox v1.32.1 () multi-call binary.

Usage: chmod [-Rcvf] MODE[,MODE]... FILE...

Each MODE is one or more of the letters ugoa, one of the
symbols +-= and one or more of the letters rwxst

	-R	Recurse
	-c	List changed files
	-v	List all files
	-f	Hide errors
```

Ref: https://boxmatrix.info/wiki/Property:chmod

Dockerfile
amosshapira: (nitpicking on the nitpicking ;) )
`jq` is already available on the image, you can achieve the same with simpler code:
```
# jq .umask /config/settings.json
2
```

transmission/userSetup.sh
amosshapira: 1. You can pass multiple root directories to `find`, the loop is redundant.
2. You can use `find`'s `exec ... +` to do away with xargs:
```
find "${TRANSMISSION_DOWNLOAD_DIR}" "${TRANSMISSION_INCOMPLETE_DIR}" -type d \
  -execc chmod $(printf '%o\n' $((0777 & ~TRANSMISSION_UMASK_OCTAL))) +
find "${TRANSMISSION_DOWNLOAD_DIR}" "${TRANSMISSION_INCOMPLETE_DIR}" -type  f \
  -exec chmod $(printf '%o\n' $((0666 & ~TRANSMISSION_UMASK_OCTAL))) +
```

amosshapira (approved):

Issue #1499, opened 2020-11-11
Setting TRANSMISSION_RATIO_LIMIT to float value causes Transmission to be unable to connect to trackers


Yoronex:
**Describe the problem**
When setting the variable TRANSMISSION_RATIO_LIMIT to a float (e.g. 2.0) instead of an integer (e.g. 2), Transmission is unable to connect with any trackers. Transmission does seem to start successfully.

**Add your docker run command**
```
/usr/bin/docker run 
        --name transmission-openvpn \
        --cap-add=NET_ADMIN \
        --device=/dev/net/tun \
        -v /media/roy/:/data \
        -e "OPENVPN_PROVIDER=PIA" \
        -e "OPENVPN_USERNAME=***" \
        -e "OPENVPN_PASSWORD=***" \
        -e "OPENVPN_CONFIG=Sweden" \
        -e LOCAL_NETWORK=192.168.0.0/16 \
        -e "TRANSMISSION_UMASK=0" \
        -e TRANSMISSION_ALT_SPEED_DOWN=100 \
        -e TRANSMISSION_ALT_SPEED_TIME_BEGIN=0 \
        -e TRANSMISSION_ALT_SPEED_TIME_DAY=127 \
        -e TRANSMISSION_ALT_SPEED_TIME_ENABLED=true \
        -e TRANSMISSION_ALT_SPEED_TIME_END=1 \
        -e TRANSMISSION_ALT_SPEED_UP=100 \
        -e TRANSMISSION_DOWNLOAD_DIR=/data/Torrents \
        -e TRANSMISSION_INCOMPLETE_DIR_ENABLED=false \
        -e TRANSMISSION_RATIO_LIMIT=2.0 \
        -e TRANSMISSION_RATIO_LIMIT_ENABLED=true \
        -e TRANSMISSION_SPEED_LIMIT_DOWN=5000 \
        -e TRANSMISSION_SPEED_LIMIT_DOWN_ENABLED=true \
        -e TRANSMISSION_SPEED_LIMIT_UP=5000 \
        -e TRANSMISSION_SPEED_LIMIT_UP_ENABLED=true \
        -p 9091:9091 \
        --dns 8.8.8.8 \
        --dns 8.8.4.4 \
        haugene/transmission-openvpn
```

**Logs**

```
Starting container with revision: 05f657435af4ca118b8cdf6a03f5ce895149cbc8
mknod: /dev/net/tun: File exists
Using OpenVPN provider: PIA
[...]

Issue #432, opened 2018-02-20
web ui on transmission doesn't show any activity


hildebrau:
I've been battling this all container all day.  It was working fine a few months back.  Today I had to overhaul the VPN connection and set it to custom and provide it with a new .ovpn file to get it to connect.  Then transmission finally started.  That's where this issue starts.

I've tried the latest, 1.9, and dev tags.  I'm currently on the haugene/transmission-openvpn:dev tag.

I've tried all three WEB UI's.  The default throws an error of not being able to connect and asks me to reload the page.  Mind you, the Transmission UI loads up fine, but it simply doesn't show any active torrent history, etc.

The other two WEB UI's do not throw errors; but they also do not show me any active torrent or past torrent history.

Inside the container terminal, I was able to do a "transmission-remote" command to list the current torrent queue and even add a torrent.  That torrent has since wrapped up to completion.  But I've yet to get the web UI to show anything about it.

I'm sort of out of ideas.  I've been through the container console logs, and the transmission.log; but I don't see anything of use in there.  It seems to think it is working just peachy.

I've checked permissions; and near as I can tell, all the /data subdirs and files have user +rw access at least, and most have g+rw and o+r access.  User and Group are the "abc:users" default.  transmission-daemon appears to be running as user 'abc' in the container.

[2018-02-20 14:36:11.440 CST] Transmission 2.84 (14307) started (session.c:738)
[2018-02-20 14:36:11.440 CST] RPC Server Adding address to whitelist: 127.0.0.1 (rpc-server.c:878)
[2018-02-20 14:36:11.440 CST] RPC Server Serving RPC and Web requests on port 127.0.0.1:9091/transmission/ (rpc-server.c:1122)
[2018-02-20 14:36:11.440 CST] Port Forwarding Stopped (port-forwarding.c:180)
[2018-02-20 14:36:11.440 CST] UDP Failed to set receive buffer: requested 4194304, got 425984 (tr-udp.c:78)
[2018-02-20 14:36:11.440 CST] UDP Please add the line "net.core.rmem_max = 4194304" to /etc/sysctl.conf (tr-udp.c:83)
[2018-02-20 14:36:11.440 CST] UDP Failed to set send buffer: requested 1048576, got 425984 (tr-udp.c:89)
[2018-02-20 14:36:11.440 CST] UDP Please add the line "net.core.wmem_max = 1048576" to /etc/sysctl.conf (tr-udp.c:94)
[...]

PR #1523, opened 2020-11-16
Remove legacy options and update defaults

ilike2burnthing:
https://github.com/transmission/transmission/wiki/Editing-Configuration-Files

Deprecated in 1.5x and 1.4x


ilike2burnthing:
_There are some other differences between Transmission's default settings and those here. I haven't added them to this PR, as there's probably a good reason for most if not all of them. Hopefully someone can shed some light for those comments in bold._

### Missing
`lazy-bitfield-enabled: true` - should really be a default of 'false' anyway, if it even works


### Extra
`watch-dir-force-generic:` - from what I can tell, this does still exist, it's just not documented


### Different
*`incomplete-dir-enabled: false` - **why is the default 'true' used here?**

*`peer-limit-global: 240` - not much of a difference between that and '200'. **Update?**
*`peer-limit-per-torrent: 60` - same as above for '50'. **Update?**

*`peer-port-random-low: 1024` - '49152' makes more sense anyway
*`port-forwarding-enabled: true` - **why is the default 'false' used here?**

*`prefetch-enabled: true` - **set to '1' only in DockerEnv**

*`ratio-limit: 2.0` - python issue #1499

`rpc-host-whitelist-enabled: true` - **set to 'false' only in default-settings.json** - #432
*`rpc-whitelist: 127.0.0.1` - **set to '127.0.0.1,::1' only in default-settings.json**
*`rpc-whitelist-enabled: true` - explained in  #432 

*`umask: 18` - **why is the default '2' used here?**

`utp-enabled: true` - really should be a default of 'false' anyway #978

*= value unchanged from initial commit - https://github.com/haugene/docker-transmission-openvpn/commit/4eaaae0ec12cd2bad021190dcc10895c356f0f92

haugene:
Thanks for taking a look at this! :+1:

I'll comment on what I remember/think to be true.

1. incomplete-dir-enabled: I think that's just me, from the old days. We built it to have a setup that we thought to be the "optimal one". And the container was dumber back then. Could be just because there are scripts that ensure that the folder exists etc. I have no problem with it being reverted to false - as long as we update the userSetup.sh scripts and others that modify permissions etc
2. peer-limit*: Agree that they should be changed to Transmission's default settings.
3. prefetch-enabled... Are we only having weird values in DockerEnv? The dafault is correct other than that?
4. The ratio is unfortunate because of the bug. We should maybe add a list to the python script on what variables can be double.
5. rpc-whitelist stuff. I guess this has to do with all clients being on other hosts than the container.
6. umask: :shrug: 
7. utp-enabled: Ah. So I re-introduced that one. Good catch, thanks. Should be false :+1: 
8. watch-dir-force-generic: yes, not very documented but I did find some reference to it when I looked once. It's for the systems that cant use inotify to detect new files in watch dir.

While we're on it. When I saw the DockerEnv file I got the urge to delete it. Thoughts?

ilike2burnthing:
Sorry, I was typing all that at 4am. Reading it back now, it really seems like I was saying that settings with default values different to those from Transmission are somehow _**wrong**_. Really not what I was going for, was mostly just curious.

1. Given that many people use this along side Sonarr, Radarr, *arr, I think a default of true makes sense; separate incomplete and complete download folders to avoid any issues and make logs clearer. So I'd say keep it as true.
2. Any preference for whether I make a separate PR or just add them to this one?
3. Yea, in default-settings.json it is `"prefetch-enabled": true,`, whereas in DockerEnv it is `#TRANSMISSION_PREFETCH_ENABLED=1`, though it may be that both can be used but isn't documented (just a guess).
4. Outside of my skillset I'm afraid, so I'll leave that one to you (not like you have enough to work on or anything). 😉 
5. I suppose my concern there is the default in DockerEnv is `#TRANSMISSION_RPC_HOST_WHITELIST_ENABLED=true`, so could that potentially cause issues?
6. Lets just smile and nod. 😋 
7. Apologies for the confusion, the listed settings above (e.g. `utp-enabled: true`) are the defaults provided by Transmission. So don't worry, it's still false here, that issue hasn't been reintroduced.
8. Makes sense!

As for DockerEnv, I have no strong opinions either way:
**+ve:** it's a a good reference for users to have, so they know the out-of-the-box settings
**-ve:** double the upkeep for changes

haugene:
I see. It wasn't 4 am here when I responded, but it seems I've gotten it the wrong way around a couple of times as well :laughing: 
Anyways. I think I'll opt for deleting DockerEnv and try to put the upkeep effort into the docs instead. That solves number 3. and 5. here as well. For the rest:

1. :+1: We keep it
2. Same for me, just push everything in here if you want.
3. N/A :point_up: 
4. Hah. Getting there :crossed_fingers: :smile: 
5. N/A :point_up: 
6. Yes... These guys seem to know what they're doing though #1518. Might have a preference here.
7. Then we'll keep it :+1: 
8. :+1: 

Sounds like we have a plan :clap:

ilike2burnthing:
Ready to go! Thanks for the insight

clement-z:
I think keeping in sync with transmission's defaults is good. I wonder whether it's necessary with `umask` though, as IMO `002` is a relatively sane default for this container's intents.

Changing back `umask` to the default might be unexpected on some systems where group write permission is assumed on torrents. For example, if `sonarr` runs as `sonarr:mygroup`, `transmission` as `transmission:mygroup`, and `sonarr` is set to delete completed torrents after copying, it will result in (non-critical) errors.

So I think the best option is either leaving `umask` as-is or re-aligning it with transmission's default in the next major release.

haugene:
Great stuff. Thanks for the update @ilike2burnthing :+1:

PR #1530, opened 2020-11-18
Ensure transmission shutdown

### LINKED ISSUE #1513: It is not checked whether transmission actually exited

clement-z:
* Carry and improve #1514 
* Close #1513 

After testing in the container, I found out `ps -p` is not supported in alpine (with `busybox`). So we avoided a regression 😅 

I set the default timeout to 10 seconds, and set the timeout step to 1 second, which both prevent the "Waiting for ..." message if transmission shuts down under 1s, and reduce code verbosity.

### DIFF
--- a/transmission/stop.sh
+++ b/transmission/stop.sh
@@ -13,7 +13,7 @@ PID=$(pidof transmission-daemon)
 kill $PID
 # Give transmission-daemon time to shut down
 for i in {1..10}; do
-    ps -p $PID &> /dev/null || break
+    [[ -z "$(pidof transmission-daemon)" ]] && break
     sleep .2
 done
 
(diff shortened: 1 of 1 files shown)


haugene:
Looks great! Tested it locally and it works like a charm :+1: I think 10 sec is a bit too long, so I'll set the default to 5 sec.
But I'll just do that after merging. Thanks for a great contribution :clap:

PR #1533, opened 2020-11-19
Remove cause of "yes: Broken pipe" error

### LINKED ISSUE #1532: yes: broken pipe error
Since 56a6a6f6196, the PIA port forwarding script (always?) prints a `yes: Broken pipe` message, which often confuses people reporting issues.

From what I can see, it is just stemming from this:

https://github.com/haugene/docker-transmission-openvpn/blob/baed33520630e88ba34842cab05f2cd3fd6cf28f/openvpn/pia/update-port.sh#L39

This essentially is equivalent to `echo -e "\n\n\n"`. I don't think it serves any other purpose. Can we delete it?

clement-z:
Closes #1532 

I'm pretty sure this line has no purpose (other than printing 3 new lines).

### DIFF
--- a/openvpn/pia/update-port.sh
+++ b/openvpn/pia/update-port.sh
@@ -36,8 +36,6 @@ get_auth_token () {
 
 get_auth_token
 
-yes '' | sed 3q
-
 get_sig () {
   pf_getsig=$(curl --insecure --get --silent --show-error \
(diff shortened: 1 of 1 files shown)


haugene:
LGTM :+1: Thanks! :raised_hands:

PR #1552, opened 2020-11-24
Update PIA configure-openvpn.sh with updated URLs

jaller200:
It appears that PIA has changed their URLs from "openvpn{-strong/tcp/strong-tcp}-nextgen" to "openvpn{-strong/tcp/strong-tcp}". The old URLs no longer work properly.

See here: https://www.privateinternetaccess.com/helpdesk/kb/articles/where-can-i-find-your-ovpn-files

### DIFF
--- a/openvpn/pia/configure-openvpn.sh
+++ b/openvpn/pia/configure-openvpn.sh
@@ -3,14 +3,13 @@
 set -e
 
 # These are the possible bundles from PIA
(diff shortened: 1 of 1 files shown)


haugene:
Yes indeed. Seems the nextgen servers are now just the current generation and they skip that postfix on the naming.

Thanks for pointing it out and providing a fix :+1: I'm just on my phone here and can't test it but it looks good so I'll merge it now.

PR #1568, opened 2020-11-29
Add transmission-daemon options

marinierb:
Allows putting additional options on the transmission-daemon start up command.


marinierb:
My first ever contribution to an open source project! Hope it's all good!

Allows putting additional options to the transmission-daemon at start up.

marinierb:
Turns out I can achieve what I wanted with TRANSMISSION_MESSAGE_LEVEL=0. No need for this change.

haugene:
Well, I guess we won't introduce it before there is a real need but the changed looked good at least :smile:  :+1: Thanks for wanting to contribute.

PR #1580, opened 2020-12-02
PIA: normalize config file names so changing case or switching spaces…

### LINKED ISSUE #1572: OPENVPN_CONFIG suddenly became case sensitive (or the case changed for PIA)
My Docker container kept unexpectedly terminating and restarting itself. I checked the logs and found that none of the [suggested PIA servers](https://github.com/haugene/docker-transmission-openvpn/issues/1522#issuecomment-729205947) were working (France,Sweden,Italy,Belgium,Austria,Denmark,Norway,Ireland). The logs indicated that the same servers in all lowercase were valid options. So I changed the case, and things are working now:

```
OPENVPN_CONFIG=france,sweden,italy,belgium,austria,denmark,norway,ireland
```

Perhaps consider normalizing the value before parsing it?

ammmze:
… and underscores will not break this

This should resolve #1572

### DIFF
--- a/openvpn/pia/configure-openvpn.sh
+++ b/openvpn/pia/configure-openvpn.sh
@@ -26,5 +26,19 @@ curl -sSL "${baseURL}/${PIA_OPENVPN_CONFIG_BUNDLE}.zip" -o "$tmp_file"
 echo "Extract OpenVPN config bundle into PIA directory $VPN_PROVIDER_HOME"
 unzip -qjo "$tmp_file" -d "$VPN_PROVIDER_HOME"
 
+# Normalize *.ovpn files to be lower case and replace sequential non-alphanumeric with an underscore
+pushd "$VPN_PROVIDER_HOME" > /dev/null
+for file in *.ovpn; do
+    normalized="$(echo "$file" | awk '{gsub(/[^a-zA-Z0-9\.]+/, "_") ; print tolower($0)}')"
+    mv "$file" "$normalized"
+done
+unset normalized
+popd > /dev/null
+
+# Normalize OPENVPN_CONFIG environment variable to be lower case and replace sequential non-alphanumeric with an underscore
+if [[ -n "${OPENVPN_CONFIG-}" ]]; then
+    OPENVPN_CONFIG="$(echo "$OPENVPN_CONFIG" | awk '{gsub(/[^a-zA-Z0-9\.,]+/, "_") ; print tolower($0)}')"
+fi
+
 # Select a random server as default.ovpn
 ln -sf "$(find "$VPN_PROVIDER_HOME" -name "*.ovpn" | shuf -n 1)" "$VPN_PROVIDER_HOME"/default.ovpn


ammmze:
With these changes in place, I set my `OPENVPN_CONFIG` back to `CA Toronto` and verified that upon startup it used `ca_toronto.ovpn` as seen in some of the later logs:

```
Starting OpenVPN using config ca_toronto.ovpn
Modifying /etc/openvpn/pia/ca_toronto.ovpn for best behaviour in this container
```

openvpn/pia/configure-openvpn.sh
```diff
@@ -26,5 +26,19 @@ curl -sSL "${baseURL}/${PIA_OPENVPN_CONFIG_BUNDLE}.zip" -o "$tmp_file"
 echo "Extract OpenVPN config bundle into PIA directory $VPN_PROVIDER_HOME"
 unzip -qjo "$tmp_file" -d "$VPN_PROVIDER_HOME"
 
+# Normalize *.ovpn files to be lower case and replace sequential non-alphanumeric with an underscore
+pushd "$VPN_PROVIDER_HOME" > /dev/null
+for file in *.ovpn; do
+    normalized="$(echo "$file" | awk '{gsub(/[^a-zA-Z\.]+/, "_") ; print tolower($0)}')"
```
ammmze, line 32: Just realized now that I'm laying in bed, I forgot to include numbers, so this will basically only keep alpha characters. I'll try to remember to update that in the morning after I get all my morning stuff taken care of.

ilike2burnthing: reminder for you

ammmze: Thanks! I've just updated the pattern to include numbers.

clement-z (approved):
I think the changes are good quality (and although I am not a bash expert, I thing you did a good job on the quoting 👍). IMO it can be merged as-is but I still have a few comments: 

* On the use of `for file in *.ovpn`: see http://mywiki.wooledge.org/BashPitfalls#line-57 (what if they set up a folder hierarchy later?)
* It would make sense to move the normalizing into a function `normalize_config_name` or something, but it's a nitpick.

ammmze:
@clement-z Thanks for a taking a look. 👍🏻

I think on your first point about a folder hierarchy, those wouldn't get picked up here. IMO that could be an enhancement in the future, but can probably work it in if there is a strong enough desire. This would also necessitate normalizing directory names as well probably. So it adds a fair bit of complexity when we start going recursive. Also I'm not sure if the shared code supports nested config files, does it? For example, let's say they make a folder `CA` and a file `Toronto.ovpn`, would the shared code support `OPENVPN_CONFIG=CA/Toronto`? I don't have enough experience or depth of knowledge to know.

clement-z:
### Handling directories

Very good point! I quickly checked and it seems like it should indeed support nested config files, at least in `openvpn/start.sh`. It would need more checking and testing to make sure though.

I just thought it would be better to future-proof it now, rather than wait for it to maybe break and issues pop up! I think it just doesn't hurt to make it handle recursion:
* Use `./**.ovpn` instead of `*.ovpn`
* Don't substitute `/`
* Add `mkdir -p "$(dirname "$normalized")"` before the `mv`
* Deleting the empty directory after normalization is not necessary

### No `.ovpn` files

If there are no `.ovpn` files, your loop is **still executed once** (see [BashPitfalls#line-70](http://mywiki.wooledge.org/BashPitfalls#line-70)).

You should consider simply testing for existence of file and `exit 1` if it does not.

haugene:
Thank you for a good proposal @ammmze, but I'm going to close this for now. I actually don't support changing the file names of what has been downloaded. I don't doubt that this would lead to more stable naming and that things would break less often.

But the names of the configs are for PIA to decide, and a config location could always be renamed completely or discontinued and you would still get an error that the name is wrong. We log it, and we list alternative configs which should make this relatively intuitive to the users.

If we were to add functionality like this I would rather that it was in the lookup phase. That we grep for files that match, ignoring case. Or even use some tool to establish the closest match. Then we could have an option to go with the closest match, and an opt-out to ensure that you get the exact config you specified  (some don't want fallbacks to closest match).

If you're up for implementing a feature like that you're very welcome to create a new PR for it :+1:

ammmze:
No worries! I appreciate your feedback. What you propose sounds like a good solution as well. However, I don't think i'll be the one to implement it. I'm working towards learning kubernetes and running a new stack configuration in my home lab with kubernetes that approaches VPN for a container in a different way that decouples the torrent client from the vpn. But I do thank you for your work creating and maintaining this docker image. It has been great in my home lab!

PR #1610, opened 2020-12-16
transmission: fix settings ratio limits to float values

paulfantom:
Transmission default value for the ratio limit is `2.0`, not `2` (source: https://github.com/transmission/transmission/wiki/Editing-Configuration-Files#scheduling). This slight difference causes the` updateSettings.py` script to treat this setting as `int` instead of `float`. 

Currently, passing float causes the following error:
```
Generating settings.json for Transmission from environment and defaults default-settings.json
Could not coerce TRANSMISSION_RATIO_LIMIT value 4.1 to expected type <class 'int'>
Traceback (most recent call last):
  File "updateSettings.py", line 69, in <module>
    env_value = setting_type(env_value)
ValueError: invalid literal for int() with base 10: '4.1'
```

After applying this patch everything works as expected:
```
Generating settings.json for Transmission from environment and defaults default-settings.json
Overriding ratio-limit because TRANSMISSION_RATIO_LIMIT is set to 4.1
```

In addition to the issue fixed here, there is another issue - any errors inflicted by updateSettings.py should result in a hard fail for the transmission container. Without this user can be unaware of problems. (Potential fix for this is in #1611)

### DIFF
--- a/transmission/default-settings.json
+++ b/transmission/default-settings.json
@@ -37,7 +37,7 @@
     "prefetch-enabled": true,
     "queue-stalled-enabled": true,
     "queue-stalled-minutes": 30,
-    "ratio-limit": 2,
+    "ratio-limit": 2.0,
     "ratio-limit-enabled": false,
     "rename-partial-files": true,
     "rpc-authentication-required": false,


clement-z:
Actually, I don't think the `json` format makes any distinction between integer or floating point. I don't know how transmission parses the `json`, but it would be very weird if they did not just cast it to `int` when the value must be so.

Of course, your change is 👍 and it fixes the issue for new setups. 

But it doesn't if the user manually set its ratio to e.g. `2` in `settings.json`. So I'm wondering whether there would be any problem if we decided to simply parse all numbers in the `json` as `float` using `json.load(config_file, parse_int=float)` [(doc)](https://docs.python.org/3/library/json.html#json.load).

paulfantom:
With this change, if user sets value to int (ex. `2`) in settings.json it still works as expected.

clement-z:
Well I think you're wrong (although it's the current code which is somehow illogical, not your reasoning). Your fix only works for new setups (when `settings.json` does not exist yet).

Correct me if I am wrong, but as far as I can see ([updateSettings.py#L44](https://github.com/haugene/docker-transmission-openvpn/blob/a4d65774f855a04070766e53346661f48c76fa0e/transmission/updateSettings.py#L44)), the script as it stands reads `settings.json` if it exists, and deduces the type from **it**, instead of `default-settings.json`. So if the user manually set the ratio to a string that `json.load()` parses as an integer, the problem is still there (if he then sets `TRANSMISSION_RATIO_LIMIT` to a float value).

The potential fixes, which do work **after** `settings.json` has already been created/modified are either:

* to always deduce types from `default-settings.json`, which IMO is very reasonable;
* as I outlined in my previous comment, to parse all numbers as float, instead of making the distinction between integer and floats and **then** print it without the decimal point if it is indeed an integer.

I think the first option is the cleanest, what do you think? (Having the second one on top wouldn't hurt anyway, as JSON only knows "numbers" AFAIK, but it is unnecessary I guess.)

paulfantom:
> I think the first option is the cleanest, what do you think?

Yes, that makes sense.

Thanks for your extended clarification of the topic :+1:

clement-z:
Just a reminder that this fix needs a follow up fix to work on an already existing setup: deduce type information from default settings rather than current.

Issue #1455, opened 2020-11-04
DNS Is Leaking! Uses ISP DNS instead of VPN DNS!


RickBankers:
**Describe the problem**
When running a DNS leak test from this site we see DNS is leaking and using your router or ISPs DNS instead of the VPN. This does not protect your information.

**Add your docker run command**
Nothing fancy, Just using PIA standard config.

**Logs**
Run the following from the container:
wget https://raw.githubusercontent.com/macvk/dnsleaktest/master/dnsleaktest.sh
chmod +x dnsleaktest.sh
./dnsleaktest.sh
You will see you have your VPN IP however the container is still using your router's DNS. It should be using the VPN's DNS.

**Host system:**
Ubuntu 18.04

haugene:
Any suggestion for a fix? Using Google DNS is widely communicated as a recommended setup, but there should probably be something we can do with OpenVPN here?

Another question. Is is using the DNS servers that it gets from the router. Which means that it is inherited from the host? The DNS packets themselves go through the VPN and then to the ISP DNS server?

RickBankers:
This may fix the issue. https://community.openvpn.net/openvpn/wiki/Pushing-DNS-to-clients

RickBankers:
I would also suggest adding a dnsleak test option to your build to inform users there is no dns leak. You can use the referenced dnsleak test I pointed to above. I would be nice to get that output when the container and vpn startup.

RickyXwang:
Normally Server will push the DNS config to the client. Not much can be done on the client-side unless there is some config issue that overwrites or ignored the push. 
Something like 
`PUSH: Received control message: 'PUSH_REPLY,redirect-gateway def1,dhcp-option DNS 10.160.0.1,comp-lzo no,route 10.160.0.1,topology net30,ping 10,ping-restart 60,ifconfig 10.160.0.150 10.160.0.149,peer-id 42,cipher AES-256-GCM'`
Can you check the docker log?

naythan90:
A while back, I stumbled on someone mentioning DNS leaks on here so I modified my docker file from the Google DNS to the PIA DNS. This seemed to work well at the time. 

After reading this post today, it looks like it's not working as well as it used to and I assume it's because of the updates which PIA are making to their environment around DNS:
https://www.privateinternetaccess.com/helpdesk/kb/articles/next-generation-dns-custom-configuration

[...]

PR #1614, opened 2020-12-18
Fix DNS leak by updating /etc/resolve.conf Fix for issue #1455

Xenov-X:
Uses OpenVPN scripts to update Resolv.conf at VPN start with DNS entries provided by OpenVPN server - only changes are to trigger the scripts to be run.

Ensures all DNS queries are routed via VPN provider, obscuring from ISPs

Issue ref  #1455


clement-z:
It seems there also is a `up.sh` and `down.sh` in `/etc/openvpn/` (from the distro as far as I know ?).

The files are different. Why use one over the other? A quick check showed me both work (using PIA) for setting it up; I didn't try stopping openvpn and checking.

clement-z:
Hmm @Xenov-X I think this is not what you meant to do ?

Xenov-X:
ahhhhhh - you're right - the extra changes were not intended - lost track of what branch I was on.

ultimately, from what I can see, there seems no difference in functionality between the two  - both achieve updating /etc/resolv.conf with the VPN providers DNS entries.

Will work out how I can remove these changes...

Xenov-X:
Going to close this Pull req - re fork and resubmit. It's a bit of a bombsite, I can't really work out whats rolled back.

clement-z:
Yes I think it's the easiest solution 😃

Issue #558, opened 2018-07-16
Not sure my traffic is going over the VPN


rdavey228:
Hello - 

I have setup a custom provider with Cyberghost and I can see the VPN is working and connected to a server in Switzerland

Running the following command inside the container gives me a swiss ip address. So far so good!

`curl -s checkip.dyndns.org | sed -e 's/.*Current IP Address: //' -e 's/<.*$//'`

I have installed traceroute tools inside the container to check the route of the traffic.

The first hop is always shown as my ISP rather than the swiss IP. The swiss IP comes in second after my ISP. This leads me to believe DNS lookups are going over the VPN and going out to my ISP first!

Can anyone confirm this?

coolboiime:
Did you specify the custom Google DNS?

rdavey228:
No I removed that line....

Should the first hop be my ISP or the VPN?

coolboiime:
The VPN should be connecting using Google’s DNS

haugene:
If you don't override the dns it probably gets this from the host which gets it from your router which gets it from your ISP. The requests will still go through the VPN but your ISPs DNS servers will be used, i guess.

rdavey228:
DHCP is set to use my internal DNS server which uses google as the DNS forwarders if it cant resolve it internally (which it obviously wont be able to for an external address)

I have changed the docker-compose file to use googles DNS any way as per coolboiime's comment. But my first hop is still a UK based address with the VPN being the next hop based in switzerland.

So this change made no difference......

coolboiime:
What's the UK based IP address? I'm sure Google's DNS is spread out through a varities of CDN nodes.

rdavey228:
Im running a trace route from inside the docker container

```
traceroute to bbc.co.uk (151.101.0.81), 30 hops max, 60 byte packets
 1  10.251.200.1 (10.251.200.1)  24.680 ms  24.611 ms  24.607 ms
 2  gw-104.datasource.ch (176.10.104.3)  103.595 ms  103.676 ms  103.817 ms
 3  hu-b69-10gigabit-slx9540.datasource.ch (91.201.56.132)  25.066 ms  25.190 ms  25.187 ms
 4  te0-2-1-3.rcr51.b021037-0.zrh02.atlas.cogentco.com (149.14.212.145)  26.708 ms  26.731 ms  26.742 ms
[...]

PR #1621, opened 2020-12-19
Fix DNS leak by updating /etc/resolve.conf

Xenov-X:
Add scripts to OpenVPN startup to update /etc/resolv.conf with VPN providers DNS servers, stopping DNS leakage. 

@clement-z already observed files to complete similar function are available in /etc/openvpn/. Visually they appear near identical (I haven't diff'd them), but both achieve the functionality desired.

This implementation uses:
/usr/share/openrc/support/openvpn/up.sh
/usr/share/openrc/support/openvpn/down.sh

However a valid alternative would be:
/etc/openvpn/up.sh
/etc/openvpn/down.sh

Only other point of note is setting DNS to correct values before transmission starts and after it closes.

Ref Issue #1455 
Prev pull req #1614 (cancelled because I can't manage my branches) (Sorry)

### DIFF
--- a/openvpn/tunnelDown.sh
+++ b/openvpn/tunnelDown.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
-
 /etc/transmission/stop.sh
+/usr/share/openrc/support/openvpn/down.sh
 [[ ! -f /opt/tinyproxy/stop.sh ]] || /opt/tinyproxy/stop.sh
--- a/openvpn/tunnelUp.sh
+++ b/openvpn/tunnelUp.sh
@@ -1,3 +1,4 @@
 #!/bin/bash
+/usr/share/openrc/support/openvpn/up.sh
 /etc/transmission/start.sh "$@"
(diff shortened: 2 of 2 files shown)


pkishino:
this causes issues as seen here #1667

clement-z:
I thought the down.sh was supposed to take care of this 🤔 

I am currently on mobile so I can't check but I will take a look.

Xenov-X:
Seems like somehow they've stopped openvpn without down.sh running. 

May need to use --down-restart and --up-restart too to account for openvpn process being killed & it not running the down action

clement-z:
So any news on that side ?

ntalbo:
I seem to have fixed the DNS leaks by modifying my tunnelUp.sh, as per https://github.com/haugene/docker-transmission-openvpn/issues/558#issuecomment-596058869. It seems to allow the Mullvad server to properly push the DNS to resolv.conf. No issues while restarting the container and it seems to stay connected while running.

Not sure whether this is a proper fix, I sadly lack the technical know-how to fully understand the problem. I'd be glad to help troubleshooting if needed.

PR #1632, opened 2020-12-22
Added additional configuration options to table

samr28:
I had a hard time finding a list of "transmission options" so I ended up looking through the set environment variables in the container to find the two that I was looking for (idle limits)

### DIFF
--- a/docs/config-options.md
+++ b/docs/config-options.md
@@ -66,12 +66,14 @@ You may override Transmission options by setting the appropriate environment var
 The environment variables are the same name as used in the transmission settings.json file
 and follow the format given in these examples:
 
-| Transmission variable name | Environment variable name             |
-| -------------------------- | ------------------------------------- |
-| `speed-limit-up`           | `TRANSMISSION_SPEED_LIMIT_UP`         |
-| `speed-limit-up-enabled`   | `TRANSMISSION_SPEED_LIMIT_UP_ENABLED` |
-| `ratio-limit`              | `TRANSMISSION_RATIO_LIMIT`            |
-| `ratio-limit-enabled`      | `TRANSMISSION_RATIO_LIMIT_ENABLED`    |
+| Transmission variable name   | Environment variable name                 |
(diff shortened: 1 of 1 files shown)


clement-z:
Maybe the right way to do it would be to make the explanation stand out more ? All transmission settings can be set from the environment (after converting `-` to `_`, making uppercase, and prefixing with `TRANSMISSION_`), and we are not going to list them all (it is not the job of this repo to keep a documented track of all transmission settings). The ones present in the table are, as mentioned just above it, an example.

What in your opinion should have been done to make this process stand out more to someone who looks for this information? What have you looked for? Maybe we can set a more explicit section title, etc.

samr28:
Well I think including a link to a list of all of the available configurations in transmission (if it exists) would be sufficient. I understood the idea of that part of the docs, just didn't know how to convert "stop seeding if idle for (mins)" into "idle-seeding-limit" if that makes sense.
image

clement-z:
Alright, well this is dependent on the UI you use (although it doesn't mean they don't all use the same labels).

The transmission daemon list can be found [here](https://github.com/transmission/transmission/wiki/Editing-Configuration-Files) (although it is not in a very pretty format). If you have the time, could you add a commit where you make the section text more explicit maybe, and add a link to the transmission doc?

samr28:
Oh I see, perhaps I just needed to read up more on how transmission actually works. I'll see what I can do to the wording tomorrow. Can I link the transmission wiki doc there too or should I leave that out?

clement-z:
Great ! And yes I think the link should be included as well. If you would have found it useful when looking for information, then it probably belongs there. The only thing which would be really out of scope would be duplicating work that really is/ought to be done by the transmission project itself. 😉 

Regarding how transmission works, well, it's quite simple really. The gist is that transmission spawns a daemon which is the main process, and frontends (plain command-line, Web, Qt, GTK...) which can be used to control this process.

samr28:
Just made those changes we discussed, lmk if there's anything else.

PR #1662, opened 2021-01-17
Remove perfect privacy provider

varac:
Perfect Privacy was founded and run by German/Austrian neo-Nazis,
some of them convited to jail sentence for engaging in neo-nazi activities
in Austria (https://en.wikipedia.org/wiki/Verbotsgesetz_1947).
PP has denied to answer recent questions about this fact. Neither it has
denied current relation to organised Nazi movements or the possibe continuation
of the companies extreme right-wing consistence.
Because of this it's removed by this commit.

See more details at:

* https://medium.com/@winstonsmith_/the-problem-with-vpns-c22c9c82cf13
* https://thevpnshop.com/reviews/perfect-privacy-vpn-review/
* https://securitygladiators.com/perfect-privacy/

Articles in german:

* https://www.derstandard.at/story/1336697911724/kuessel-prozess-perfect-privacy-wegen-verbindungen-zu-rechtsextremisten-im-zwielicht
* https://www.zeit.de/gesellschaft/zeitgeschehen/2012-09/neonazi-website-admin-robert-m
* https://taz.de/Rechtsextremer-im-Chaos-Computer-Club/!5089711/
* http://de.indymedia.org/2012/05/330386.shtml
* https://autonome-antifa.org/?article251


masterofmain:
I was wondering why my setup stopped working and this change is the reason why.

@varac @pkishino Firstly, I don't think it's right that you've made an immediate and breaking change without any wider discussion and without any notice that support for the service will be deprecated.

Secondly, I think it's quite dangerous to impose your own political and moral views and judgements on the people who use this software - especially when the justification for your action is based on rather weak evidence. (Basically, two people with Nazi views were historically involved with the company, and the company has chosen to make no further comment. Questionable, sure, but hardly compelling.)

By all means highlight the issues you believe exist with Perfect Privacy. By all means encourage people to use a different service. But don't prevent them from using a service that they have already paid for. And don't impose your will on other people.

Please revert this pull request.

ilike2burnthing:
_Disclaimer: I have sweet FA to do with this project other than being a user, so my comment here is my own. Also, this is the only comment I'll be making, so it isn't the start of some long internet 'debate'._

The money users paid was to PP, not this repo, the maintainers of which have no need to support any particular provider, let alone one with such questionable ties.

No one's will is being forced on anyone else; you can keep using an old build, or you can make your own fork, but the maintainers here will no longer put effort into supporting a service which won't explain why it definitely isn't Nazi-adjacent (something which would be in their interest and not exactly difficult, if it were true).

clement-z:
I think it's more complicated than that though. Removing without warnings configs for political/ideological reasons in such a breaking manner is indecent of an open source project manager.

We have been made aware of the problem, and we probably all feel the same way about this. Why not *schedule* removal for, e.g. the next point release release, while we still give users a possibility to agree/disagree with the decision in a related issue ticket?

Mave95:
I've forked the repo and reverted the removal of the Perfect Privacy support.
https://github.com/Mave95/docker-transmission-openvpn
https://hub.docker.com/r/mave95/transmission-openvpn

pkishino:
> I think it's more complicated than that though. Removing without warnings configs for political/ideological reasons in such a breaking manner is indecent of an open source project manager.
> 
> We have been made aware of the problem, and we probably all feel the same way about this. Why not _schedule_ removal for, e.g. the next point release release, while we still give users a possibility to agree/disagree with the decision in a related issue ticket?

Hello,
just a quick comment,
considering there is a clear connection to ILLEGAL and CRIMINAL (not political or ideological) activity of this provider by its founders there will be no further discussion around this.
anyone that still feels the need to support that kind of company is free to either look elsewhere or use the custom provider option

haugene:
https://github.com/haugene/vpn-configs-contrib/pull/2

PR #1669, opened 2021-01-19
Ovpn2.5

pkishino:
this is currently NOT building due to some strange tar problem between 1.32 and 1.33..been trying to fix this but no luck yet


wilmardo:
Superseeded by #1696 

When did you get the tar problem? I build the image for amd64 and had no issues whatsoever.

PR #696, opened 2019-02-04
Update Windscribe VPN configs

wilmardo:
Related: #554 

Scraped the configs from the WebUI with a dirty little scraper ([link](https://github.com/wilmardo/windsribe-ovpn-config-scraper)) since I couldn't find a zip download option anywhere. The steps I took are described in the README.
Then updated them with the adjustConfigs.sh

PR #1691, opened 2021-01-27
feat: update Windscribe configs

wilmardo:
Just like #696 I used and updated my little scraper:
https://github.com/wilmardo/windsribe-ovpn-config-scraper

I sadly can't port this script to grab the config at runtime since Windscribe hosts the config behind a login.


wilmardo:
Windscribe is currently broken since they updated all they urls to `*.whiskergalaxy.com`

PR #1696, opened 2021-02-02
feat: update to alpine 3.13

wilmardo:
Followup of #1669 

Currently running this with #1691 on my cluster, feel free to try from ` lansible/transmission-openvpn:develop`

More info on `npm ci` here:
https://docs.npmjs.com/cli/v6/commands/npm-ci#description

Backticks in bash are the 'old style', new way is `$()` so updated that too :)


pkishino:
great,
yeah, the tar problem was with the old way of building flood I think..
this will also close #1511

PR #1711, opened 2021-02-11
Mullvad: Updated us_all, added city-specific configs

jfessard:
- us_all seems to use UDP1300 when I downloaded a config today, so setting that as the new port
- Ports only work within a user-selected city. I need just one of them, but I added all US cities to be nice (And make sure that other people select the actual correct ones and not just the one I added, for load balancing.

## Tests
I tested both the new us_all (worked with UDP 1300) and a specific city (connected fine) and the port I had picked with Mullvad was marked "open" in transmission. (its **closed** if I pick `us_all` instead)


jfessard:
image
Mullvad screenshot

jfessard:
Are there any issues/concerns blocking this merge? (or just that the maintainers are busy, which is totally understandable)

pkishino:
We will not be updating or merging in any providers as we plan to completely remove all providers from the project and separate them to their own project, thus feel free to keep them updated here for others to pull and build

jfessard:
Thanks for the response! I'll keep an eye out for when the providers are spun out to a different repo.

Issue #781, opened 2019-04-17
RSS plugin not supporting HTTP authorization


YardaX:
Hi!
I'm trying to start the RSS plugin in a separate docker container. It works very well, however it seems I'm not able to add into the environment username & password when the transmission has the authentication enabled.

These variables are missing completely in the environment.

Did someone manage to run it with the authentication?

rubenvarela:
Seems like the [template file](https://github.com/haugene/docker-transmission-openvpn/blob/0e0f6786a41b504250fb055053f7ae1a7ac8acbc/plugins/rss/transmission-rss.tmpl) for the settings is not passing the value. 

It's missing the username and password entries. Here's an example - https://github.com/nning/transmission-rss/blob/8c13dd87ce23d35ded993e9ff475d610ff4b0a04/transmission-rss.conf.example#L52-L54

On `plugins/rss/transmission-rss.tmpl`  you should be able to add

```
login:
  username: {{ .Env.TRANSMISSION_RPC_USERNAME }}
  password: {{ .Env.TRANSMISSION_RPC_PASSWORD }}
```

YardaX:
Thanks! I will look into it.

mrjackyliang:
Looks solved! Closing for now. Let us know if there additional issues.

canzone:
Hello. I know this was closed long ago, but this issue still remains in the latest version of this docker container. Would you please add it to the latest version, so the fix we apply manually won't be overwritten when we pull upgrades from docker? Thanks a lot.

pkishino:
> Hello. I know this was closed long ago, but this issue still remains in the latest version of this docker container. Would you please add it to the latest version, so the fix we apply manually won't be overwritten when we pull upgrades from docker? Thanks a lot.

create a PR with the fix and I'll take a look

PR #1741, opened 2021-02-25
RSS plugin not supporting HTTP authorization

canzone:
This fork will fix the issue #781

### DIFF
--- a/plugins/rss/start.sh
+++ b/plugins/rss/start.sh
@@ -28,6 +28,12 @@ else
   else
     sed -i "s#regexp: placeholder#regexp: $rss_regex_esc#" /etc/transmission-rss.conf
   fi
+  if [[ $TRANSMISSION_RPC_ENABLED == 'true' ]] && [ -z $(grep login /etc/transmission-rss.conf) ]; then
+    echo "RPC enabled, adding login details to rss config as no login details exist"
+    printf "login:\n  username: $TRANSMISSION_RPC_USERNAME\n  password: $TRANSMISSION_RPC_PASSWORD" >>/etc/transmission-rss.conf
+  else
+    echo "Login already provided in config OR RPC does not seem to be enabled"
+  fi
 fi
 
 echo "Starting RSS plugin with the following config:"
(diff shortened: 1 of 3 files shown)


haugene:
This won't work. We've stopped using dockerize to render templates. Had issues getting stuff to work on arm devices. Need to find another template language to use expressions like that. And... If no username/password is provided the config block should not be there.

Right now we're using pure bash with sed to do substitutions. I guess you could do an if-check and append those lines if you find the username and password variables.

pkishino:
@canzone please update this if you want this merged

pkishino:
Hello,
Then you are running an old build, right?
Cause the templating doesn’t work in newer releases, so this needs to be
changed if it aims to be merged

pkishino:
I just noticed that one can mount their own transmission-rss.conf file where one could include the login details..
that would be sufficient
but I'm doing a small update to the script to add login if transmission_rpc_enabled is true

pkishino:
@canzone you could try my latest commit and see how it works

canzone:
Hello @pkishino ... I've just tested and it's working perfectly.

Thanks a lot!

plugins/rss/start.sh
```diff
@@ -28,6 +28,12 @@ else
   else
     sed -i "s#regexp: placeholder#regexp: $rss_regex_esc#" /etc/transmission-rss.conf
   fi
+  if [[ $TRANSMISSION_RPC_ENABLED == 'true' ]] && [ -z $(grep login /etc/transmission-rss.conf) ]; then
```
haugene, line 31: Should `TRANSMISSION_RPC_ENABLED` maybe be `TRANSMISSION_RPC_AUTHENTICATION_REQUIRED`?

That being said, we know we're contacting Transmission. And over RPC, right? There is no other way. So maybe the switch should just be `RPC_AUTH_ENABLED` and `RPC_USERNAME` + `RPC_PASSWORD`?

haugene: Or something like that.

pkishino: Hmm, I was checking inside my container and only had the TRANSMISSION_RPC_ENABLED
Available so that’s why I used it but feel free to verify and I can fix

Issue #1489, opened 2020-11-10
Separate providers into a 2nd repo


ksurl:
The linuxserver/swag repo (nginx reverse proxy with text based config) has a separate repo for sample reverse proxy config files that are downloaded and copied into the image at build and updated on every container start. This dissociates the changes to config files from the functionality of the core image. I think this principle will carry over nicely for the transmission image so vpn config changes won't need a new build of the container. PR and updates can be handled at the 2nd repo independently of the main repo. This is in line with simplifying the config maintenance with 3.x

What do you think?

haugene:
I like it :+1: Have been thinking a bit about a solution like that myself. See #1427 that also tries to address this.
It's a good idea to clone it on build time and update it on run time. But I have also been thinking that if we clone it on run time then we can also allow for people to use different repositories. That could be handy when doing a PR to the config-repo and be able to specify your own fork as the source.

That can also be done if we pre-clone as well. But I'm worried about size. Not that it matters too much but the image is today ~28MB and the repo is 100MB. So much historic config files I guess.

And lastly as mentioned in the other issue. It would be nice if this could be set up in a way with the codeowners stuff from Github to distribute ownership.

ksurl:
I guess we could start if you create a repo for the configs (something like haugene/docker-vpn-configs or providers, keeping it agnostic since WireGuard is on the roadmap as well. Though that may be a little different since each conf file has a unique private key? Unless WireGuard supports merging configs since it would have a common server fqdn and psk).

Then work can begin on a branch that downloads it on build and boot. wget or curl with tar and gzip can work so we don't have git files laying around.

As for custom repo, we could use another environment variable that defaults to the main one.

haugene:
Yes. That sounds good. I was about to name it "openvpn-configs" and didn't think of wireguard. could have a "wireguard-configs" repo as well I guess. What do you think about git history? Should we start from scratch, or make it like a fork of the main repo and just re-structure it so that we keep a history of who has maintained a provider in the past as well?

ksurl:
Should it be separated by provider or by vpn type? Do they have to be separate repo's?

That sounds good. I didn't think of that, who has been submitting PR for various providers.

haugene:
https://github.com/haugene/vpn-configs

I started here. Has the complete history, but completely wiped for anything that's not .ovpn or txt, crt, ca, key files.
[...]

PR #1744, opened 2021-02-26
Moving configs to separate repo

haugene:
As discussed in #1489 and that we have been planning for a while. We should move all the .ovpn config files to a separate repo to make a clearer distinction between config and code. That would reduce the amount of PRs and issues posted here and we could set up a codeowner-structure or otherwise engage more of the community to maintain configs.

Separating the configs like this also enables people to fork, edit and test new configs without getting their PRs merged. While many of our contributors are familiar with git, this would allow any user to fork the config repo and copy-paste a config from his/her provider and be up and running.

The PR is not done. But it might not be far from it. I'm wondering what you think about naming of the variables and what we'll call these "modes" of operation. I'm not sure that we'll forever say that scripts live here, configs live there, so there should be some flexibility. For now this will default to "business as usual" and you have to say "static" config to make it download it. So it's opt-in.

As a final solution I'm up for suggestions, but I'm thinking that the default mode should be "auto" where we check if there are config scripts in the container for the given provider. If so we assume that they should be used to fetch/configure the provider. If we can't find them then we download the zip and look there. And then we fail if we can't find it.

If someone wanted to run their own scripts, they could then say "static" and point to their own repo and their scripts would be run instead. Calling it "static" is therefore a bit misleading for that scenario.

Anyways. I must be off for tonight! To be continued...

### DIFF
--- a/openvpn/fetch-external-configs.sh
+++ b/openvpn/fetch-external-configs.sh
@@ -1,0 +1,42 @@
+#! /bin/bash
+
+VPN_CONFIG_SOURCE_TYPE="${VPN_CONFIG_SOURCE_TYPE:-github_zip}"
+
+# Set default GitHub config repo
+GITHUB_CONFIG_SOURCE_REPO="${GITHUB_CONFIG_SOURCE_REPO:-haugene/vpn-configs-contrib}"
+GITHUB_CONFIG_SOURCE_REVISION="${GITHUB_CONFIG_SOURCE_REVISION:-main}"
+
+if [[ "${VPN_CONFIG_SOURCE_TYPE}" == "github_zip" ]]; then
+
+  # Concatenate URL for config bundle from the given GitHub repo
+  GITHUB_CONFIG_BUNDLE_URL="https://github.com/${GITHUB_CONFIG_SOURCE_REPO}/archive/${GITHUB_CONFIG_SOURCE_REVISION}.zip"
+  
+  # Create a temporary file and download bundle to it
+  config_repo_temp_zip_file=$(mktemp)
+  echo "Downloading configs from ${GITHUB_CONFIG_BUNDLE_URL} into ${config_repo_temp_zip_file}"
+  curl -sSL -o "${config_repo_temp_zip_file}" "${GITHUB_CONFIG_BUNDLE_URL}"
+
+  # Create a temporary folder and extract configs there
+  config_repo_temp_dir=$(mktemp -d)
+  echo "Extracting configs to ${config_repo_temp_dir}"
+  unzip -q "${config_repo_temp_zip_file}" -d "${config_repo_temp_dir}"
+
+  # Find the specified provider folder. Should be under <tmpDir>/<some-root-folder>/openvpn/<provider>
+  provider_configs=$(find "${config_repo_temp_dir}"/*/openvpn -type d -name "${VPN_PROVIDER}")
+  if [[ -z "${provider_configs}" ]]; then
+    echo "ERROR: Could not find any configs for provider ${VPN_PROVIDER^^} in downloaded configs"
+    exit 1
+  fi
+
+  # Replace current provider home folder with the downloaded directory
+  echo "Found configs for ${VPN_PROVIDER^^} in ${provider_configs}, will replace current content in ${VPN_PROVIDER_HOME}"
+  rm -r "${VPN_PROVIDER_HOME}"
+  mv "${provider_configs}" "${VPN_PROVIDER_HOME}"
+
+  # Clean up temporary files
+  rm -rf "${config_repo_temp_zip_file}" "${config_repo_temp_dir}"
+
+else
+    "ERROR: VPN config source type ${VPN_CONFIG_SOURCE_TYPE} does not exist..."
+    exit 1
+fi
--- a/openvpn/fetch-external-configs.sh
+++ b/openvpn/fetch-external-configs.sh
@@ -24,7 +24,7 @@ if [[ "${VPN_CONFIG_SOURCE_TYPE}" == "github_zip" ]]; then
   # Create a temporary file and download bundle to it
   config_repo_temp_zip_file=$(mktemp)
   echo "Downloading configs from ${GITHUB_CONFIG_BUNDLE_URL} into ${config_repo_temp_zip_file}"
-  curl -sSL -o "${config_repo_temp_zip_file}" "${GITHUB_CONFIG_BUNDLE_URL}"
+  curl -sSL --fail -o "${config_repo_temp_zip_file}" "${GITHUB_CONFIG_BUNDLE_URL}"
 
   # Create a temporary folder and extract configs there
   config_repo_temp_dir=$(mktemp -d)
--- a/openvpn/fetch-external-configs.sh
+++ b/openvpn/fetch-external-configs.sh
@@ -1,5 +1,9 @@
 #! /bin/bash
 
+set -o nounset
+set -o errexit
+set -o pipefail
+
 VPN_CONFIG_SOURCE_TYPE="${VPN_CONFIG_SOURCE_TYPE:-github_zip}"
 
 # Set default GitHub config repo
@@ -7,6 +11,12 @@ GITHUB_CONFIG_SOURCE_REPO="${GITHUB_CONFIG_SOURCE_REPO:-haugene/vpn-configs-cont
 GITHUB_CONFIG_SOURCE_REVISION="${GITHUB_CONFIG_SOURCE_REVISION:-main}"
 
 if [[ "${VPN_CONFIG_SOURCE_TYPE}" == "github_zip" ]]; then
+
+  function cleanup {
+    echo "Cleanup: deleting ${config_repo_temp_zip_file} and ${config_repo_temp_dir}"
+    rm -rf "${config_repo_temp_zip_file}" "${config_repo_temp_dir}"
+  }
+  trap cleanup EXIT
 
   # Concatenate URL for config bundle from the given GitHub repo
   GITHUB_CONFIG_BUNDLE_URL="https://github.com/${GITHUB_CONFIG_SOURCE_REPO}/archive/${GITHUB_CONFIG_SOURCE_REVISION}.zip"
@@ -33,8 +43,7 @@ if [[ "${VPN_CONFIG_SOURCE_TYPE}" == "github_zip" ]]; then
   rm -r "${VPN_PROVIDER_HOME}"
   mv "${provider_configs}" "${VPN_PROVIDER_HOME}"
 
-  # Clean up temporary files
-  rm -rf "${config_repo_temp_zip_file}" "${config_repo_temp_dir}"
+  exit 0
 
 else
     "ERROR: VPN config source type ${VPN_CONFIG_SOURCE_TYPE} does not exist..."
--- a/openvpn/start.sh
+++ b/openvpn/start.sh
@@ -46,8 +46,8 @@ fi
 
 # If no OPENVPN_PROVIDER is given, we default to "custom" provider.
 VPN_PROVIDER="${OPENVPN_PROVIDER:-custom}"
-VPN_PROVIDER="${VPN_PROVIDER,,}" # to lowercase
-VPN_PROVIDER_HOME="/etc/openvpn/${VPN_PROVIDER}"
+export VPN_PROVIDER="${VPN_PROVIDER,,}" # to lowercase
+export VPN_PROVIDER_HOME="/etc/openvpn/${VPN_PROVIDER}"
 mkdir -p "$VPN_PROVIDER_HOME"
 
 # Make sure that we have enough information to start OpenVPN
--- a/openvpn/start.sh
+++ b/openvpn/start.sh
@@ -59,19 +59,43 @@ fi
 echo "Using OpenVPN provider: ${VPN_PROVIDER^^}"
 
 if [[ -n $OPENVPN_CONFIG_URL ]]; then
-  echo "Found URL to OpenVPN config, will download it."
+  echo "Found URL to single OpenVPN config, will download and use it."
   CHOSEN_OPENVPN_CONFIG=$VPN_PROVIDER_HOME/downloaded_config.ovpn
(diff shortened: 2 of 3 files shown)


ksurl:
how about checking if the static config already exists locally (already downloaded before container restart)?

also partial git download to target a subfolder. I think there was a web app mentioned that requires hosting to enable as a middleware for this functionality. maybe a plugin for later versions.

haugene:
Good point. Should probably have the option to avoid downloading on every restart. :+1: But if you've targeted a branch then I think many want to get the latest version of that branch, so it'd be a opt-in to not download every time I would say. Or maybe we can store a hash and check if there are newer commits :thinking: 

About partial downloads. It sounds nice, but it would have to be both super simple and reliable to maybe be worth it imho. In the end, what are we saving? Megabytes on the download?

ksurl:
I think after cleaning up a few providers, the folder has shrank a lot. may not be an issue (it's under 100mb now).

haugene:
The vpn-configs repo zip is 23MB now. That's a benefit of doing zip download instead of git clone as well. We only get one revision and not the complete history, plus .ovpn files are quite zip-friendly. Lots of duplication -> good compression.

ksurl:
`git clone --depth=1` would also allow efficient git pulls on restarts if the local folder is already detected.

fetzu:
Just as a few thoughts: 

1. Why not have the script download only the config files for the set provider instead of all the configs? Maybe this could be achieved through a partial git pull on said subfolder.

2. How do you plan to handle providers that have a "dynamic" script (ie: configs pulled through an API)? Should the "static" configs still be allowed in the repo or should the repo be reserved exclusively for providers which don't have a "dynamic" config option?

3. With the planned feature of WireGuard implantation, should we already consider this with this PR ? AFAIK WireGuard configs would mostly be dynamic because of the key exchange; which means that the script would only be called in case OpenVPN is used/set.

(Sorry in advance if my remarks are out of bounds/useless..!)

haugene:
All input is appreciated @fetzu. And you make good points. Here are my remarks on a per-item basis.

1. This comes down to simplicity. Partial pulls on sub-folders sounds great, but the priority should be robustness. I've maintained this project for so long that I know that anything that can fail, will fail at some point. So I'm not making any arguments against it, other than - why not download it all? It's 23 MB of files where everything you don't need gets deleted later. If this was a IoT project on mobile networks then 23 MB could matter, but we're talking torrenting here. 23 MB is negligible. But I'm all for it as long as it's simple, clean and elegant. Just want to note [for example this commit](https://github.com/haugene/docker-transmission-openvpn/commit/e81b8fc293a1a86ef6f19e1e49e009bfe7f3a0a0) where we had issues with git cloning on arm devices. This is in the node build of a frontend and not in the main container, but you get my point. 
2. As I mention in the original description "static" and "dynamic" might be misleading terms. "internal" and "external" might be better. I had a thought that we would still keep scripts in the main repo. So any openvpn provider where we have automated download and extraction of config files should still be managed here. That way we can re-use common parts of the scripts across providers. The idea was to only extract ".ovpn" files to a separate repo to avoid tons of lines of code and many commits just updating static information in text files. But with the proposed setup nothing would stop you from adding scripts in the external repo and it would be run as if it was internal. The externally downloaded provider folder is just hot swapped in place.
3. I think we should plan for the WireGuard feature as well, at least a bit. I'm not a fan of trying to think our way to an optimal setup and how this will be when we have WireGuard. Then it's better to code something now and iterate on it in the future. But discussing it a bit now and making some right key decisions so that it won't bite us in the ass later is good. Always a hard balance to strike. Again. Dynamic/static might not be the right distinction for WireGuard either. I still have no experience in using WireGuard - but I assume there are config files here as well? Or would it be easier to implement it as scripts?

As an overall note. I'm pretty sure that we will not get the perfect solution on the first try on any of the points. And I don't even think we should aim for it, because it's always about improving and there is no one size fits all here. I want a working system we can improve, but where the core assumptions and mechanisms are relatively correct. If we fetch the configs by curl, git or crawling a specific twitter-feed is not my primary concern. That's just a building block we can exchange later.

haugene:
Updated the PR now and moved the download of the external configs to a separate script. For now it still defaults to downloading a zip from GitHub but we can have multiple configurable options. I've also set MODIFY_CHOSEN_CONFIG to be true by default. I think that most of the provider configs should be able to be added without any modifications and then they should be automatically modified to match the project setup. For some providers there might be need for custom tweaking, but they can then disable it.

Any thoughts @pkishino?

And. When we come to the point of merging this. We have a bunch of open PRs now for updated configs. Any suggestion on the way forward? We could close all the PRs and refer to the config-repo. But we could also merge them all and tag a 3.5 version of the image, and then merge this PR where we delete all the configs. Then I could sync the config-repo from the 3.5 tag and we would get all the PR changes over into the config repo.

ksurl:
I would suggest merging existing pr for config update and set a cutoff for future pr to be directed to the new repo. Easier than chasing down all the open pr to open a new one on the new repo

pkishino:
I agree, I held off on mergind the other PR so we could all agree an this, but I think that a 3.5 with all updated configs will be best to move forward.
then a 4.0 would be the aim for the split

openvpn/fetch-external-configs.sh
```diff
@@ -0,0 +1,42 @@
...
+
+if [[ "${VPN_CONFIG_SOURCE_TYPE}" == "github_zip" ]]; then
+
+  # Concatenate URL for config bundle from the given GitHub repo
+  GITHUB_CONFIG_BUNDLE_URL="https://github.com/${GITHUB_CONFIG_SOURCE_REPO}/archive/${GITHUB_CONFIG_SOURCE_REVISION}.zip"
+  
+  # Create a temporary file and download bundle to it
+  config_repo_temp_zip_file=$(mktemp)
+  echo "Downloading configs from ${GITHUB_CONFIG_BUNDLE_URL} into ${config_repo_temp_zip_file}"
+  curl -sSL -o "${config_repo_temp_zip_file}" "${GITHUB_CONFIG_BUNDLE_URL}"
```
amosshapira, line 17: missing check of curl success

openvpn/fetch-external-configs.sh
```diff
@@ -0,0 +1,42 @@
...
+
+  # Create a temporary folder and extract configs there
+  config_repo_temp_dir=$(mktemp -d)
+  echo "Extracting configs to ${config_repo_temp_dir}"
+  unzip -q "${config_repo_temp_zip_file}" -d "${config_repo_temp_dir}"
```
amosshapira, line 22: missing check of unzip success

openvpn/fetch-external-configs.sh
```diff
@@ -0,0 +1,42 @@
...
+    exit 1
+  fi
+
+  # Replace current provider home folder with the downloaded directory
+  echo "Found configs for ${VPN_PROVIDER^^} in ${provider_configs}, will replace current content in ${VPN_PROVIDER_HOME}"
+  rm -r "${VPN_PROVIDER_HOME}"
+  mv "${provider_configs}" "${VPN_PROVIDER_HOME}"
+
+  # Clean up temporary files
+  rm -rf "${config_repo_temp_zip_file}" "${config_repo_temp_dir}"
```
amosshapira, line 37: Move this to run in a `trap`?

openvpn/fetch-external-configs.sh
```diff
@@ -0,0 +1,42 @@
...
+
```
amosshapira, line 38: missing explicit `exit 0`

openvpn/start.sh
```diff
@@ -59,19 +59,43 @@ fi
...
-  # Preserve $PWD in case it changes when sourcing the script
-  pushd -n "$PWD" > /dev/null
-  # shellcheck source=/dev/null
-  . "$VPN_PROVIDER_HOME"/configure-openvpn.sh
-  # Restore previous PWD
-  popd > /dev/null
-  MODIFY_CHOSEN_CONFIG=yeah
+fi
+
+if [[ -z ${CHOSEN_OPENVPN_CONFIG} ]]; then
```
amosshapira, line 67: missing double quotes

openvpn/start.sh
```diff
@@ -59,19 +59,43 @@ fi
...
+      VPN_CONFIG_SOURCE=external
+    fi
+  fi
+
+  if [[ "${VPN_CONFIG_SOURCE}" == "external" ]]; then
+    # shellcheck source=openvpn/fetch-external-configs.sh
+    . /etc/openvpn/fetch-external-configs.sh
+  fi
+
+  if [[ -x $VPN_PROVIDER_HOME/configure-openvpn.sh ]]; then
```
amosshapira, line 90: missing double quotes

openvpn/start.sh
```diff
@@ -59,19 +59,43 @@ fi
...
+    # Preserve $PWD in case it changes when sourcing the script
+    pushd -n "$PWD" > /dev/null
+    # shellcheck source=/dev/null
+    . "$VPN_PROVIDER_HOME"/configure-openvpn.sh
+    # Restore previous PWD
+    popd > /dev/null
+  fi
 fi
 
 if [[ -z ${CHOSEN_OPENVPN_CONFIG} ]]; then
```
amosshapira, line 101: missing double quotes

openvpn/fetch-external-configs.sh
haugene: good idea

openvpn/fetch-external-configs.sh
haugene: what are you thinking here? do a `if [ $? -ne 0 ]; then` something? Or postfix the line with `|| exit 1` maybe? Or we could do a `set -e` to fail on any error? But we source the script from start.sh so setting -e might then leak into that script?

openvpn/start.sh
haugene: I know that openvpn/start.sh has a bit of syntactical issues. Have run shellcheck on it and I'll do a fixup later. Don't want to bundle it with this change in case I have to roll back.

Either way. For the `-z` empty checks it doesn't seem to mind about skipping the quotes. Maybe because if the string is empty there will be no word splitting? And if it's not empty it doesn't really matter if the string is split because it's not empty anyways?

openvpn/fetch-external-configs.sh
amosshapira: As much as I dislike `set -e`, I've grown used to our team's standard shell script top of:

```
set -o nounset
set -o errexit
set -o pipefail
```

It might be a matter of taste, but that's why I don't like sourcing one script from another in this type of code, it's a bit like Basic `GOSUB` if you remember that. (https://hwiegman.home.xs4all.nl/gw-man/GOSUB.html) - it makes every variable a global variable.

Does this script HAVE to be sourced (with bash `source`/`.`) in order to set any variables or change other bash internal state (e.g. `cd ...`), or can it be executed as a sub-process?

openvpn/start.sh
amosshapira: I stand corrected. The need to surround parameters for `test -z` must be an anachronism.

openvpn/fetch-external-configs.sh
haugene: Good question. I'm used to sourcing them in this project because the sub scripts usually set environment variables. But I don't think we would need that here. So then add that header and do the cleanup in a trap - the end?

I've gotten some bash experience over the years and I'm capable. But I'm not very well versed in best practices. So I'm happy to get some pointers :+1:

haugene:
Phew. Had to do some rebase magic here. GitHub did not appreciate the commit removing all the ovpn configs and related files.
This PR was very slow to load and it was impossible to see the actual diff. I've now committed the *.ovpn (++) removal to a separate branch `feature/remove-ovpn-config-files` and created a PR for that in #1769.

This PR will then target that branch so that we can review and discuss the actual code changes.

openvpn/fetch-external-configs.sh
amosshapira: I prefer passing parameters to sub scripts instead of having them rely on environment variables (even if it means passing the environment variable as a parameter). This makes the sub script call more explicit and readable as well as makes running the sub script independently easier (e.g. while developing, or if you find that you want to use it from another point in the code).

openvpn/fetch-external-configs.sh
haugene: :heavy_check_mark:

openvpn/fetch-external-configs.sh
haugene: :heavy_check_mark:

openvpn/fetch-external-configs.sh
haugene: I set
```
set -o nounset
set -o errexit
set -o pipefail
```
in the script as proposed. it now fails if the unzip command fails.

openvpn/fetch-external-configs.sh
haugene: I'm planning a bigger rewrite at some point. Been coding a bit more Python lately and I think that a lot of the code here could be done cleaner and easier with Python. To take this one step at a time I don't think I'll get away from the env vars any time soon. The project is too reliant on them already. So I'll let it be for this script as well.

But I've added
```
set -o nounset
set -o errexit
set -o pipefail
```
as suggested. and also added `errexit` and `pipefail` to the main script. I'm not sure if it can stay there. We'll have to see how much issues it creates. There are steps that I've seen failing in many logs where the container manages without. Like when people mount /dev/net/tun and don't set CREATE_TUN_DEVICE=false. Then we get a "file exists" type of error but it's fine.

It is probably better to fail fast and add ignores for those cases where a failure is accepted or tolerated in certain scenarios. I'm marking this as resolved for now.

If you have time for another review as it stands now I'd be grateful. But I think it's closing in on merging to `dev` at least. And planning to do some hardening there. And updating documentation etc.

haugene:
what do you think @pkishino? Getting ready to merge to dev and start a small rewrite of the documentation?

pkishino:
yeah, finally some cleanup here :)
I'll proably be quite busy in the next week or two but try to look in from time to time and see what I can do

pkishino (approved):

Issue #1758, opened 2021-03-06
Privoxy alternative in place of tinyproxy


jay-to-the-dee:
### Is your feature request related to a problem? Please describe.
This is a `privoxy` alternative for the HTTP proxy in place of `tinyproxy`. I did this because you then get ad blocking and privacy enhancing features etc. that tinyproxy lacks :) As well as more control with using the web driven privoxy rules what you then do and don't permit via the built in HTTP proxy.

### Describe the solution you'd like
https://github.com/jay-to-the-dee/docker-transmission-openvpn/tree/privoxy is my working implementation of this (based off current master branch) if you want to give it a try and see it working for yourself! Maybe could be mergd into `dev` or `beta` at some point?

### Describe alternatives you've considered
Squid may well give better performance in some configs because of the superior caching. But I think privoxy is easier to set up, has a smaller footprint and less steep learning curve overall for now! :)

### Additional context
It's optional, but I find this useful and I got it working for myself, so why not share? 😁

jay-to-the-dee:
We can also do HTTP header modifciation very easily with Privoxy which I noticed the current TinyProxy config has as an option.

pkishino:
I think this is a valid enhancement request, why not create a PR for this for review ?

jay-to-the-dee:
Will do :)

jay-to-the-dee:
https://github.com/haugene/docker-transmission-openvpn/pull/1760 PR here for further discussion :)

PR #1760, opened 2021-03-08
Privoxy in place of tinyproxy #1758

jay-to-the-dee:


Dockerfile
```diff
@@ -41,6 +53,7 @@ ADD openvpn/ /etc/openvpn/
 ADD transmission/ /etc/transmission/
 ADD tinyproxy /opt/tinyproxy/
```
pkishino, line 54: @haugene should we still keep tinyproxy as well?

Dockerfile
```diff
@@ -31,7 +33,17 @@ RUN echo "@community http://dl-cdn.alpinelinux.org/alpine/edge/community" >> /et
     && rm -rf /tmp/* /var/tmp/* \
     && groupmod -g 1000 users \
     && useradd -u 911 -U -d /config -s /bin/false abc \
-    && usermod -G users abc
+    && usermod -G users abc \
+    && addgroup -S privoxy && adduser -S privoxy -G privoxy \
+    && mkdir /opt/privoxy \
+    && curl -sL https://www.privoxy.org/sf-download-mirror/Sources/3.0.29%20%28stable%29/privoxy-3.0.29-stable-src.tar.gz | tar -C /opt/privoxy --strip-components=2 -xz \
+    # && git clone https://www.privoxy.org/git/privoxy.git /opt/privoxy \
```
pkishino, line 40: comment should be removed

Dockerfile
```diff
@@ -31,7 +33,17 @@ RUN echo "@community http://dl-cdn.alpinelinux.org/alpine/edge/community" >> /et
     && rm -rf /tmp/* /var/tmp/* \
     && groupmod -g 1000 users \
     && useradd -u 911 -U -d /config -s /bin/false abc \
-    && usermod -G users abc
+    && usermod -G users abc \
+    && addgroup -S privoxy && adduser -S privoxy -G privoxy \
+    && mkdir /opt/privoxy \
+    && curl -sL https://www.privoxy.org/sf-download-mirror/Sources/3.0.29%20%28stable%29/privoxy-3.0.29-stable-src.tar.gz | tar -C /opt/privoxy --strip-components=2 -xz \
```
pkishino, line 39: Hmm, I checked and it's a bit inconvenient that they don't have a "latest" tag or such to use for here..

Dockerfile
```diff
@@ -79,5 +92,10 @@ LABEL org.opencontainers.image.revision=$REVISION
 LABEL autoheal=true
 
 # Expose port and run
+
+#Transmission-RPC
 EXPOSE 9091
+# Privoxy
+EXPOSE 8118
```
pkishino, line 99: this should be optional depending on if it's enabled or not

tinyproxy/start.sh
```diff
@@ -46,33 +46,34 @@ set_authentication()
 
 if [[ "${WEBPROXY_ENABLED}" = "true" ]]; then
 
-  echo "STARTING TINYPROXY"
+  echo "STARTING PRIVOXY"
 
-  find_proxy_conf
-  echo "Found config file $PROXY_CONF, updating settings."
+  # find_proxy_conf
```
pkishino, line 51: remove commented out code please

tinyproxy/start.sh
```diff
@@ -46,33 +46,34 @@ set_authentication()
...
+  # echo "Found config file $PROXY_CONF, updating settings."
 
-  set_port ${WEBPROXY_PORT} ${PROXY_CONF}
+  # set_port ${WEBPROXY_PORT} ${PROXY_CONF}
```
pkishino, line 54: configuration should still be possible, no?

pkishino (changes requested):
A number of items still remaining..commented out code, references to tinyproxy @haugene  I would like your view on this, if we should drop this
also, documentation updates necessary etc
how to config etc..

pkishino:
I also changed the base to dev, please rebase and make sure only your commits are included in the PR

Dockerfile
amosshapira: I expect this comment is a blocker because the shell doesn't interpret the `\` at the end of the line as a continuation.

amosshapira:
Can you add a description of this PR?

Dockerfile
amosshapira: Yep. The closest I could find to automatically becoming aware of Privoxy releases is to parse their release feed from here https://www.privoxy.org/feeds/privoxy-releases.xml 

Here is a one-liner to extract the latest source URL from the RSS feed and give it to `curl` to download from:

```suggestion
    && curl -sL $(curl -s https://www.privoxy.org/feeds/privoxy-releases.xml | egrep -o -m 1 'https:\/\/www\.privoxy\.org\/sf-download-mirror\/Sources\/[0-9]+\.[0-9]+\.[0-9]+%20\(stable\)\/privoxy-[0-9]+\.[0-9]+\.[0-9]+-stable-src\.tar\.gz<' | sed -n -e 's/<$//' -e 1p) | tar -C /opt/privoxy --strip-components=2 -xz \
```

And BTW - 3.0.29 is already 3 releases behind the current latest as of today: https://www.privoxy.org/sf-download-mirror/Sources/3.0.32%20%28stable%29/

Dockerfile
```diff
@@ -31,7 +33,17 @@ RUN echo "@community http://dl-cdn.alpinelinux.org/alpine/edge/community" >> /et
...
+    && addgroup -S privoxy && adduser -S privoxy -G privoxy \
+    && mkdir /opt/privoxy \
+    && curl -sL https://www.privoxy.org/sf-download-mirror/Sources/3.0.29%20%28stable%29/privoxy-3.0.29-stable-src.tar.gz | tar -C /opt/privoxy --strip-components=2 -xz \
+    # && git clone https://www.privoxy.org/git/privoxy.git /opt/privoxy \
+    && cd /opt/privoxy \
+    && autoheader \
+    && autoconf \
+    && ./configure --enable-compression --with-brotli  --with-mbedtls --enable-extended-statistics  \
+    && make -j4 \
+    && make install
```
amosshapira, line 46: `make install-strip` will also strip symbols from the binary, making it a bit smaller.

Ref: https://www.privoxy.org/gitweb/?p=privoxy.git;a=blob;f=GNUmakefile.in;h=fcd44c4f449897df249686c034466a1da031e215;hb=HEAD#l841

amosshapira (changes requested):
1. Suggestion to fetch the latest privoxy source from the RSS feed.
2. The comment in the middle of a `&&` chain breaks it.

haugene:
> A number of items still remaining..commented out code, references to tinyproxy @haugene I would like your view on this, if we should drop this
> also, documentation updates necessary etc
> how to config etc..

I agree on the comments to code and structure. The PR also needs to be rebased as it's now affecting a lot of files that I don't think has anything to do with this. A documentation revamp would be great to go with this. We've had a proxy for a long time but not a good section on how/why we use it etc. Also curious for a description of this PR itself. I guess the suggestion is to change tinyproxy for privoxy because it is better, but I'm not familiar with any of the solutions and would like a summary or pitch somehow. Also if this would be relatively backwards compatible or not?

ilike2burnthing:
Brief explanation here - https://github.com/haugene/docker-transmission-openvpn/issues/1758

pkishino:
@jay-to-the-dee I hope you have not abandoned this and can help address the remaining issues and comments, hope to get this in as it def. looks like an improvement

Dockerfile
jay-to-the-dee: 3.0.29 is intentional :) I found later releases crash after a few hours but 3.0.29 does not have this issue :)

jay-to-the-dee:
Hi @pkishino not abandoned, just been very busy IRL this last week and wanted to let the discussion run it's course first :) I will try and remerge against `dev` branch later with the suggested changes (truthfully I'm more familiar with `stable` branch over `dev` at present)

Dockerfile
jay-to-the-dee: Worked fine for me but I appreciate may not for others so removed :)

Dockerfile
jay-to-the-dee: @pkishino Keeping TinyProxy or not will come down to how much the HTTP auth feature means to you/ other users (`WEBPROXY_USERNAME` and `WEBPROXY_PASSWORD`) This can be supported by setting Privoxy to go upstream to TinyProxy, in which case we would need both services. Otherwise Privoxy can run fine without TinyProxy, which is how I run my instance (and tunnels through the VPN just fine in the same way TinyProxy does).

This is because Privoxy does not contain built in support for HTTP proxy auth. (see https://www.privoxy.org/user-manual/config.html 7.4.9) and you would need to use the `enable-proxy-authentication-forwarding` feature of Privoxy as a part of sending the HTTP auth. provision upstream (to TinyProxy). Personally I'm fine without HTTP auth. which isn't that secure anyway and Privoxy instead supports acccess control by IP address via `permit-access` and `deny-access` directives, if you wanted to do that at the Privoxy level.

As for how to config, it is all done in text files so things can continue in a similair fashion to the present `sed`-ing methods, but I am not an expert at using `sed` so would appreciate some help there :)

Dockerfile
jay-to-the-dee: Agreed, but I don't believe`Dockerfile` supports conditional EXPOSE keyword ? (Correct me if I'm wrong :) )

Dockerfile
jay-to-the-dee: `install-strip` I have added, thanks for the great idea :) Once Privoxy becomes stable past 3.0.29 again, I will implement your XML parser 👍

tinyproxy/start.sh
jay-to-the-dee: Fixed in https://github.com/haugene/docker-transmission-openvpn/pull/1760/commits/0e3391120b0898d3588138a3608264d8eed3b503

Dockerfile
jay-to-the-dee: 3.0.29 is my cherry picked release for now as I found later 'stable' versions crashing after a few hours of use. :) Would love to use your one liner in future though!

amosshapira:
Sorry I feel like a dick for not mentioning these before:

Consider using Docker multi-stage to build the `privoxy` binary in a separate "build" image. That way you don't include the whole build chain in the final image. That's good both for size and security.

Ref: https://docs.docker.com/develop/develop-images/multistage-build/

Dockerfile
amosshapira: `EXPOSE` is more of a documentation of the image rather than an actual command. The port is determined during run time.

Ref: https://docs.docker.com/engine/reference/builder/#expose

openvpn/tunnelUp.sh
```diff
@@ -40,6 +40,6 @@ if [ "${PEER_DNS}" != "no" ]; then
 fi
 
 /etc/transmission/start.sh "$@"
-[[ ! -f /opt/tinyproxy/start.sh ]] || /opt/tinyproxy/start.sh
+[[ ! -f /opt/privoxy/start.sh ]] || /opt/privoxy/start.sh
```
amosshapira, line 43: I know you just updated paths here, but I think this line will be much clearer with the reversed condition.
Also the test doesn't check if the file is executable or not:
```suggestion
[[ -f /opt/privoxy/start.sh && -x /opt/privoxy/start.sh ]] && /opt/privoxy/start.sh
```

privoxy/scripts/start.sh
```diff
@@ -0,0 +1,55 @@
...
+    echo "Port [$1]: Not a number" >&2; exit 1
+  fi
+
+  # Port: Specify the port which privoxy will listen on.  Please note
+  # that should you choose to run on a port lower than 1024 you will need
+  # to start privoxy using root.
+
+  if test $1 -lt 1024
+  then
+    echo "privoxy: $1 is lower than 1024. Ports below 1024 are not permitted.";
```
amosshapira, line 33: Why is that? Inside the container, it might be running as the container's root.

privoxy/scripts/start.sh
```diff
@@ -0,0 +1,55 @@
...
+
+  echo "Setting privoxy port to $1";
+  sed -i -e"s,^listen-address .*,listen-address 0.0.0.0:$1," $2
+}
+
+if [[ "${WEBPROXY_ENABLED}" = "true" ]]; then
+
+  echo "STARTING PRIVOXY"
+
+    find_proxy_conf
```
amosshapira, line 45: weird indentation.

privoxy/scripts/start.sh
jay-to-the-dee: It's the exact same code that tinyproxy run under (i.e. not mine). But it's a pretty standard Unix thing, but can simpy remove the check if you want?

jay-to-the-dee:
> Sorry I feel like a dick for not mentioning this before:
> 
> Consider using Docker multi-stage to build the `privoxy` binary in a separate "build" image. That way you don't include the whole build chain in the final image. That's good both for size and security.
> 
> Ref: https://docs.docker.com/develop/develop-images/multistage-build/
> 
> Edit: you are actually already doing this with `FloodUI` build, just do something similar for `privoxy`

That's a great idea, and will do :) It's something I've not done before (The FloodUI stuff is by the other docker-transmission-openvpn maintainer's), so will have a look into later 👍

privoxy/scripts/start.sh
jay-to-the-dee: I was trying to stick to the existing style and must have missed that, thanks!

jay-to-the-dee:
> > Sorry I feel like a dick for not mentioning this before:
> > Consider using Docker multi-stage to build the `privoxy` binary in a separate "build" image. That way you don't include the whole build chain in the final image. That's good both for size and security.
> > Ref: https://docs.docker.com/develop/develop-images/multistage-build/
> > Edit: you are actually already doing this with `FloodUI` build, just do something similar for `privoxy`
> 
> That's a great idea, and will do :) It's something I've not done before (The FloodUI stuff is by the other docker-transmission-openvpn maintainer's), so will have a look into later 👍

@amosshapira check out https://github.com/haugene/docker-transmission-openvpn/pull/1760/commits/569c8b18eca5103df9bea4b5a42d37a650ea6a58 and see if you think I've done it right https://github.com/haugene/docker-transmission-openvpn/pull/1760/commits/1c37fbd1fc01fda2f3eb9f1379f9f97fba1e3071 :)

pkishino:
still a bunch of @haugene commits in your commit log that you should remove ;)

openvpn/tunnelDown.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 /etc/transmission/stop.sh
-[[ ! -f /opt/tinyproxy/stop.sh ]] || /opt/tinyproxy/stop.sh
+[[ ! -f /opt/privoxy/stop.sh ]] || /opt/privoxy/stop.sh
```
amosshapira, line 4: Same comment here about reversing the condition to make it clearer:
```suggestion
[[ -f /opt/privoxy/stop.sh ]] && /opt/privoxy/stop.sh
```

privoxy/scripts/start.sh
amosshapira: never mind. Another time if there is a need.

amosshapira:
> @amosshapira check out [569c8b1](https://github.com/haugene/docker-transmission-openvpn/commit/569c8b18eca5103df9bea4b5a42d37a650ea6a58) and [1c37fbd](https://github.com/haugene/docker-transmission-openvpn/commit/1c37fbd1fc01fda2f3eb9f1379f9f97fba1e3071) and see if you think I've done it right :)

The specific changes I asked for look good. Still have to update the base branch before merge is possible.

Dockerfile
jay-to-the-dee: Will mark this as resolved then @pkishino :)

jay-to-the-dee:
So the current PR design intent is to replace tinyproxy with privoxy as a drop in replacement, rather than trying to use tinyproxy as an upstream proxy. Are we okay to try it this way and see if anyone misses the (not very secure anyway) HTTP Proxy auth feature?

pkishino (approved):
@haugene I think this looks good, if you don't have any further comments, let's add this in

docs/index.md
```diff
@@ -54,5 +54,5 @@ could be wrong. So starting with number 1 is never a bad idea.
 
 ## Additional features
 
-* [Web proxy: Tinyproxy](web-proxy.md)
+* [Web proxy: Privoxy](web-proxy.md)
```
pkishino, line 57: Can you please also update the web-proxy.md with any necessary documentation changes? This can be done separately though

amosshapira:
I don't have permission to add a preview but the code looks ok now.  Good job!

docs/index.md
jay-to-the-dee: Have done with https://github.com/jay-to-the-dee/docker-transmission-openvpn/commit/f07e865be5d9e0d47fc40f6b3ba65ddb45b9c6b3 :)

haugene:
I think this overall looks good and I'm looking forward to testing out Privoxy. This will be a good addition to the 4.0 release and good timing that we got it in on an already planned major version bump.

But something happened to the Dockerfile here that shouldn't have. We have changed the build setup for flood-for-transmission web ui and this pulled back the old and slow variant.

privoxy/scripts/start.sh
```diff
@@ -0,0 +1,55 @@
...
+{
+    if [[ -f /usr/local/etc/privoxy/config ]]; then
+      PROXY_CONF='/usr/local/etc/privoxy/config'
+    elif [[ -f /usr/local/etc/privoxy/privoxy/config ]]; then
+      PROXY_CONF='/usr/local/etc/privoxy/privoxy/config'
+    else
+     echo "ERROR: Could not find privoxy config file. Exiting..."
+     exit 1
+    fi
+}
```
haugene, line 16: The reason why we had to check two places was that tinyproxy had different default location for config based on distro. And we used to have both ubuntu and alpine support so we had to figure out where it was. Should not be an issue anymore.

jay-to-the-dee: So alpine does have a privoxy build, but I chose to build manually because a) more up to date version than the repo b) Being able to compile in and try out the HTTPS MITM support which will greatly enhance Privoxy's effectiveness and ability to change requests on the now mostly HTTPS web (it's still experimental though, so have left it off for now, but a user could enable it simply by changing the config without recompiling)

Issue #1737, opened 2021-02-23
Reverse proxy of other services (Sonarr, Radarr, etc.) connected to the docker-transmission-openvpn network


garret:
I have connected other containers (Sonarr, Radarr, etc...) to the `docker-transmission-openvpn` container by using the `--network container:transmission-openvpn` option. Everything works, thus all the containers are also under the vpn network. 

However, I would like to access also their web interface outside my local network. I have been using the `haugene/transmission-openvpn-proxy` image to get the transmission web interface to be accessible from outside. That works, but what if I also want to add the other services (Sonarr, Radarr, etc...)? 

May I ask for some help in how to modify the `nginx.conf` present in https://github.com/haugene/docker-transmission-openvpn/tree/master/proxy so that I could build my own image of the proxy?

pkishino:
To be honest, for accessing any of my internal interfaces I went away from opening it up and instead closed down everything and setup WireGuard server in my local and only my setup devices can now access everything from everywhere as if they were local, it’s just more secure and simple

garret:
@pkishino I agree, I had also actually set up a wireguard instance which was working as you mention. However, I recently moved to a new place where port forwarding is not possible and, thus, don't have the possibility to set up a wireguard server. To circumvent the problem I am using [ZeroTier](https://www.zerotier.com/) (for those who don't know what it is, it is just a solution like Hamachi for those _old_ people who remember it).

pkishino:
had a quick look and should be simple if you google nginx config etc.. you would need to create new locations for the different services you want to use, that should be all and then rebuild locally. Will close for now..
if you get it working feel free to submit a PR that would update and allow other services as well

garret:
@pkishino as I wrote before, I have always found this nginx stuff quite difficult to understand. I ended up solving by using ZeroTier as mentioned above. Hope someone will update with a more elegant solution anytime in the future.

garret:
Sorry, I reopen this topic as only now I am finally having time to play around. I have `transmission-openvpn` and `sonarr `(from linuxserver on standard port `8989`) that share the same network (so sonarr is also behind vpn too). I also started the `transmission-proxy` container to be able to access such container from "outside". 

This is an extract of my docker-compose file (removed some not-needed lines to make it simpler): 

```
version: '3'
services:
    transmission:
        container_name: transmission
        cap_add:
            - NET_ADMIN
[...]
# Ports disabled as I will have them enabled in the proxy to access from outside
[...]

PR #1766, opened 2021-03-10
Added simple example to nginx for further services

pkishino:
This example should help people doing what #1737 suggests


haugene (approved):
Good idea :+1:

PR #1769, opened 2021-03-14
Remove .ovpn configs, moving to separate repo

haugene:
Removing internal config files for providers that rely on checked in .ovpn files, certs and supporting scripts.
This PR prepares the repo for #1744


haugene:
Let's do it...

Issue #1637, opened 2020-12-26
fastestvpn many of udp configs are not working


rsquarev:
**Which provider?**
FastestVpn

few of the configs that didnt work. There may be many, but I tried these
India-Jaipur-UDP.ovpn
US-Denver-UDP.ovpn
US-Chicago1-UDP.ovpn
US-Chicago3-UDP.ovpn

Could someone please update these files?

post4unme:
I have the same problem. Please update the fastestvpn files.

Gansgar:
The list of VPN servers by FastestVPN are listed here:
https://support.fastestvpn.com/vpn-servers

pkishino:
Create a PR for this and we can merge it in

bijfy:
Fastestvpn.com doesnot work now.  Says cannot resolve hostname from Fastestvpn servers

bijfy:
What is pr .. I can try help something

pkishino:
please check #1773 for the new repo url, then create an issue there.. we are not accepting any more configs on this repo for now

haugene:
Hey @pkishino. I thought I'd use this as an opportunity to test the new config repo. I found a better URL for configs from a previous commit and created this issue https://github.com/haugene/vpn-configs-contrib/issues/1 (issue number one, should be a good example to follow :smile:)

I have now updated the configs and they should be available on the dev tag of the image :crossed_fingers: 

@rsquarev @post4unme @Gansgar @bijfy : Are you up for testing it? You have to switch the version of the image from `haugene/transmission-openvpn` to `haugene/transmission-openvpn:dev` and possibly do a `docker pull haugene/transmission-openvpn:dev` to make sure you are on the latest version of dev.

Please post your results here. If it works, that's great and if not we will have a look at why.

bijfy:
Yes I am ready to test it out. Thanks

bijfy:
did not get it working. here is the output

./etc/openvpn/fetch-external-configs.sh: line 16: config_repo_temp_dir: unbound variable


Starting container with revision: 8099b6c575f769bed839b275fb8dab2f72ff12ab


WARNING: initial DNS resolution test failed


Creating TUN device /dev/net/tun


Using OpenVPN provider: FASTESTVPN


Running with VPN_CONFIG_SOURCE auto


No bundled config script found for FASTESTVPN. Defaulting to external config


[...]

PR #1773, opened 2021-03-15
Fastestvpn new servers update request

bijfy:
Hi. Can you please update fastestvpn.com servers for haugene transmission-openvpn.  Thanks


haugene:
You are completely ignoring the templates that is clearly stated that you should use.

We are also in the process of moving all the configs out of this repository and into this one: https://github.com/haugene/vpn-configs-contrib
It will then be loaded from there into the container when it starts. So please go to that repository and create an issue there instead. But follow the template or it will get closed. We're not asking for much, but a link to where we can find the configs is a minimum.

haugene:
I see now that this is a pull-request(?). Didn't catch that at first. Create an issue next time :sweat_smile:

haugene:
Aha... #1637 It's coming together now...

PR #1778, opened 2021-03-18
Created workflow for auto-close issues

### LINKED ISSUE #1754: Add GitHub action to automatically close issues that don't follow issue templates
### Is your feature request related to a problem? Please describe.
The problem is that we have a lot of issues that are created without providing the necessary details.

### Describe the solution you'd like
A bot that could automatically close issues that did not conform. Reduce the manual overhead of asking for more information.

### Describe alternatives you've considered
Better documentation. But we probably need a combination here.

### Additional context
Tried to add this action https://github.com/lucasbento/auto-close-issues and tested it in #1753. It didn't work. It output some error about invalid input options.

There are multiple issues filed with the repo, but it seems to have gone stale. I tested other forks and I think this is the closest one and the best starting point for a possible fix:  https://github.com/Stanzilla/auto-close-issues. It also seems to have died out.

So I guess we have to consider fixing it ourselves? Adding this issue as a reminder for later, and maybe someone with a bit of JavaScript experience could give it a go? It would help us maintain this project more efficiently.

I think it should be tested on another temporary repo though. I pushed a handful of tests to master and it would make a mess of the git history before we got it running I'm afraid :laughing:

pkishino:
I extended the previously mentioned action and added some more support for checkboxes and code boxes..tried it successfully on one of my repos 
https://github.com/pkishino/dive-dimmer
fixes #1754

### DIFF
--- a/.github/workflows/issue_invalid_action.yml
+++ b/.github/workflows/issue_invalid_action.yml
@@ -1,0 +1,16 @@
+on:
+  issues:
+    types: [opened, edited]
+
+jobs:
+  auto_close_issues:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: "Automatically close issues that don't follow the issue template"
+        uses: pkishino/auto-close-issues@v1.0.7
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


pkishino:
This will not affect any of the current issues etc and it only runs when issues are opened or edited

haugene:
Awesome stuff @pkishino ! :clap: 
When I saw the action we tried the last time and that it didn't work I really wanted to fork it and try to fix it.
I have not been close to finding the time - but I'm glad you did. Things are feeling cleaner already :smile:

PR #1785, opened 2021-03-22
Don't fail `rm` if VPN_PROVIDER_HOME doesn't exist.

jake-melikov:
There are only very few providers's directories in https://github.com/haugene/docker-transmission-openvpn/tree/dev/openvpn and VPN_PROVIDER_HOME might very likely be missing for many providers without bundled setup scripts.

### DIFF
--- a/openvpn/fetch-external-configs.sh
+++ b/openvpn/fetch-external-configs.sh
@@ -40,7 +40,7 @@ if [[ "${VPN_CONFIG_SOURCE_TYPE}" == "github_zip" ]]; then
 
   # Replace current provider home folder with the downloaded directory
   echo "Found configs for ${VPN_PROVIDER^^} in ${provider_configs}, will replace current content in ${VPN_PROVIDER_HOME}"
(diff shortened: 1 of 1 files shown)


pkishino:
you are aware all configs have moved to here, right?
https://github.com/haugene/vpn-configs-contrib

jake-melikov:
@pkishino Yes, I'm aware of that.

As a matter of fact, I was testing the container with configs being fetched from https://github.com/haugene/vpn-configs-contrib . For a provider which doesn't have "bundled setup scripts" (and thus `openvpn/VPN_PROVIDER` directory in THIS repository), the `rm` command in question fails.

pkishino:
Ah, good find.
@haugene should take a look, it’s still a bit trial and error

pkishino (approved):

haugene:
I was testing mostly before deleting all the providers, so didn't catch that one. Thanks @jake-melikov :+1:

Issue #1726, opened 2021-02-18
DNS resolution fails on arm7 based machines due to upstream change to time64


dabiged:
**Before creating this issue I have:**
- [x] Searched for [similar issues](https://github.com/haugene/docker-transmission-openvpn/issues)
- [x] [Read the documentation](https://haugene.github.io/docker-transmission-openvpn/). Especially the troubleshooting section and FAQ
- [x] Tried to add as much relevant information to the issue as possible

**Describe the problem**
The datetime system on the alpine base image is broken on my raspberry pi4, and probably all arm7 based systems.
I believe this stems from the recent alpine change to the time64 system explained here: https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.13.0#time64_requirements. Although this is an upstream issue I believe it impacts this project and there may be other affected users.
This manifests on my system as a DNS error, that has nothing to do with networking. 

I think this is also related to the recent activity on #1564 as they have very similar issues.

This is my error log from my base image.
```
Starting container with revision: 9d975aba54f79e4fecc57969e51c780a6f9a97af
One or more OVERRIDE_DNS addresses found. Will use them to overwrite /etc/resolv.conf
WARNING: initial DNS resolution test failed
Creating TUN device /dev/net/tun
mknod: /dev/net/tun: File exists
Using OpenVPN provider: PIA
Provider PIA has a custom setup script, executing it
Downloading OpenVPN config bundle openvpn into temporary file /tmp/tmp.CKNIaf
curl: (6) Could not resolve host: www.privateinternetaccess.com
```

**Logs**
During the trouble shooting we are asked to check if networking is working for the basic alpine linux image. 
It is not working for my raspberry pi arm7 based machine and I believe it isn't because the date-time system is completely borked.

```
pi@raspberrypi4:~/docker $ docker run --rm -it alpine sh
/ # date
Sun Jan  0 00:100:4174038  1900
/ # nslookup google.com
nslookup: clock_gettime(MONOTONIC) failed
```
similarly the base config for this image is not working either

```
pi@raspberrypi4:~/docker $ docker run --rm -it   -e OPENVPN_PROVIDER=PIA   -e OPENVPN_CONFIG=france   -e OPENVPN_USERNAME=donald   -e OPENVPN_PASSWORD=duck   haugene/transmission-openvpn:latest /bin/bash
bash-5.1# date
Sun Jan  0 00:100:4174038  1900
bash-5.1# nslookup google.com
[...]

Issue #1536, opened 2020-11-20
High CPU usage , low bandwidth


chmol:
**Describe the problem**

Hi, since a few day I've been experiencing really high cpu load (1 core at 100% for transmission) when downloading : the speed reach the maximum my connection can handle (~200mbps) then drop to 50mbps whereas the load stays the same.

Openvpn usage is low.

**systemd**
```

[Unit]
Description=haugene/transmission-openvpn docker container
After=docker.service
Requires=docker.service

[Service]
User=bittorrent
TimeoutStartSec=0
ExecStartPre=-/usr/bin/docker kill transmission-openvpn
ExecStartPre=-/usr/bin/docker rm transmission-openvpn
ExecStartPre=/usr/bin/docker pull haugene/transmission-openvpn
ExecStart=/usr/bin/docker run --sysctl  net.ipv6.conf.all.disable_ipv6=0 \
       --name transmission-openvpn \
       --cap-add=NET_ADMIN -d \
       -e CREATE_TUN_DEVICE=true \
       -v /home/bittorrent/data:/data \
       -v /mnt/pool/media:/mnt/pool/media \
       -v /srv/pool/media:/srv/pool/media \
       -v /mnt/pool/downloads/series:/mnt/pool/downloads/series \
       -v /srv/pool/downloads/series:/srv/pool/downloads/series \
       -e "OPENVPN_PROVIDER=MULLVAD" \
       -e "OPENVPN_CONFIG=nl_all" \
       -e "OPENVPN_USERNAME=user" \
       -e "OPENVPN_PASSWORD=username" \
       -e "LOCAL_NETWORK=192.168.2.0/24" \
       -e "TRANSMISSION_CACHE_SIZE_MB=1400" \
       -e "TRANSMISSION_DOWNLOAD_DIR=/mnt/pool/downloads/" \
       -e "TRANSMISSION_DHT_ENABLED=false" \
       -e "TRANSMISSION_ENCRYPTION=2" \
       -e "TRANSMISSION_INCOMPLETE_DIR_ENABLED=false" \
       -e "TRANSMISSION_PEER_PORT=19279" \
       -e "TRANSMISSION_PEX_ENABLED=false" \
       -e "TRANSMISSION_RPC_AUTHENTICATION_REQUIRED=true" \
       -e "TRANSMISSION_RPC_USERNAME=blabla" \
       -e "TRANSMISSION_RPC_PASSWORD=dldldldllddl"\
[...]

PR #1792, opened 2021-03-29
Changing back to Ubuntu base image

haugene:
## Breaking change
```txt
No. There should not be any breaking changes. People running custom scripts might get errors if they've based
the scripts on busybox versions of certain tools - but it should hopefully be a quick fix for those affected.
```

## Proposed change
```txt
I propose that we move back to basing the image on Ubuntu.
```
This relates to issues with the latest Transmission version: #1536 
And it relates issues on arm architecture with the latest alpine: #1726

But also other minor issues people have had with busybox tooling ++.
Long story short. Ubuntu is known and safe for many.

## Type of change
It is a bit of everything I guess. And hopefully nothing :smile: 

## Additional information:

Some history. We started on ubuntu, and only for amd64 architecture. Then the community contributed to
adding an alpine image and it was tagged `alpine`, then we got `armhf` architecture support and we got
`latest-armhf` and `alpine-armhf` or what it was. And then arm64 showed up, and we got a lot of builds to maintain.
At this time we were more or lett maintaining a separate Dockerfile for each image tag.

Things have gotten easier. We have got the multiarch builds now, so we build armhf, arm64 and amd64 at the same time
and expose one image to pull. In return we've got better version tagging. That was more work previously (with so
many images) that it naturally happened more rarely.

Maintaining this image takes time. And it would make all our lives easier if it just worked. Some will want it to be very small, very fast, latest version of all libraries and base images +++. In this PR I'm suggesting to set a course for the future development. That it should be based on stable versions and that we'll take our time and grow with the LTS'es out there.

I think the majority of the users prefers this, and we can think about running an `edge` version of the image for those that want to contribute, test and be ahead of the curve.

## Checklist


haugene:
FYI. I know this PR is not complete. I wanted to get a version up and running quickly on my localhost.
Have to add back privoxy that was introduced by @jay-to-the-dee in #1760. I'll rebase on top of that when I've got it working.

jay-to-the-dee:
@haugene I'm sure we can get the build process running for Privoxy running fine on Ubuntu, may just need to find the package name equivalents is all. :)

Personally opinion is I really don't mind basing the image on Ubuntu. Tbh, torrenting needs a decent amount of disk space, Internet bandwidth and CPU power anyway, so no real imperative reason to insist on the docker image being Alpine or lightweight when the intended application itself usually isn't ;)

haugene:
Glad to hear it, and thanks for the quick reply :+1: I don't think the difference for either users or maintainers will be big here. The point is rather that where it is different it leans towards stability and known territory.

About that... I was tempted and just added `privoxy` to the list of packages to install here. That would intsall version 3.0.28, and I see that your build from source gives us 3.0.29. But there was another benefit to building it that we could run  some new feature?

pkishino (approved):
Looks good to me and I agree that it will be a good move back to stability for all platforms

haugene:
With my last commit Privoxy is working again within the container. It is just the version that is included in Ubuntu package repos but whether we should run a newer version might be another discussion.

I also cleaned up the startup script and changed killall for pkill as pkill is not in ubuntu by default.

What's the verdict here? Merge now and consider adding the custom build for newer version of privoxy later?
Would be nice to provide a fix for some of the alpine instabilities and have people test this out. I'm feeling that v4 of the image should go to master soon. But want to run it in dev for a little while at least to weed out the worst bugs.

pkishino:
Yeah, I’m happy for the dev merge to go ahead.
Currently no fixed network at home due to move etc so won’t be able to test
until another week at least.

jay-to-the-dee:
The HTTPS MITM support is still experimental/ alpha so much so, I don't use it at at all at present. It will however be a really great feature when it is ready and a bit more user friendly to use potentially (currently you have to make your own certificates)! 

So for now I would just use the Ubuntu package if it's that up to date (I didn't know it was that recent) ‐ especially if like you say, you're going for stability. The build script could be useful in the future though when those newer features come out that may or may not be enabled in the Ubuntu builds.

haugene:
Definitely! Let us know when it's released. Anything that auto-generates certificates makes life so much easier. So let's see when they release it and how far it is from getting into the Ubuntu repos at that time.

jay-to-the-dee:
@haugene Yeah will have to develop a script based on `easy-rsa` that does that and the appropriate config if Privoxy doesn't end up including a built in mechanism to do so.

PR #1805, opened 2021-04-13
Implement selfheal.sh

### LINKED ISSUE #1747: Watchdog that would restart container or openVPN if connection lost for certain period of time.
Every day I discover that container lost it's internet access.
`curl ifconfig.me` displays error, as ping to any address as well.
It's probably OpenVPN issue itself. I think so, because I had that problem also when was running OpenVPN on the router (not by docker). Every day gateway was discovered as not responding. 

I would suggest add watchdog option that in case of error would restart whole container. 

Alternatively it could be solution that I used on router - simple watchdog based on cron that in case of lost connection  was killing and relaunching OpenVPN client.

mustanggb:
## Breaking change

```txt
None
```

## Proposed change

```txt
An integrated version of autoheal to avoid reliance on a separate container.
```

## Type of change

## Additional information

- This PR fixes or closes issue: ```fixes #1747```

## Checklist

- [x] Added environmental variable `SELFHEAL=false`
- [x] Added environmental variable `SELFHEAL_INTERVAL=1m`
- [x] Added environmental variable `SELFHEAL_TIMEOUT=30s`
- [x] Added environmental variable `SELFHEAL_START_PERIOD=0s`
- [x] Added environmental variable `SELFHEAL_RETRIES=3`


pkishino (approved):
Looks nice, @haugene  what do you think?

PR #1810, opened 2021-04-20
Remove git

mustanggb:
## Breaking change

```txt
None
```

## Proposed change

```txt
It was decided in #1744 to curl configs rather than git clone them.
Around 12% of the image is taken up by git which is unused.
Remove git for a smaller image.
```

## Type of change

## Additional information

- None

## Checklist

- None


mustanggb:
Nevermind, looks like it is being used after all.

PR #1812, opened 2021-04-20
Make selfheal less verbose

mustanggb:
## Breaking change

```txt
None
```

## Proposed change

```txt
Tidy up selfheal with a few aims:
- Be less verbose, only displaying useful log messages, basically if the status changes. (Feature)
- More closely mimic how the regular HEALTHCHECK behaves, namely during the starting phase. (Bugfix)
```

## Type of change

## Additional information

- None

## Checklist

- None

### DIFF
--- a/scripts/selfheal.sh
+++ b/scripts/selfheal.sh
@@ -16,38 +16,43 @@ string_to_seconds() {
   esac
 }
 
-INTERVAL=$(string_to_seconds ${SELFHEAL_INTERVAL})
-TIMEOUT=$(string_to_seconds ${SELFHEAL_TIMEOUT})
+INTERVAL=$(string_to_seconds 1m)
+TIMEOUT=$(string_to_seconds 1m)
 START_PERIOD=$(string_to_seconds ${SELFHEAL_START_PERIOD})
-
-RETRIES_REMAINING=${SELFHEAL_RETRIES}
+RETRIES_REMAINING=2
 CONTAINER_STATUS="starting"
 START_PERIOD_END=$(echo "$(date +%s) + ${START_PERIOD}" | bc -l)
+
+echo "SELFHEAL: Container is starting, waiting for it to become healthy..."
 while true; do
-  echo "SELFHEAL: Running health check..."
+  sleep ${INTERVAL}
+  MESSAGE=""
   timeout ${TIMEOUT} setsid /etc/scripts/healthcheck.sh > /dev/null
   STATUS=$?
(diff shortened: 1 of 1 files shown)


scripts/selfheal.sh
```diff
@@ -16,38 +16,43 @@ string_to_seconds() {
   esac
 }
 
-INTERVAL=$(string_to_seconds ${SELFHEAL_INTERVAL})
-TIMEOUT=$(string_to_seconds ${SELFHEAL_TIMEOUT})
+INTERVAL=$(string_to_seconds 1m)
```
pkishino, line 19: why remove the configureable time? why hardcode it ?

scripts/selfheal.sh
```diff
@@ -16,38 +16,43 @@ string_to_seconds() {
...
+TIMEOUT=$(string_to_seconds 1m)
 START_PERIOD=$(string_to_seconds ${SELFHEAL_START_PERIOD})
-
-RETRIES_REMAINING=${SELFHEAL_RETRIES}
+RETRIES_REMAINING=2
```
pkishino, line 22: what is the benefit of hardcoding this instead of having it configurable with defaults?

scripts/selfheal.sh
```diff
@@ -16,38 +16,43 @@ string_to_seconds() {
...
+    if [[ $(date +%s) -ge ${START_PERIOD_END} ]]; then
+      MESSAGE="failure"
       let "RETRIES_REMAINING-=1"
-      echo "SELFHEAL: Health check failed, ${RETRIES_REMAINING} retries remaining"
-    else
-      echo "SELFHEAL: Health check failed"
     fi
   else
+    if [[ ${CONTAINER_STATUS} == "starting" ]]; then
+      INTERVAL=$(string_to_seconds ${SELFHEAL_INTERVAL})
```
pkishino, line 39: why not just re-use from above?

pkishino (commented):
Hey, have a few comments, could you please elaborate ?

mustanggb:
So basically when the container is in "starting" mode it uses hard coded defaults, then once it becomes "healthy" it switches to using the user defined options.

As mentioned this mimics how the real HEALTHCHECK works.

Originally I figured that HEALTHCHECK didn't start checking until it became healthy or the start period ended, but this isn't the case. It does it in a pre-determined way to give the container time to spin up.

pkishino:
ok, thanks for the explanation :) it looks good to me then

PR #1813, opened 2021-04-20
Added dynamic configuration script for surfshark.

jake-melikov:
## Breaking change

```txt
```

## Proposed change

```txt
Add a dynamic configuration script for the Surfshark provider.

Am I doing it rite? As far as I understand this is the new way of doing things. I wanted to create a PR for one of the providers to see if this is the way to move forward.
```

## Type of change

## Additional information

## Checklist

### DIFF
--- a/openvpn/surfshark/configure-openvpn.sh
+++ b/openvpn/surfshark/configure-openvpn.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+set -e
+
+if [ -z "$VPN_PROVIDER_HOME" ]; then
+    echo "ERROR: Need to have VPN_PROVIDER_HOME set to call this script" && exit 1
+fi
+
+# Download and extract wanted bundle into temporary file
+tmp_file=$(mktemp)
+echo "Downloading OpenVPN config bundle into temporary file $tmp_file"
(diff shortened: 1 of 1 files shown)


pkishino:
looks good to me :)
@haugene shall we start merging these in here or have all these provider scripts at the other repo?

jake-melikov:
This is how I understood @haugene's 2nd point in https://github.com/haugene/docker-transmission-openvpn/pull/1744#issuecomment-787106985  - the automatic download and extraction scripts should live in this repo and the `.ovpn` files should live in the external repo. 

There's still the question of whether the automatic extraction scripts should also live in the [official external repo](https://github.com/haugene/vpn-configs-contrib). Adding both a script and `.ovpn` files in the [official external repo](https://github.com/haugene/vpn-configs-contrib) doesn't really makes sense - IIUC the code, in such cases the script from the external repo will be executed and `.ovpn` files won't ever be used.

On the other hand, someone could point to a different external repo and have a custom script there with a custom logic baked in.

pkishino:
yes, I hoenstly think keeping the provider scripts on the other repo is a better approach as I believe these will change and require more regular updates..

Issue #1667, opened 2021-01-19
PIA dns resolv.conf misconfiguration


izenn:
**Before creating this issue I have:**

- [X] Searched for [similar issues](https://github.com/haugene/docker-transmission-openvpn/issues)
- [X] [Read the documentation](https://haugene.github.io/docker-transmission-openvpn/). Especially the troubleshooting section
- [X] Tried to add as much relevant information to the issue as possible

**Describe the problem**

Resolv.conf changed to PIA-internal dns (10.0.0.243) before the VPN tunnel is established causing the curl to fail when starting the container

**Add your docker run command or docker-compose file**

Attached

**Logs**

Starting container with revision: 7b453b9d31ca11be10eca5e5abe744ef1bb0708b
Creating TUN device /dev/net/tun
mknod: /dev/net/tun: File exists
Using OpenVPN provider: PIA
Provider PIA has a custom setup script, executing it
Downloading OpenVPN config bundle openvpn-strong into temporary file /tmp/tmp.PIJEml
curl: (6) Could not resolve host: www.privateinternetaccess.com

**Host system:**

Synology nas Docker version 18.09.8, build bfed4f5

izenn:
<img width="644" alt="Origin - Synology DiskStation1" src="https://user-images.githubusercontent.com/744288/104989490-331cbd80-59e0-11eb-9e77-f0e44f3ff2f9.png">
<img width="645" alt="Origin - Synology DiskStation2" src="https://user-images.githubusercontent.com/744288/104989501-37e17180-59e0-11eb-85bd-db6ad7d01cf7.png">
<img width="640" alt="Origin - Synology DiskStation3" src="https://user-images.githubusercontent.com/744288/104989509-3b74f880-59e0-11eb-8d4f-504872e9fcd7.png">

izenn:
additional investigation shows that it is not the initial run/dns settings that causes the dns issue, but after the first stop/start the dns is not reverted from the 10.0.0.243 to a publicly available dns server

derekn:
i'm experiencing this issue also. i have 2 different boxes running the same image and config, one works and the other cannot resolve www.privateinternetaccess.com... weird. did a down/up and still not working. switching to the dev image fixed it, but hope it gets resolved in latest

huntastikus:
Same here!

pkishino:
I reverted the dns fix, please let me know if it works (I was not having any issues with this..)

izenn:
[...]

PR #1819, opened 2021-04-26
fix: force PEER_DNS servers through the vpn tunnel

ferringb:
## Breaking change

```
```

## Proposed change

```txt
This PR fixes cidr conflicts between LOCAL_NETWORK and
any NS advertised by the provider; see #1667 for an example.

Specifically, PIA advertised 10.0.0.243, users have `LOCAL_NETWORK=10.0.0.0/8`.

This results in 10.0.0.243 traffic being routed around the tunnel, rather than through it.
```

## Type of change

## Additional information

- This PR fixes or closes issue: ```fixes #1713, #1667 ```

## Checklist


docs/config-options.md
```diff
@@ -14,6 +14,8 @@
 | `OPENVPN_OPTS`      | Will be passed to OpenVPN on startup                                                                | See [OpenVPN doc](https://openvpn.net/index.php/open-source/documentation/manuals/65-openvpn-20x-manpage.html) |
 | `LOCAL_NETWORK`     | Sets the local network that should have access. Accepts comma separated list.                       | `LOCAL_NETWORK=192.168.0.0/24`                                                                                 |
 | `CREATE_TUN_DEVICE` | Creates /dev/net/tun device inside the container, mitigates the need mount the device from the host | `CREATE_TUN_DEVICE=true`                                                                                       |
+| `PEER_DNS`          | Controls whether to use the DNS provided by the OpenVPN endpoint. | To use your host DNS rather than what is provided by OpenVPN, set `PEER_DNS=no`.  This allows for potential DNS leakage. |
+| `PEER_DNS_PIN_ROUTES` | Controls whether to force traffic to peer DNS through the OpenVPN tunnel. | To disable this default, set `PEER_DNS_PIN_ROUTES=no`. |
```
pkishino, line 18: Are you going to add these to default settings file as well?

pkishino:
Looks good, just think these should be added to defaults

ferringb:
> Looks good, just think these should be added to defaults
So uh... which defaults we talking about here? :)

If you're referring to the Dockerfile env defaults- I'll push a commit shortly addressing that.  If you're looking for something else, please clarify (I doubt you're referring to `default-settings.json` since that's transmission config).

pkishino:
yeah sorry, I meant the dockerfile.. battling bit of a cold currently, not in the right mind :P

pkishino (approved):

mustanggb:
Why is it using `yes/no` when other env flags are using `true/false`.

I think it would make more sense to users to stick with one standard.

pkishino:
Wow, 
That’s bad of me to miss that.. I’ll see if I can fix that later tonight. You are right, we should stick to true and false

ferringb:
See #1820 ; I cut a quick PR for changing the configurables in the way y'all are discussing above.

PR #1820, opened 2021-05-02
fix: use true/false for boolean configurables for PEER_DNS*

ferringb:
`PEER_DNS` previously was using 'yes/no', but the standard is 'true/false'.  This realigns
#1819 to that standard.

For `PEER_DNS_PIN_ROUTES`, no one should be effected by the configurable change- it's been in
the dev branch only for a few days.

`PEER_DNS` however, has been available for users for a few months now, and users have
been told to use `PEER_DNS=no` in places like #1713 or #1667 .  Realigning the boolean definition
here is 'a breaking change' technically, although it's my strong suspicion that #1819  fixed
the underlying problem that warranted users using `PEER_DNS=no` in the first place.

I defer to the codebase owner's opinion for how they want this to be handedl.

## Breaking change

```txt
`PEER_DNS=no` is now `PEER_DNS=false`.  It's suggested end users revalidate if they actually need this setting in light of route pinning fixes in #1819 
```

## Proposed change

```txt
Rewrite PEER_DNS* to use 'true/false' per https://github.com/haugene/docker-transmission-openvpn/pull/1819#issuecomment-830840281
```

## Type of change

## Additional information

- This PR is related to issue: ```relates to #1819 ```

## Checklist


ilike2burnthing:
Does the Dockerfile also need changed?
https://github.com/haugene/docker-transmission-openvpn/blob/77d307df1f9b15a710846dbc7101ed3cc00cfce0/Dockerfile#L59-L60

openvpn/tunnelUp.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-if [ "${PEER_DNS}" != "no" ]; then
+if [ "${PEER_DNS}" != "false" ]; then
```
pkishino, line 2: I would just check if true.. should work to just check the variable directly and interpret it as a bool

ferringb: No argument there, I've just been following the style w/in that file... I've generally not looked through the other code until now.  Related to that, this should be `[[ "${PEER_DNS,,}" == true ]]`; for case reasons.

Pushing an update shortly to address this.

PR #1874, opened 2021-07-13
Reflected watch directory support

jeffplang:
## Breaking change

```txt
```

## Proposed change

This is an opinionated take on multiple watch directory support.  This PR adds a script based on inotify that recursively watches a directory, specified by a new environment variable `REFLECTED_WATCH_DIR`, for newly created .torrent files.  It adds them to Transmission and sets their download location to the same subdirectory, but relative to `TRANSMISSION_DOWNLOAD_DIR`.  Better explained with an example:

* REFLECTED_WATCH_DIR is `/data/refl_watch`:
* User adds Ubuntu-iso.torrent to `/data/refl_watch/ISOs/`
* Transmission will download the files to `/data/completed/ISOs/`

It basically works the same way Transmission's current watch functionality does, but it "reflects" the download location based on where you dropped the .torrent file.  

I realized after the fact this could all really be done in a `transmission-post-start.sh` script, but if the functionality is useful enough, it would be great to add it to the image.

## Type of change

## Additional information

Based on a script found here: https://forum.transmissionbt.com/viewtopic.php?t=15705

* Adds dependencies of `inotify-tools` and `coreutils` (for `realpath` with --relative-to support)
* The size checking code is kinda ugly.  The better alternative would be to monitor for `close_write` events, but I could not get those to fire in the container when adding a torrent file.  Maybe I am missing something.

## Checklist

### DIFF
--- a/scripts/reflected-watch.sh
+++ b/scripts/reflected-watch.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+inotifywait -q -r -m $1 --format %w%f -e create |
+while read file; do
+  if echo "$file" | grep -iq "\.torrent" ; then
+    canonfile=$(readlink -f "$file")
+    relpath=$(realpath --relative-to="$REFLECTED_WATCH_DIR" "$canonfile")
+    dir=$(dirname "$relpath")
+
+    if echo `basename "$file"` | grep -iq "^\._"; then
+      echo
+    else
+      counter=0
+      while : ; do
+        if [ "$counter" -gt "10" ]; then break; fi
+        sleep 0.5s
+        a=`stat -c%s "$canonfile"`
+        if [ "$?" -ne "0" ]; then break; fi
+        echo "$canonfile" stat: "$a"
+        if [ "$a" -ne "0" ]; then break; fi
+        counter=`expr "$counter" + 1`
+      done
+      while : ; do
+        a=`stat -c%s "$canonfile"`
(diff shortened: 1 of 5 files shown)


scripts/reflected-watch.sh
```diff
@@ -0,0 +1,36 @@
...
+      done
+      while : ; do
+        a=`stat -c%s "$canonfile"`
+        if [ "$?" -ne "0" ]; then break; fi
+        sleep 5s
+        if [ "$a" -eq `stat -c%s "$canonfile"` ]; then break; fi
+      done
+      a=`stat -c%s "$canonfile"`
+      if [ "$?" -eq "0" ]; then
+        transmission-remote -a "$canonfile" -w "$TRANSMISSION_DOWNLOAD_DIR/$dir";
```
pkishino, line 30: This needs to add support for when the user has auth enabled , otherwise it will fail here

pkishino (changes requested):
please see comment

scripts/reflected-watch.sh
haugene: We're doing this auth handling a couple of other places. You need to check if auth env vars are set and if so use them in the command. It's kind of tempting to make a `/usr/local/bin/authed-transmission-remote` script and make it check and set the flags if needed. Then we could use that without thinking about auth from all the other places :thinking:

pkishino:
no updates recently, still interested in this feature?

PR #2080, opened 2021-11-27
Fix IP leak for first few seconds after tunnel up

XMB5:
## Proposed change

For the first few seconds after transmission and tinyproxy start, their traffic is not routed through the VPN. This is because we are calling openvpn with the option `--up /etc/openvpn/tunnelUp.sh`, and the `--up` script runs before ip routes are added. In other words, the VPN will be connected, but traffic will be routed outside of the VPN interface because openvpn has not changed the routing tables with the `ip route` command. In my testing, this seems to not cause an IP leak for transmission, because transmission takes a few seconds to load (however, depending on your setup, it could be a problem). I did find that tinyproxy can leak a few requests though, because tinyproxy starts up faster.

To fix this leak, I changed `--up` to `--route-up`, which runs after openvpn has modified the routing table. Here are the relevant docs from the [openvpn reference manual](https://openvpn.net/community-resources/reference-manual-for-openvpn-2-4/):
> --up _cmd_
    Run command _cmd_ after successful TUN/TAP device open (pre --user UID change). _cmd_ consists of a path to script (or executable program), optionally followed by arguments. The path and arguments may be single- or double-quoted and/or escaped using a backslash, and should be separated by one or more spaces. **The up command is useful for specifying route commands** which route IP traffic destined for private subnets which exist at the other end of the VPN connection into the tunnel...

> --route-up _cmd_
    Run command _cmd_ **after routes are added**, subject to --route-delay. _cmd_ consists of a path to script (or executable program), optionally followed by arguments. The path and arguments may be single- or double-quoted and/or escaped using a backslash, and should be separated by one or more spaces...

(emphasis mine)

## Type of change

## Additional information

To test the IP leak on your own, create a script `print-ip.sh`:
```bash
#!/bin/bash

echo "start $0, RUNNING_IN_BACKGROUND=${RUNNING_IN_BACKGROUND:-false}"

if [ -z "$RUNNING_IN_BACKGROUND" ]; then
  # emulate launching program in background (like transmission or tinyproxy)
  RUNNING_IN_BACKGROUND=true "$0" &
else
  # running in background now
  curl -s https://icanhazip.com/
fi
```

Now, run the docker container with a volume that mounts `print.sh` on `/etc/openvpn/tunnelUp.sh` (for example `-v /tmp/print-ip.sh:/etc/openvpn/tunnelUp.sh:ro`. When you run the container, it will print out your non-VPN IP address:
 ```
Fri Jun 10 01:23:00 2022 TUN/TAP device tun0 opened
Fri Jun 10 01:23:00 2022 TUN/TAP TX queue length set to 100
Fri Jun 10 01:23:00 2022 /sbin/ip link set dev tun0 up mtu 1500
Fri Jun 10 01:23:00 2022 /sbin/ip addr add dev tun0 local 10.12.0.98 peer 10.12.0.97
Fri Jun 10 01:23:00 2022 /etc/openvpn/tunnelUp.sh tun0 1500 1556 10.12.0.98 10.12.0.97 init
start /etc/openvpn/tunnelUp.sh, RUNNING_IN_BACKGROUND=false
start /etc/openvpn/tunnelUp.sh, RUNNING_IN_BACKGROUND=true
1.2.3.4  <-- non-VPN IP address
Fri Jun 10 01:23:02 2022 /sbin/ip route add 44.55.66.77/32 via 192.168.100.1
Fri Jun 10 01:23:02 2022 /sbin/ip route add 0.0.0.0/1 via 10.12.0.97
Fri Jun 10 01:23:02 2022 /sbin/ip route add 128.0.0.0/1 via 10.12.0.97
Fri Jun 10 01:23:02 2022 /sbin/ip route add 10.12.0.1/32 via 10.12.0.97
Fri Jun 10 01:23:02 2022 Initialization Sequence Completed
```

With the patch:
```
Fri Jun 10 01:24:00 2022 TUN/TAP device tun0 opened
Fri Jun 10 01:24:00 2022 TUN/TAP TX queue length set to 100
Fri Jun 10 01:24:00 2022 /sbin/ip link set dev tun0 up mtu 1500
Fri Jun 10 01:24:00 2022 /sbin/ip addr add dev tun0 local 10.12.0.98 peer 10.12.0.97
Fri Jun 10 01:24:02 2022 /sbin/ip route add 44.55.66.77/32 via 192.168.100.1
Fri Jun 10 01:24:03 2022 /sbin/ip route add 0.0.0.0/1 via 10.12.0.97
Fri Jun 10 01:24:03 2022 /sbin/ip route add 128.0.0.0/1 via 10.12.0.97
Fri Jun 10 01:24:03 2022 /sbin/ip route add 10.12.0.1/32 via 10.12.0.97
start /etc/openvpn/tunnelUp.sh, RUNNING_IN_BACKGROUND=false
Fri Jun 10 01:24:03 2022 Initialization Sequence Completed
start /etc/openvpn/tunnelUp.sh, RUNNING_IN_BACKGROUND=true
44.55.66.77  <-- VPN IP address
```

## Checklist

### DIFF
--- a/openvpn/start.sh
+++ b/openvpn/start.sh
@@ -175,7 +175,7 @@ echo "${TRANSMISSION_RPC_PASSWORD}" >> /config/transmission-credentials.txt
 # Persist transmission settings for use by transmission-daemon
 python3 /etc/openvpn/persistEnvironment.py /etc/transmission/environment-variables.sh
 
-TRANSMISSION_CONTROL_OPTS="--script-security 2 --up-delay --up /etc/openvpn/tunnelUp.sh --route-pre-down /etc/openvpn/tunnelDown.sh"
+TRANSMISSION_CONTROL_OPTS="--script-security 2 --route-up /etc/openvpn/tunnelUp.sh --route-pre-down /etc/openvpn/tunnelDown.sh"
 
 ## If we use UFW or the LOCAL_NETWORK we need to grab network config info
 if [[ "${ENABLE_UFW,,}" == "true" ]] || [[ -n "${LOCAL_NETWORK-}" ]]; then


pkishino:
@haugene  looks good to me, I’ll merge this to dev for now

pkishino:
Thanks btw, if you want/can look at any other hardening or improvements, always welcome :)

Issue #2081, opened 2021-11-27
Warning: Failed running command (--route-up) after latest dev release


tim0901:
### Is there a pinned issue for this?

- [X] I have read the pinned issues and could not find my issue

### Is there an existing or similar issue/discussion for this?

- [X] I have searched the existing issues
- [X] I have searched the existing discussions

### Is there any comment in the documentation for this?

- [X] I have read the documentation, especially the FAQ and Troubleshooting parts

### Is this related to a provider?

- [X] I have checked the provider repo for issues
- [X] My issue is NOT related to a provider

### Are you using the latest release?

- [X] I am using the latest release

### Have you tried using the dev branch latest?

- [X] I have tried using dev branch

### Config used
```

version: '3.3'
services:
        transmission-openvpn:
                container_name: transmission_openvpn
                image: haugene/transmission-openvpn:dev
                restart: always
                cap_add:
                    - NET_ADMIN
                volumes:
                    - ~/data/transmission-home:/config/transmission-home
                    - ~/data/media:/data
                    - networkShare:/networkShare
                environment:
                    - PUID=1001
                    - GUID=1001
                    - OPENVPN_PROVIDER=WINDSCRIBE
                    - OPENVPN_CONFIG=${OPENVPN_CONFIG}
                    - OPENVPN_USERNAME=${OPENVPN_USERNAME}
                    - OPENVPN_PASSWORD=${OPENVPN_PASSWORD}
                    - LOCAL_NETWORK=192.168.0.0/16
                    - TRANSMISSION_SPEED_LIMIT_UP=100
                    - TRANSMISSION_SPEED_LIMIT_DOWN=2048
                    - TRANSMISSION_ALT_SPEED_UP=50
                    - TRANSMISSION_ALT_SPEED_DOWN=1024
                    - TRANSMISSION_RPC_USERNAME=${RPC_USERNAME}
                    - TRANSMISSION_RPC_PASSWORD=${RPC_PASSWORD}
                    - TRANSMISSION_RPC_AUTHENTICATION_REQUIRED=true
                    - TRANSMISSION_ALT_SPEED_TIME_BEGIN=480
                    - TRANSMISSION_ALT_SPEED_TIME_END=1320
                    - TRANSMISSION_IDLE_SEEDING_LIMIT=1
                    - TRANSMISSION_IDLE_SEEDING_LIMIT_ENABLED=true
                    - TRANSMISSION_PEER_LIMIT_GLOBAL=2000
[...]

PR #2020, opened 2021-10-19
Added fallback and more explicit warning message

pkishino:
## Proposed change

```txt
This will use /data/transmission-home (previous default) as fallback if the folder exists.

```

## Type of change

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (which adds functionality to a this container)
- [ ] Breaking change (fix/feature causing existing functionality to break)

## Additional information

- This PR fixes or closes issue: ```fixes #```
- This PR is related to issue: ```relates to #```
- Link to documentation updated (if done separately): ```https://...```

## Checklist

- [x] The code change is tested and works locally.
- [x] There is no commented out code in this PR.

If user exposed functionality or configuration variables are added/changed:

- [x] Documentation added/updated

PR #2082, opened 2021-11-27
 Fix scripts to work with --route-up, and update settings.json in transmission-home folder 

XMB5:
## Proposed change

OpenVPN calls the `--up` script with arguments and environment variables, but it calls `--route-up` only with environment variables. Here, I switched the scripts to use the corresponding environment variables instead of arguments.

While I was testing this patch, I noticed another bug, related to #2020. The `updateSettings.py` script does not work for the legacy config location (`/data/transmission-home`), because we check for the legacy config folder and update `TRANSMISSION_HOME` after the python script has already run. I moved the `userSetup.sh` script before `updateSettings.py` to fix this.

## Type of change

## Additional information

- This PR fixes or closes: #2081

## Checklist

### DIFF
--- a/openvpn/tunnelUp.sh
+++ b/openvpn/tunnelUp.sh
@@ -51,7 +51,7 @@ if [[ "${PEER_DNS,,}" == "true" ]]; then
         fi
 fi
 
-/etc/transmission/start.sh "$@"
+/etc/transmission/start.sh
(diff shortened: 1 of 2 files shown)


pkishino:
Looks good, understand the problem now.

pkishino:
Closes #2081

PR #2096, opened 2021-12-04
DEBUG: print all lines before execution

edgd1er:
if var DEBUG is set to anything other than false, each bash script will be run with a set -x. This would ease debugging
var is added to persisted var as it may be useful to debug transmission and proxy startup scripts

## Breaking change

```txt
None
```

## Proposed change

```txt
when setting env variable DEBUG to any value except false, all bash commands are echoed to stdout
```

## Type of change

## Additional information

## Checklist


pkishino:
yeah, this is a great idea..just thinking if it might be "easier" and long-term more flexible to just make one script that is sourced into every other one with these kinds of "shared" vars and methods etc included?
then it becomes easier to change and update in one place instead of multiple.
Big thanks for the PR

edgd1er:
Yes, good idea, I'll to send an adjustment as suggested.

pkishino:
looks good.
only commend I have would be to use . instead of source as we use in other includes..but thats more formatting.
at next refactor I'll try to refresh these things to get them all in line (as were only using bash using source is would be more appropriate anyway)

ilike2burnthing:
If `DEBUG` isn't present or is set to `false`, I just get a single line in the log:

```text
No log line matching the '' filter
```

Container repeatedly restarts and Transmission is not accessible.

Setting to `true` resolves this.

Happy to open an issue/discussion if this is case dependent rather than an obvious oversight.

pkishino:
@edgd1er out for now, back tonight, can you check this please?

edgd1er:
@pkishino , @ilike2burnthing 

my bad, I will push a fix for this.

PR #2142, opened 2022-01-08
Add feature to NORDVPN: connect directly to a known server.

edgd1er:
## Breaking change

```txt
<placeholder>
```

## Proposed change

```txt
get configuration file based on server's fqdn, bypassing all api's recommendations. Connection may fail when the server is offline or overloaded. has precedence over NORDVPN_COUNTRY and NORDVPN_CATEGORY.
NORDVPN_SERVER=<#>.nordvpn.com: eg: nl568.nordvpn.com

useful if you want to stick to a specific server.

- possible values extracted for the API are downloaded only if needed to prevent hitting NORDVPN API requests throttling.
- added a check if returned values are html error pages.
- added a test to validate new feature.

```

## Type of change

## Additional information

## Checklist


Dockerfile
```diff
@@ -49,6 +49,8 @@ ENV OPENVPN_USERNAME=**None** \
     GLOBAL_APPLY_PERMISSIONS=true \
     TRANSMISSION_HOME=/config/transmission-home \
     TRANSMISSION_RPC_PORT=9091 \
+    TRANSMISSION_RPC_USERNAME= \
```
pkishino, line 52: Why this sudden addition here?

openvpn/nordvpn/configure-openvpn.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
pkishino, line 1: For the container I don’t see any point changing this

edgd1er: I've read somewhere (can't find the source though), that a logical she-bang (env ...) was preferred over an absolute one. this link sums up both she-bang types: https://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my
So now by default, I put a logical she-bang in my scripts, for the container, you are right there is no advantage to have this change as path to bash will not change. If this change is blocking the PR, I will push a revert commit to restore the line

Dockerfile
edgd1er: A best practice in bash is to add `set -e -u -o pipefail` in order to have a stricter usage of bash.
some exemple: https://gist.github.com/atomlab/65e70d6f9d0ed5b3aeb3fc76ff28255a
some information about strict mode: http://redsymbol.net/articles/unofficial-bash-strict-mode/

these two lines are added to avoid an unbound variable error due to `set -u` and a non existing variable.
the `set -u -e -o pipefail` will ease debug and prevent subtle bugs. 
Dockerfile will initialise in container environment, the variable to empty or default value.
the results of  `[ -z ${TRANSMISSION_RPC_USERNAME} ]  and `[ -n ${TRANSMISSION_RPC_USERNAME} ]` will remain the same 

the var is dumped in [openvpn start's script](https://github.com/edgd1er/docker-transmission-openvpn/blob/c43dd6ab2304662f4e000010e7e96303e5f5307f/openvpn/start.sh#L174), defaulting it to a known value is a good practice. If a set -u was added in that script, execution would failed. 

https://github.com/haugene/docker-transmission-openvpn/blob/c43dd6ab2304662f4e000010e7e96303e5f5307f/openvpn/start.sh#L174

this change is not mandatory for the feature to be available. I could commit a revert for this two lines. just let me know what do you prefer.

PR #2252, opened 2022-04-30
Execute a script on VPN connection failure

mustanggb:
## Breaking change

```txt
None
```

## Proposed change

```txt
When a config fails it would be nice if a script could be ran to check things are okay before it runs again.
For example check for updates to the config itself, or for updates to credentials.

There isn't enough time before container restart on failure so how this works in reality is the script will run
the next time the config is used.

This is a little bit like `openvpn-post-config.sh`, but only runs if the config has previously failed to connect.
```

## Type of change

## Additional information

- Add variable `CONFIG_MOD_FAILURE_SCRIPT`
- Allows configs to have a `config-failure.sh`

## Checklist

### DIFF
--- a/openvpn/custom/config-failure.sh
+++ b/openvpn/custom/config-failure.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+CONFIG=$1
+
+echo "Config ${CONFIG} has failed, here might be a good place to check credentials"
+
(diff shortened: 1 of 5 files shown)


pkishino:
Looks good, I’ll try to rest and get it merged this week

mustanggb:
That would be amazing.

PR #2321, opened 2022-08-18
Added docker secret support for rpc credentials

janmatias:
## Breaking change

```txt
```

## Proposed change

```txt
Issue #534 Adding secret support for rpc credentials as well.
```

## Type of change

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (which adds functionality to a this container)
- [ ] Breaking change (fix/feature causing existing functionality to break)

## Additional information

- This PR fixes or closes issue: ```closes #534 ```

## Checklist

- [x] The code change is tested and works locally.
- [x] There is no commented out code in this PR.

If user exposed functionality or configuration variables are added/changed:

- [x] Documentation added/updated

PR #2480, opened 2023-01-02
Simplify RPC creds setup

ColinHebert:
Reduce the complexity associated with creating RPC creds

## Breaking change

```txt
<placeholder>
```

## Proposed change

```txt
Simplify and fix RPC creds set up with docker secrets.
```

## Type of change

## Additional information

The current integration with docker secrets is flakey/overcomplicated.

For example in the case where there is an `rpc_creds` docker secret, which is the same as `transmission-credentials.txt` the environment variables won't be set to the correct credentials.

This pull requests, not only fixes that issue, but drastically reduces the complexity of the code.

If `rpc_creds` exists, its content will **always** be used as environment variables.
Environment variables are **always** written into the `transmission-credentials.txt` file.

## Checklist

### DIFF
--- a/openvpn/start.sh
+++ b/openvpn/start.sh
@@ -197,17 +197,11 @@ else
 fi
 
 if [[ -f /run/secrets/rpc_creds ]]; then
-  #write creds if no file or contents are not the same.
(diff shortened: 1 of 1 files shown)


ColinHebert:
cc @janmatias @pkishino who worked on the original #2321

PR #2502, opened 2023-01-20
Update config-options.md

AJConnelly:
I would add TRANSMISSION_WATCH_DIR on a new row, but given I don't use this option I'm not too familiar with what it does. Reason for adding this section is that this information is important when using other servarr docker containers. While bugfixing I read through some old issue pages that - TRANSMISSION_COMPLETED_DIRECTORY was a option whereas on current version and probably older the older versions as well was not a valid option. I would probably also add somewhere on the "Configuration options" page that these environment variables are fetched from https://github.com/haugene/docker-transmission-openvpn/blob/master/Dockerfile which are the most up to date variables. Thanks for your work on this project it is awesome! Regarding pull request, I believe this should go on master branch as its just extra documentation which should link to the wiki for this project. Sorry if a pull request is not the proper way to request a wiki update, this is the only time I've done this.

## Breaking change

```txt
<placeholder>
```

## Proposed change

```txt
Add directory environment variables to wiki
```

## Type of change

## Additional information

- Link to documentation updated (if done separately): ```https://haugene.github.io/docker-transmission-openvpn/config-options/```

## Checklist


ilike2burnthing:
TRANSMISSION_WATCH_DIR is the 'watch' or 'blackhole' directory that Transmission monitors for new .torrent files.

pkishino:
all the variables that start with TRANSMISSION are part of transmission configs, that is why we link to the official documentation for these..

Issue #2469, opened 2022-12-16
High passive memory usage (leak?) since version 4.3 (Upgrade to Ubuntu 22.04 and Transmission 3.00)


ameinild:
### Is there a pinned issue for this?

- [X] I have read the pinned issues and could not find my issue

### Is there an existing or similar issue/discussion for this?

- [X] I have searched the existing issues
- [X] I have searched the existing discussions

### Is there any comment in the documentation for this?

- [X] I have read the documentation, especially the FAQ and Troubleshooting parts

### Is this related to a provider?

- [X] I have checked the provider repo for issues
- [X] My issue is NOT related to a provider

### Are you using the latest release?

- [X] I am using the latest release

### Have you tried using the dev branch latest?

- [ ] I have tried using dev branch

### Docker run config used

```
version: '3'

services:

  transmission:
    image: haugene/transmission-openvpn:latest
    container_name: transmission-openvpn
    healthcheck:
      disable: true 
    environment:
      - TRANSMISSION_DOWNLOAD_DIR=/data/download
      - CREATE_TUN_DEVICE=true
      - OPENVPN_PROVIDER=CUSTOM
      - OPENVPN_USERNAME=****
      - OPENVPN_PASSWORD=****
      - WEBPROXY_ENABLED=false
      - TRANSMISSION_PEER_PORT=49115
      - TRANSMISSION_PORT_FORWARDING_ENABLED=false
      - LOCAL_NETWORK=10.10.0.0/22
      - DROP_DEFAULT_ROUTE=true
      - TRANSMISSION_WATCH_DIR_ENABLED=false
      - TRANSMISSION_DHT_ENABLED=true
      - TRANSMISSION_PEX_ENABLED=true
      - PUID=1000
      - PGID=1000
      - NO_LOGS=true
    ports:
      - '8333:9091'
    volumes:
      - /mnt/zfs/postern/torrents:/data
      - /mnt/docker-data/transmission/str-ams102_a309011.ovpn:/etc/openvpn/custom/default.ovpn
      - /mnt/docker-data/transmission:/config
      - /etc/localtime:/etc/localtime:ro
    cap_add:
      - NET_ADMIN
    logging:
      driver: json-file
      options:
        max-size: 10m
    restart: unless-stopped
```

### Current Behavior

The Transmission Container uses over 10 GB of memory after running for 10 days with around 25 torrents.

[...]

PR #2573, opened 2023-04-10
use alpine:latest for TransmissionUIs build stage

Entepotenz:
## Breaking change

```txt
None
```

## Proposed change

```txt
Use most recent alpine docker image to create TransmissionUIs layer
```

## Type of change

- [x] Update version

## Additional information

- use most recent alpine image instead of an alpine version which is already End of support (https://www.alpinelinux.org/releases/)

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.13 as TransmissionUIs
+FROM alpine:latest as TransmissionUIs
 
 RUN apk --no-cache add curl jq \
     && mkdir -p /opt/transmission-ui \


ilike2burnthing:
Having a fixed version number might be better for stability.

Currently `latest` is `3.17`, so either go with `3.17` (which is what LS.io are using) or play it safe and go with `3.16`?

Entepotenz:
I do not see what you want to play safe in this case?

This image layer is just using wget and curl to download data for later use.
IMHO: The likelihood forgetting to update alpine to the next version is much higher than wget or curl breaking in an official stable alpine release.

If you still prefer to use 3.17 then I can change the pull request accordingly.

ilike2burnthing:
I'm not a dev for this repo, so I'd wait for them to comment before making any changes.

Elsewhere though, the practice I've seen and used has been specifying a fixed version number, so that updates can be tested rather than being rolled out with some random commit and potentially causing issues, which this repo has occasionally had to deal with in the past.

There's even some testing around reverting to an Ubuntu base image due to some high memory usage for some - https://github.com/haugene/docker-transmission-openvpn/issues/2469.

It does mean remembering to update it, but that's easily enough done with calendar alert or just a reminder bot - https://github.com/probot/reminders.

Issue #2489, opened 2023-01-07
Web proxy stops working after period of time


justified999:
### Is there a pinned issue for this?

- [X] I have read the pinned issues and could not find my issue

### Is there an existing or similar issue/discussion for this?

- [X] I have searched the existing issues
- [X] I have searched the existing discussions

### Is there any comment in the documentation for this?

- [X] I have read the documentation, especially the FAQ and Troubleshooting parts

### Is this related to a provider?

- [X] I have checked the provider repo for issues
- [X] My issue is NOT related to a provider

### Are you using the latest release?

- [X] I am using the latest release

### Have you tried using the dev branch latest?

- [ ] I have tried using dev branch

### Docker run config used

docker run  \
--name transmission \
--privileged \
--label=com.centurylinklabs.watchtower.monitor-only=true \
--restart unless-stopped \
-v /volume1/Media/:/Media \
-e TRANSMISSION_DOWNLOAD_DIR=/Media/Downloads/Torrents/Movies \
-e TRANSMISSION_INCOMPLETE_DIR=/Media/Downloads/Torrents/Incomplete \
-e TRANSMISSION_WATCH_DIR=/Media/Downloads/Torrents/ToFetch \
-v /volume1/docker/Transmission/:/config \
-e OPENVPN_PROVIDER=EXPRESSVPN \
-e OPENVPN_CONFIG=my_expressvpn_uk_-_docklands_udp \
-e OPENVPN_USERNAME=* \
-e OPENVPN_PASSWORD=* \
-e PUID=* \
-e PGID=* \
-e LOCAL_NETWORK=192.168.0.0/24 \
-e TZ=* \
-e TRANSMISSION_WEB_UI=flood-for-transmission \
-e WEBPROXY_ENABLED=true\
              --log-driver json-file \
              --log-opt max-size=10m \
-p 27800:9091 \
-p 8888:8118 \
haugene/transmission-openvpn:4.2

### Current Behavior

The web proxy works for routing traffic for a varying period of time after creating and starting the container.  After a period of time (usually a few days) the web proxy no longer works despite the VPN still being active and downloads possible.  Restarting the container does not fix the issue.  Deleting and creating the container again with the same docker-run works for a few days.

### Expected Behavior

Web proxy should continue as long as VPN is active

### How have you tried to solve the problem?

Tried v4.3.
Tried with various configs from the same VPN provider.  
Tried fresh install with new config files.

### Log output

[...]

PR #2672, opened 2023-07-29
Fix sed regex not assigning new IP when the container reboots

mathiasflorin:
When a docker container restarts, `eth0` may receive a new IP address, causing the proxy to face issues binding to the correct listening address, as the previous 127.* IP has already been replaced by `sed` with the regex. The replacement works only once, as the configuration is changed and stored in the writable container layer. To reset the configuration and allow `sed` to replace any IP with the evaluated IP address, the container needs to be stopped and removed using `docker rm <container_id>`. This change ensures proper functioning of the proxy and addresses IP changes during container restarts as the new regex will replace any IP.

This merge request is still required even with the following `healthcheck.sh` script:
image

The `healthcheck.sh` script currently only monitors changes in IP addresses and does not update the privoxy config file accordingly. To address this, I suggest modifying the sed regular expression in `privoxy/scripts/start.sh`. This change is essential since the script should always be executed as part of the default container executable (specified by the `CMD` instruction in the Dockerfile). This script runs before the `healthcheck.sh` script.

## Breaking change

```txt
This merge request should not affect users already using the privoxy http proxy.
```

## Proposed change

```txt
The merge request allows users to continue to use privoxy immediately (before healthcheck) when the container restarts or the docker host reboots.
```

## Type of change

## Additional information

- This PR is related to issue: ```relates to #2489 ```

## Checklist

### DIFF
--- a/privoxy/scripts/start.sh
+++ b/privoxy/scripts/start.sh
@@ -27,7 +27,7 @@ set_port()
   # Set the port for the IPv4 interface
   adr=$(ip -4  a show eth0| grep -oP "(?<=inet )([^/]+)")
   adr=${adr:-"0.0.0.0"}
-  sed -i -E "s/^listen-address\s+127.*/listen-address ${adr}:$1/" "$2"
+  sed -i -E "s/^listen-address\s+(\b[0-9]{1,3}.){3}[0-9]{1,3}\b/listen-address ${adr}:$1/" "$2"
 
   # Remove the listen-address for IPv6 for now. IPv6 compatibility should come later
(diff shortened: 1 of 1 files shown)


pkishino:
please rebase this towards dev branch..I'll be away for the next few weeks, lets see if I have time to check this during that time, otherwise perhaps someone else can check this, thanks

mathiasflorin:
Thanks Patrick, I will do this rebase today. Sorry, I did not see the dev
branch.

mathiasflorin:
The dev branch already contains a regular expression  `.*`

PR #2691, opened 2023-08-30
Create separate image for transmission build from source

ksurl:
* moved transmission build stage to separate docker image
* created workflows to check for new transmission versions and build new image when found

## Breaking change

I'm not sure if the first version may need to be manually built and pushed to docker hub (so 4.0.4 first, and maybe 4.0.5 in the future will work as intended)
```txt
```

## Proposed change

Move the transmission build stage out of the main Dockerfile to speed up build times for image releases. The new base image for the transmission build would update on new transmission releases and only be rebuilt on new versions. The image releases would not rely on a new source build every time and just use the prebuilt image since the last release upstream.

```txt
```

## Type of change

- [x] New feature (which adds functionality to this container)

## Additional information

- This PR fixes or closes issue: ```fixes #2595 ```
- This PR is related to issue: ```relates to #2595```

## Checklist

- [ ] The code change is tested and works locally.


ksurl:
we can also try this approach 

https://docs.docker.com/build/ci/github-actions/multi-platform/

using matrix to run simultaneous builds for each platform

pkishino:
I think it looks great, if nothing further I think we can merge this

ksurl:
> I think it looks great, if nothing further I think we can merge this

There may be some setup required with the docker hub side creating the new image repo. I forgot if having a token automatically creates the repo if it doesn't exist when pushing a new image (and if it's public). Also it may fail the first run if it doesn't already exist when referencing the image in the main image build.

pkishino:
hmm, sure failing at a strange place..it cannot download the transmission release eh?..

ksurl:
It downloads it but there's a typo in the compressed file and the extract command. I had tried changing the path but changed my mind and missed that file name (missing the '4' in transmission4). Will fix in a bit.

PR #2744, opened 2023-11-12
Fix heading level of "Running this container with Podman" chapter in tips and tricks

MastermindPanda:
- Change heading level of "Running this container with Podman" in tips and tricks doc

## Breaking change

```txt
none
```

## Proposed change

```txt
Fix heading level of "Running this container with Podman"
```

## Type of change

## Additional information

As far as I understand, the chapter "Running with Traefik reverse proxy" and "Running this container with Podman" are independent and thus the latter one should be on top level.

But please correct me if I am wrong.

## Checklist

No code changes:


pkishino:
my bad ..fixed this manually, thanks

PR #2812, opened 2024-03-11
Allow specifying privoxy's listen address

Mossop:
## Breaking change

```txt
<placeholder>
```

## Proposed change

```txt
Allow setting the Privoxy listen address via an environment variable in case eth0 is not the preferred interface to listen on.
```

## Type of change

- [x] New feature (which adds functionality to this container)

## Additional information

## Checklist


pkishino:
Looks good to me. What kind of situations do you see this being useful for?

Mossop:
In my case I have my transmission container connected to two networks, one is bridged with some other containers but otherwise not accessible, the other a macvlan bridged to my LAN. I want privoxy listening on the latter but there appears no way to force docker to make that one eth0 so I added this change and set the bind address to 0.0.0.0 and I can then use the proxy.

PR #2846, opened 2024-05-30
fix: version file path

ksurl:
## Proposed change
```txt
add quotes to fix file detection for on push trigger
```

## Type of change

## Additional information

## Checklist

- [ ] The code change is tested and works locally.

### DIFF
--- a/.github/workflows/docker-transmission-builds.yml
+++ b/.github/workflows/docker-transmission-builds.yml
@@ -6,7 +6,7 @@ on:
     branches:
       - dev
     paths:
-      - upstream/transmission-version.txt
+      - "upstream/transmission-version.txt"
 
 jobs:
   build-transmission:


ksurl:
base image build did not trigger on new transmission release (the version check workflow did update the version file in the upstream folder as expected). seems like a common mistake to not have quotes on the file path and the workflow does not pick up the files as changed.

ksurl:
@pkishino can you manually run the Transmission Builds workflow? it should make a new base image based on transmission 4.0.6 so main image dev builds will use that going forward

ilike2burnthing:
Quotes are not required on paths, branches, or tags, unless they start with `*` - https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#patterns-to-match-file-paths

The reason the subsequent action wasn't triggered is because it requires a personal access token - https://github.com/orgs/community/discussions/25702

Fixed in https://github.com/haugene/docker-transmission-openvpn/pull/2843.

PR #2949, opened 2025-06-21
Add support for using a local PIA zip bundle rather than downloading the zip bundle from the website

farisk:
## Breaking change

```txt
<placeholder>
```

## Proposed change

```txt
Feature:
Rather than download openvpn config directly from PIA, use a locally saved source if it is available, otherwise download as normal.
The script will now look to /etc/openvpn/pia_custom_bundle.zip and use that if it is present, else carry on with the preexisting behaviour of downloading.

Usage:
When launching the image in docker specify the following volume:
-v /home/johnSmith/openvpn.zip:/etc/openvpn/pia_custom_bundle.zip

Why:
I had an issue when using the PIA config. My ISP blocks access to privateinternetaccess.com, so when I start the image up I get an error like:  curl: (6) Could not resolve host: www.privateinternetaccess.com ( from pia/configure-openvpn.sh )

First solution I did was use the "custom" option for OPENVPN_PROVIDER, whilst it worked the nice port forwarding scripts you have for PIA didn't run, so uploading didn't work, and I felt guilty for not seeding.

I did try to find an existing option like mounting /etc/openvpn/pia/ prepopulated with the config files but that failed. 

I also tried to use some other DNS, but it seems other DNS servers are blocked as well by my ISP.

The following solution works for me, as I can source the zip file from PIA myself, store it locally, and now every time I start the container it just works.

As a note - in case anyone else faces this issue - I did then have another problem with the PORT forwarding script, it seemed to fail because my ISPs DNS was still being used so the call to pia's API to generate a token timed out on name resolution, but as the container is connected to the VPN at this point using a custom DNS ( such as 8.8.8.8 from google )  was no longer blocked, so using that did resolve everything completely, and I can now happily torrent.
```

## Type of change

- [X] New feature (which adds functionality to this container)

## Additional information

## Checklist

- [X] The code change is tested and works locally.
- [X] There is no commented out code in this PR.

( I'll do this, just want confirmation you are happy with it as it is, or that I haven't missed some existing config option that makes this PR redundant )


haugene:
Hey :wave: I think this sounds like a good feature :+1: 

There can be many reasons not to download a zip file on every container startup, and PIA being blocked by the ISP is a good one. This is a good way to still get the port-forwarding and avoid going the CUSTOM provider route.

All in all I think the changes looks good. But I was thinking a little bit about what I see that Copilot also mentions in its suppressed comments.
1. The name `CUSTOM_BUNDLE` is a bit generic
2. Maybe the file location could be the environment variable so that people can place it wherever they want

The rest of its comments we can ignore. There is no documentation for the `PIA_OPENVPN_CONFIG_BUNDLE` so those who go looking into the scripts will know, others not. And we have no tests :see_no_evil: 


So, suggestion: Can you change the variable name to `PIA_CUSTOM_BUNDLE` and that it's value is a file path?
Then the if logic can look something like this:
```bash
# If a custom bundle is provided, use that instead of downloading from PIA
if [ -n "$PIA_CUSTOM_BUNDLE" ] && [ -f "$PIA_CUSTOM_BUNDLE" ]; then
    echo "Using custom OpenVPN config bundle: $PIA_CUSTOM_BUNDLE"
    unzip -qjo "$PIA_CUSTOM_BUNDLE" -d "$VPN_PROVIDER_HOME"
...
```

And you would need to set the environment variable `PIA_CUSTOM_BUNDLE=/etc/openvpn/pia_custom_bundle.zip` in addition to mounting the file, but it would be more generic and allow for different file names and locations to be used.

what do you think?

haugene:
Great :clap: Thanks for contributing and making the tweaks I wanted :)

PR #2978, opened 2025-10-26
The current setup is allowing data leaks in a dual stack network over IPv6.  Disabling IPv6 in the docker network should solve this.

MuttMutt-Outdoors:
## Breaking change

```txt
I do not believe this will break anything but the network changes potentially could cause an issue.  I have tested and researched the changes but those mainly are in the docker network.  I did my best to use the same information from a working container so it should be seamless.  I also tested the docker compose setup as a fresh install on Ubuntu and verified that the IPv6 leaks are fixed.
```

## Proposed change

```txt
The current setup is allowing DNS and potentially traffic to pass over the IPv6 address from the container in a dual stack network.  Simply disabling the IPv6 connectivity for the network that is being used by the Transmission-OpenVPN container should fix this issue.  It's mainly a change in how a container is setup in docker or podman and should be fairly  simple.  I was able to test the Docker Compose changes however I did not test the docker run so it should be tested but the documentation I read shows it should work. 

I also added a couple sections dealing with the testing of leaks coming from transmission for those people who decide they want to use a dual stack network.  The instructions should be fairly simple to follow and allow people to fix the potential issues or at least help them understand they need to test before they start using the setup in a dual stack environment.

Sorry if I goofed anywhere this is my first ever pull request to a public repo.  Let me know or feel free to fix. 
```

## Type of change

- [X] Bugfix (non-breaking change which fixes an issue)

## Additional information

## Checklist

- [X] The code change is tested and works locally.

- [X] Documentation added/updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,8 @@ $ podman run --privileged -d \
 version: '3.3'
 services:
     transmission-openvpn:
+        networks: 
+          - transmission_default
         cap_add:
             - NET_ADMIN
         volumes:
@@ -94,9 +96,10 @@ services:
         ports:
             - '9091:9091'
         image: haugene/transmission-openvpn
-        networks:
-          transmission_default:
-            enable_ipv6: false
+networks:
+  transmission_default:
+    name: transmission_default
+    enable_ipv6: false
 ```
 
 ### Docker version 2.x Compose
@@ -104,6 +107,8 @@ services:
 version: "2.0"
 services:
     transmission-openvpn:
+        networks: 
+          - transmission_default
         container_name: transmission
         cap_add:
             - NET_ADMIN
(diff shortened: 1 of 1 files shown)


ilike2burnthing:
This would likely break Mullvad and OVPN setups:

https://github.com/haugene/docker-transmission-openvpn/blob/dad76c5478d3721ae923852d2810537b5ab815e0/docs/provider-specific.md?plain=1#L66-L81

MuttMutt-Outdoors:
Simple enough to put in the documentation to comment out or remove that section for Mullvad or OVPN users.

Currently anyone who is using the default configuration in a dual stack environment is likely leaking DNS at the very least and possibly more.  Unless there is another way to disable IPv6 in the container.  I also have to assume that Mullvad and OVPN are able to work over an IPv4 connection since there are still ISP's who do have not implemented IPv6 on their network.  This change only disables IPv6 from the container to the local network, inside the container IPv6 should still continue to function.

ilike2burnthing:
```yml
         sysctls: 
             - "net.ipv6.conf.all.disable_ipv6=1" # 0 for Mullvad and OVPN users
```

Would this not be easier than creating a new network?

